PR #4, opened 2024-11-20
Shared library and shadcn/ui setup setup

niclasheun:
This PR adds shadcn/ui to the shared library and adds tailwind support. 
It is designed such that all components live in the shared library and can be imported from all other modules. 

To add a new component, execute the following command inside `shared_library`
`yarn dlx shadcn add <component-name>`

After that the component can be loaded via i.e.
`import { Button } from '@/components/ui/button'`


niclasheun:
This PR is based on the #3 . So #3 should be merged first

bgeisb (approved):
LGTM

PR #16, opened 2024-12-03
Deployment With integration of workflows

niclasheun:
Testing the workflow


niclasheun:
@Mtze 
The workflows are now fully setup and working. 
I had to adjust the workflows from the repo quite a bit:
- The workflow on the [.github main](https://github.com/ls1intum/.github/blob/main/.github/workflows/build-and-push-docker-image.yml) does not output any image tag. However, I need the image tags as I have a multi stage docker build and the components depend on the build of the root folder (=all shared dependencies). Thus, I copied and modified it. 
- And the deploy to [VM script ](https://github.com/ls1intum/.github/blob/feat/2-workflow-deploy-file-to-vm/.github/workflows/deploy-file-to-vm.yml) only provides the action to copy the file to the VM. But it does not include the docker-compose down, copying of environment variables, and the docker-compose up...

PR #22, opened 2024-12-07
Adding Keycloak To Server

niclasheun:
This PR shall be basis for discussion. I would really appreciate your comments @Mtze @robertjndw 

## The main idea:
- To implement fine-grained role management for Prompt
- Similar to Artemis, I introduced the groups: Lecturer, Editor, Student
  - On creation of every course, the server sends requests to keycloak to create the groups

## **My main question: Which roles to create?** 
- The big challenge is the following: We need access control on course and course phase level
  - Every course needs view / modify 
  - Every course phase: view meta data, modify meta data, set student passed

We have two possibilities how to solve this
1. Store in the server, which 'Main-Role' (Lecturer, Editor, Student) is allowed to do what
2. Have fine grained roles in the keycloak
  - i.e. <course_phase_uuid:view>, <course_phase_uuid:modify>, ...
  - Advantage: These roles can be associated with groups + very easy modification / adaptability
  - Problem: A course with i.e. 6 phases x 3 rights per phases x 10 Courses blows the Baerer Token up pretty quickly. 

### Keycloak Integration:

* Added Keycloak middleware for route protection and role-based access control in `server/course/router.go`. [[1]](diffhunk://#diff-a9a4468d8ae4274993f82b48c38290c8ba737029c138dd36743ec34fb3a6da58R9-R16) [[2]](diffhunk://#diff-a9a4468d8ae4274993f82b48c38290c8ba737029c138dd36743ec34fb3a6da58R28-R55)
* Implemented Keycloak client initialization and role/group management functions in `server/keycloak` package. [[1]](diffhunk://#diff-f0488fa03cd24eda29dbf08c9526a550348c5c5f8faed8444a38acbabdb18328R1-R52) [[2]](diffhunk://#diff-86d628fe462e42137f5ef408ac6233a26274aff86dd99303cc8d6306ad6fac3cR1-R37) [[3]](diffhunk://#diff-63996425b42b94f0dbd6e628ab1ec334a75f5cec94d355af53841352c3a971ceR1-R150) [[4]](diffhunk://#diff-b9cc57e3ebce81007c781af5c78921ccf22329ec789ca4215fca5e1ab095c213R1-R32) [[5]](diffhunk://#diff-9c21b859d35c60e96e83defcebdd2e1c1f26178df043f34afe5e0fda06075095R1-R64)
* Introduced Keycloak-related dependencies in `server/go.mod`. [[1]](diffhunk://#diff-7473344a45ba2ecf4f23fdb52551407eb4addcbcf539db5fcd4f6e424f24d7a5R6-R7) [[2]](diffhunk://#diff-7473344a45ba2ecf4f23fdb52551407eb4addcbcf539db5fcd4f6e424f24d7a5R39-R49) [[3]](diffhunk://#diff-7473344a45ba2ecf4f23fdb52551407eb4addcbcf539db5fcd4f6e424f24d7a5R70-R74) [[4]](diffhunk://#diff-7473344a45ba2ecf4f23fdb52551407eb4addcbcf539db5fcd4f6e424f24d7a5L81-R89)

### Service Layer:

* Added transaction handling and Keycloak role creation in `CreateCourse` function in `server/course/service.go`. [[1]](diffhunk://#diff-b7d4bf1d130b8a89da1c02e400171b75408191cd1cf1be74f3c7181d712fc2d5R77-R83) [[2]](diffhunk://#diff-b7d4bf1d130b8a89da1c02e400171b75408191cd1cf1be74f3c7181d712fc2d5R95-R104)


server/course/router.go
```diff
@@ -62,7 +62,7 @@ func getCourseByID(c *gin.Context) {
 		return
 	}
 
-	hasAccess, err := permissionValidation.CheckCoursePermission(c, id, []string{"Lecturer", "Editor", "Student"}, "courses:view-all")
+	hasAccess, err := permissionValidation.CheckCoursePermission(c, id, []string{keycloak.PromptAdmin, keycloak.CourseLecturer, keycloak.CourseEditor, keycloak.CourseStudent})
 	if err != nil || !hasAccess {
 		return
```
robertjndw, line 67: Maybe return a proper Error code here

server/course/router.go
```diff
@@ -100,7 +102,7 @@ func createCourse(c *gin.Context) {
...
 	}
 
-	course, err := CreateCourse(c, newCourse)
+	course, err := CreateCourse(c, newCourse, userID)
 	if err != nil {
 		handleError(c, http.StatusInternalServerError, err)
```
robertjndw, line 107: I would not return the plain error here. This may leak information to the enduser/attacker. This applies to most errors in this file here.

I would recommend to `log` the error and return a more generic one

server/course/service.go
```diff
@@ -96,7 +101,14 @@ func CreateCourse(ctx context.Context, course courseDTO.CreateCourse) (courseDTO
 	// create keycloak roles
 	err = CourseServiceSingleton.createCourseGroupsAndRoles(ctx, createdCourse.Name, createdCourse.SemesterTag.String)
 	if err != nil {
-		log.Debug("Failed to create keycloak roles for course: ", err)
+		log.Error("Failed to create keycloak roles for course: ", err)
+		return courseDTO.Course{}, err
+	}
+
+	roleString := fmt.Sprintf("%s-%s-Lecturer", createdCourse.Name, createdCourse.SemesterTag.String)
+	err = CourseServiceSingleton.addUserToGroup(ctx, requesterID, roleString)
```
robertjndw, line 109: What happens to the database object when this call fails? It should be rolled back, otherwise we have a database entry without its respective roles

server/permissionValidation/checkPermission.go
```diff
@@ -23,14 +24,14 @@ func checkUserRole(c *gin.Context, courseIdentifier string, allowedUsers []strin
...
-
 	// Generate the desired role keys based on input
 	for _, role := range allowedUsers {
-		desiredRole := fmt.Sprintf("%s-%s", courseIdentifier, role)
+		var desiredRole string = ""
+		if role == keycloak.PromptAdmin || role == keycloak.PromptLecturer {
+			desiredRole = role
+		} else {
+			desiredRole = fmt.Sprintf("%s-%s", courseIdentifier, role)
+		}
```
robertjndw, line 34: I would use a switch statement here: https://go.dev/wiki/Switch

server/permissionValidation/course.go
```diff
@@ -8,14 +8,14 @@ import (
 	"github.com/google/uuid"
 )
 
-func CheckCoursePermission(c *gin.Context, courseID uuid.UUID, allowedUsers []string, adminRight ...string) (bool, error) {
+func CheckCoursePermission(c *gin.Context, courseID uuid.UUID, allowedUsers []string) (bool, error) {
```
robertjndw, line 11: I would make this as a Variadic and not as a string slice
```suggestion
func CheckCoursePermission(c *gin.Context, courseID uuid.UUID, allowedUsers ...string) (bool, error) {
```

server/keycloak/addingCourse.go
```diff
@@ -17,9 +17,9 @@ func CreateCourseGroupsAndRoles(ctx context.Context, courseName, iterationName s
 
 	baseGroupName := fmt.Sprintf("%s-%s", courseName, iterationName)
 	subGroupNames := []string{
-		fmt.Sprintf("%s-Lecturer", baseGroupName),
-		fmt.Sprintf("%s-Editor", baseGroupName),
-		fmt.Sprintf("%s-Student", baseGroupName),
+		fmt.Sprintf("%s-%s", baseGroupName, CourseLecturer),
+		fmt.Sprintf("%s-%s", baseGroupName, CourseEditor),
+		fmt.Sprintf("%s-%s", baseGroupName, CourseStudent),
```
robertjndw, line 22: Is it necessary to build it in each request? Or should this be outsourced as "global" variable?

server/course/service.go
niclasheun: it is rolled back, as it happens within a transaction. If an error occurs, the function exits early and the transaction is never committed.

robertjndw: I highly doubt that this applies to actions which are executed on the Keycloak Groups and roles. Which my comment was actually referring to

PR #69, opened 2025-01-13
removing subpath and stripping url instead

niclasheun:
This is stripping the URL in the reverse proxy and removes the /template such that the components do not need to know under which path they are deployed.


niclasheun:
@Mtze FYI jetzt umgesetzt mit Path path stripping umgesetzt

Mtze:
Very nice :) Thanks!

PR #147, opened 2025-02-04
Keycloak Group Management

niclasheun:
# Important: depends on #124

Adding endpoints to
- create custom groups
- add students to custom groups
- get the members of custom groups

All custom groups are inside the subfolder "CustomGroups", inside the Course Folders. 
Each created group is associated with the role <SemesterTag>-<CourseName>-cg-<GroupName>


Mtze (approved):
tested on dev server

PR #153, opened 2025-02-09
Mailing settings improvements

niclasheun:
This addressed the issue that the mailing settings were disabled and not understandable for the user why. 
Now each disabled button has an explanatory tooltip and a little warning is displayed if any mail setting is disabled. 

image


Mtze (changes requested):
I tested this PR on our test server and noticed that the tooltip on the "manual send" buttons is still misleading. They are inactive because the mail has no subject or body but the tooltip shows to check the course mail settings.

Mtze (approved):
Now works as expected

PR #226, opened 2025-03-03
Interview Server Authentication Middleware

niclasheun:
# Overview
This PR introduces a suite of application middlewares for the Interview Server, aimed at streamlining authentication and authorization processes. The new implementation provides a robust framework for validating tokens, mapping user roles, and conditionally enforcing access controls based on allowed roles.

# New Middlewares
- **KeycloakMiddleware**: Validates incoming tokens and extracts user claims from them.
- **LecturerAndEditorMiddleware**: Retrieves token mappings from the Core service to extract and validate the correct tokens for lecturer and editor roles.
- **IsStudentMiddleware**: Queries the Core service to verify if the user is an active student for the specific course phase.
- **AuthenticationMiddleware**: Dynamically chains the above middlewares based on allowed roles, ensuring that users are authenticated and authorized correctly.

# Future Enhancements
- **Caching**: Future iterations will include caching mechanisms to reduce the number of requests to the Core service.
- **Shared Library**: These middlewares are intended to be moved to a shared Go library/package to enhance reusability across multiple services.


.vscode/settings.json
Mtze, line 1: Please remove this

Mtze (changes requested):

.vscode/settings.json
niclasheun: happened by mistake. Is fixed now

servers/intro_course/keycloakTokenVerifier/keycloakTokenVerifierDTO/get_roles.go
```diff
@@ -0,0 +1,7 @@
+package keycloakTokenVerifierDTO
+
+type GetCourseRoles struct {
+	CourseLecturerRole string `json:"courseLecturerRole"`
+	CourseEditorRole   string `json:"courseEditorRole"`
+	CustomGroupPrefix  string `json:"customGroupPrefix"`
```
Mtze, line 6: i'd prefer to call this `CustomRolePrefix` to be consistent with the other two

servers/intro_course/utils/getCoreUrl.go
```diff
@@ -0,0 +1,13 @@
+package utils
+
+import "strings"
+
+func GetCoreUrl() string {
```
Mtze, line 5: This will most likely fail in a docker env. Since you assume https only - containers can no longer communicate with each other

Mtze: I'd even argue that this logic is not necessary. I think it is enough to assume a correct value in the env.

Mtze (changes requested):
See inline comments

Mtze (approved):

PR #249, opened 2025-03-11
`Intro Course` Tutor Import via Keycloak

niclasheun:
This PR allows Instructors to import students from another course as tutors into the intro course.
(Use Case: One course manages all tutors (application, matching etc), one course manages the students) 

## Added Features

1.) You can create the required custom group via 1-Button-Click (the name additionally gets saved to the CoursePhase-MetaData (to be able to pass it to last phases)
2.) Select students from other courses
3.) Import these students (on import the student is added to the customGroup AND the Editor Role of the course)

## How To Test

- Click on "create Keycloak Group" -> look if a new custom group was added to keycloak
- import students -> test if student has been added as member to the groups

## Known Issues

- Assigned Tutors can not yet be deleted -> later PR (this is already quite big)


Mtze (approved):
Changes looking good to me - We should add an error indicator for failed keycloak imports to prevent confusion in a follow up PR

PR #268, opened 2025-03-13
`General`: Try local agents

Mtze:


Mtze:
The Build execution is significantly faster on the local agents :)

Mtze:
Next step - Try out ARM on prem agents

PR #346, opened 2025-04-01
`General`: Fixing the Loading of Courses

niclasheun:
## Problem
I noticed, that after re-login the courses are not correctly displayed in the sidebar. Only after reloading the side the courses are correctly displayed. 

## Solution
Upon inspection, the courses are not displayed because the requests fails with a 401 Unauthorized. The reason for this is, that it tries to fetch the courses before the keycloak token is correctly saved. 
Hence, I postponed the request until the keycloak token is correctly stored and can be used for the requests. No it works as expected.

### DIFF
--- a/clients/core/src/managementConsole/ManagementConsole.tsx
+++ b/clients/core/src/managementConsole/ManagementConsole.tsx
@@ -5,7 +5,7 @@ import { SidebarInset, SidebarProvider, SidebarTrigger } from '@/components/ui/s
 import { AppSidebar } from './layout/Sidebar/AppSidebar'
 import { WelcomePage } from './shared/components/WelcomePage'
 import { LoadingPage } from '@/components/LoadingPage'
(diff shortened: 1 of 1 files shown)


rappm:
LGTM

rappm (approved):

clients/core/src/managementConsole/ManagementConsole.tsx
```diff
@@ -106,14 +109,14 @@ export const ManagementRoot = ({ children }: { children?: React.ReactNode }): JS
...
 
-  if (isLoading) {
-    return <LoadingPage />
-  }
-
   if (isError) {
     return <ErrorPage onRetry={() => refetch()} onLogout={() => logout()} />
   }
 
+  if (isLoading || !keycloak) {
```
mathildeshagl, line 116: Good update!

mathildeshagl (approved):
LGTM

PR #401, opened 2025-04-13
`Team Allocation`: Student Endpoint for Tease Integration

niclasheun:
Important: This depends on the PRs #397  and #396


servers/team_allocation/tease/teaseDTO/skillResponse.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"github.com/google/uuid"
+	db "github.com/ls1intum/prompt2/servers/team_allocation/db/sqlc"
+)
+
+type SkillResponse struct {
+	ID          uuid.UUID `json:"id"`
+	Proficiency string    `json:"proficiency"`
+}
```
mathildeshagl, line 11: Since this corresponds to "StudentSkill" in Tease and there is also another type "Skill", for consistency, I would call this "StudentSkill" here, too.

```suggestion
type StudentSkillResponse struct {
	ID          uuid.UUID `json:"id"`
	Proficiency string    `json:"proficiency"`
}
```

servers/team_allocation/tease/teaseDTO/skillResponse.go
```diff
@@ -0,0 +1,38 @@
...
+func GetTeaseSkillResponseFromDBModel(skillResponses []db.GetStudentSkillResponsesRow) []SkillResponse {
+	skills := make([]SkillResponse, 0, len(skillResponses))
+	for _, skillResponse := range skillResponses {
+		skills = append(skills, SkillResponse{
+			ID:          skillResponse.SkillID,
+			Proficiency: getTeaseSkillLevel(skillResponse.SkillLevel),
+		})
+	}
+	return skills
+}
```
mathildeshagl, line 22: Same here then

```suggestion
func GetTeaseStudentSkillResponseFromDBModel(skillResponses []db.GetStudentSkillResponsesRow) []StudentSkillResponse {
	skills := make([]SkillResponse, 0, len(skillResponses))
	for _, skillResponse := range skillResponses {
		skills = append(skills, SkillResponse{
			ID:          skillResponse.SkillID,
			Proficiency: getTeaseSkillLevel(skillResponse.SkillLevel),
		})
	}
	return skills
}
```

mathildeshagl (changes requested):

Mtze:
Do we have a tease test env deployed to test this integration?

servers/team_allocation/tease/teaseDTO/skillResponse.go
niclasheun: I disagree here. This is not supposed to model the Prompt Data model but to model the objects required from Tease. 

Choosing here StudentSkill instead of skill shows that this is not the same and StudentSkill is the name used by tease. 

Using Skill would be counter intuitive in my understanding as the Prompt Skill does not have a "proficiency"

niclasheun: Fair point

niclasheun:
No we do not yet have the test env but Philipp is working on it

rappm (changes requested):

mathildeshagl (commented):
LGTM

rappm (approved):

PR #406, opened 2025-04-15
`Team Allocation`: DB Migration and Allocations Endpoints

mathildeshagl:
## Summary by CodeRabbit

- **New Features**
  - Introduced a new allocations system to manage student-to-team assignments within course phases.
  - Added endpoints to retrieve and update team allocations for authorized users.
  - Enabled batch creation and updating of allocations via JSON API.
  - Provided aggregated views of allocations grouped by teams for each course phase.

- **Bug Fixes**
  - Enhanced validation and error handling for allocation operations.

- **Chores**
  - Updated internal tooling versions for improved code generation and maintenance.


servers/team_allocation/allocations/router.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	allocation, err := GetStudentAllocation(c, courseParticipationID, coursePhaseID)
+	if err != nil {
+		handleError(c, http.StatusInternalServerError, err)
+		return
+	}
+
+	c.JSON(http.StatusOK, allocation)
+}
```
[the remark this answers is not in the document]
mathildeshagl: the :coursePhaseID is already defined in the overall URL in main.go of the team_allocation component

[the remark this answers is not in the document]
mathildeshagl: that is correct, changed that

servers/team_allocation/allocations/router.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	err := PutAllocation(c, req.CourseParticipationID, req.TeamID, req.CoursePhaseID)
+	if err != nil {
+		handleError(c, http.StatusInternalServerError, err)
+		return
+	}
+
+	c.JSON(http.StatusCreated, gin.H{"message": "Allocation created successfully"})
+}
```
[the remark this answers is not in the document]
mathildeshagl: That is now done in the service.go method, is this then still needed?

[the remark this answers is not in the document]
mathildeshagl: Thanks for the explanation, implemented this👍🏻

servers/team_allocation/allocations/allocationDTO/allocationRequest.go
```diff
@@ -0,0 +1,12 @@
+package allocationDTO
+
+import (
+	"github.com/google/uuid"
+)
+
+type AllocationRequest struct {
+	ID                    uuid.UUID `json:"id"`
+	CourseParticipationID uuid.UUID `json:"course_participation_id"`
```
niclasheun, line 9: Have you double checked this with the tease specification? I'm pretty sure it uses camelCase so the json should also be camelCase to work

servers/team_allocation/allocations/router.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/ls1intum/prompt-sdk/keycloakTokenVerifier"
+	"github.com/ls1intum/prompt2/servers/team_allocation/allocations/allocationDTO"
+	db "github.com/ls1intum/prompt2/servers/team_allocation/db/sqlc"
+)
+
+func setupAllocationsRouter(routerGroup *gin.RouterGroup, authMiddleware func(allowedRoles ...string) gin.HandlerFunc) {
+	allocationsRouter := routerGroup.Group("/allocations")
+
+	// we need the keycloak middleware here to ensure that the user has a valid token
+	allocationsRouter.GET("/course-phases", keycloakTokenVerifier.KeycloakMiddleware(), getAllCoursePhases)
```
niclasheun, line 19: Please do not create a new package for this. This is all meant to be part of the /tease package.

Could you please check my PR I created in the tease repo with all the endpoint specifications and match that ?

servers/team_allocation/allocations/router.go
```diff
@@ -0,0 +1,121 @@
...
+
+	allocationsRouter.PUT("", authMiddleware(promptSDK.PromptAdmin, promptSDK.CourseLecturer), putAllocation)
+	allocationsRouter.GET("/:courseParticipationID", authMiddleware(promptSDK.PromptAdmin, promptSDK.CourseLecturer), getStudentAllocation)
+	allocationsRouter.GET("", authMiddleware(promptSDK.PromptAdmin, promptSDK.CourseLecturer), getAllAllocations)
+}
+
+func getAllCoursePhases(c *gin.Context) {
```
niclasheun, line 26: Why have you copied this endpoint here?

servers/team_allocation/allocations/router.go
```diff
@@ -0,0 +1,121 @@
...
+}
+func putAllocation(c *gin.Context) {
+	var req allocationDTO.AllocationRequest
+
+	if err := c.BindJSON(&req); err != nil {
+		handleError(c, http.StatusBadRequest, err)
+		return
+	}
+
+	arg := db.GetAllocationForStudentParams{
```
niclasheun, line 99: Please do not do any DB logic in the router

servers/team_allocation/allocations/service.go
```diff
@@ -0,0 +1,171 @@
...
+			KickoffSubmissionPeriodEnd: deadline,
+		})
+	}
+
+	// 4. Return these course phases with names
+	return teasePhases, nil
+}
+
+func hasCoursePhasePermission(userPermissions map[string]bool, semesterTag, courseName string) bool {
+	// Admins always have permission
```
niclasheun, line 92: Same here. The idea is to have everything in the tease folder.

servers/team_allocation/db/migration/0003_allocations.up.sql
```diff
@@ -0,0 +1,17 @@
...
+
+CREATE TABLE allocations (
+    id uuid NOT NULL PRIMARY KEY,
+    course_participation_id uuid NOT NULL,
+    team_id uuid NOT NULL,
+    course_phase_id uuid NOT NULL,
+    created_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    FOREIGN KEY (course_participation_id) REFERENCES course_participation(id) ON DELETE CASCADE,
+    FOREIGN KEY (team_id) REFERENCES team(id) ON DELETE CASCADE,
+    FOREIGN KEY (course_phase_id) REFERENCES course_phase(id) ON DELETE CASCADE,
```
niclasheun, line 11: This will not work. You do not have a table with course phases in the course phase service.

Please run sqlc generate to make sure that your sqlc code actually compiles

servers/team_allocation/db/migration/0003_allocations.up.sql
```diff
@@ -0,0 +1,17 @@
+BEGIN;
+
+CREATE TABLE allocations (
+    id uuid NOT NULL PRIMARY KEY,
+    course_participation_id uuid NOT NULL,
+    team_id uuid NOT NULL,
+    course_phase_id uuid NOT NULL,
+    created_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    FOREIGN KEY (course_participation_id) REFERENCES course_participation(id) ON DELETE CASCADE,
+    FOREIGN KEY (team_id) REFERENCES team(id) ON DELETE CASCADE,
```
niclasheun, line 10: This constraint is not strong enough. We currently could get inconsistencies as both the team and the allocation has a course phase. So potentially there could be ambiguity.

niclasheun (changes requested):

niclasheun:
Have not reviewed the code completely but rather skimmed through it, but I think there are some structural issues.

servers/team_allocation/allocations/service.go
```diff
@@ -0,0 +1,171 @@
...
+			CoursePhaseID: coursePhaseID,
+			TeamID:        teamID,
+		}
+		err = qtx.UpdateAllocation(ctx, updateArg)
+		if err != nil {
+			log.Error("could not update the allocation in the database: ", err)
+			return errors.New("could not update the allocation in the database")
+		}
+	} else if !errors.Is(err, sql.ErrNoRows) {
+		log.Error("error checking for existing allocation: ", err)
```
niclasheun, line 151: also please check with @phnagy if this is the expected behavior. I'm not sure if it is a great idea to throw an error becuase just one student did not yet have a team allocation. I would rather expect a behavior, which updates a team allocation, if there is one and creates one if there is none. But I do not know how TEASE works.

servers/team_allocation/db/migration/0003_allocations.up.sql
```diff
@@ -0,0 +1,14 @@
...
+CREATE TABLE allocations (
+    id uuid NOT NULL PRIMARY KEY,
+    course_participation_id uuid NOT NULL,
+    team_id uuid NOT NULL,
+    created_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+
+    FOREIGN KEY (team_id) REFERENCES team(id) ON DELETE CASCADE,
+
+    CONSTRAINT unique_allocation_per_student UNIQUE (course_participation_id)
+);
```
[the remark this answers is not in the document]
mathildeshagl: Fk reference is not possible

servers/team_allocation/db/query/allocation.sql
```diff
@@ -0,0 +1,22 @@
+-- name: GetAllocationsByCoursePhase :many
+SELECT a.*
+FROM allocations a
+JOIN team t ON a.team_id = t.id
+WHERE t.course_phase_id = $1;
+
+-- name: CreateOrUpdateAllocation :exec
+INSERT INTO allocations (id, course_participation_id, team_id)
+VALUES ($1, $2, $3)
+ON CONFLICT (course_participation_id) DO UPDATE
```
rappm, line 10: If we use a timestamp, we should also update the timestamp on update

rappm (changes requested):

servers/team_allocation/db/migration/0003_allocations.up.sql
```diff
@@ -0,0 +1,13 @@
...
+
+CREATE TABLE
+    allocations (
+        id uuid NOT NULL PRIMARY KEY,
+        course_participation_id uuid NOT NULL,
+        team_id uuid NOT NULL,
+        created_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+        FOREIGN KEY (team_id) REFERENCES team (id) ON DELETE CASCADE,
+        UNIQUE (course_participation_id, team_id)
+    );
```
niclasheun, line 11: this does not sufficiently resolve my comment. You removed the coursePhaseID, but introduced two new problems. 

1. We now we can have a potential "permission break". Your endpoint just adds allocations to teamIDs. It is never checked whether these teams actually belong to the same course phase as the requested endpoint (could be a completely different course or coursePhase). 

2. We do not have uniquieness any more. A course participation could be potentially associated with multiple teams. This is also why your update logic will not work. 

You rather want something like the following

servers/team_allocation/db/migration/0003_allocations.up.sql
```diff
@@ -0,0 +1,13 @@
+BEGIN;
+
```
niclasheun, line 2:
```suggestion
CREATE TABLE allocations (
  id                          UUID    NOT NULL PRIMARY KEY,
  course_participation_id     UUID    NOT NULL,
  team_id                     UUID    NOT NULL,
  course_phase_id             UUID    NOT NULL,
  created_at                  TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,

  -- still enforce that team_id itself exists...
  FOREIGN KEY (team_id)
    REFERENCES team (id)
    ON DELETE CASCADE,

  -- …and also enforce that the phase matches the one on the team:
  FOREIGN KEY (team_id, course_phase_id)
    REFERENCES team (id, course_phase_id)
    ON DELETE CASCADE,

  UNIQUE (course_participation_id, team_id)
);
```

servers/team_allocation/db/query/allocation.sql
```diff
@@ -0,0 +1,22 @@
+-- name: GetAllocationsByCoursePhase :many
+SELECT a.*
+FROM allocations a
+JOIN team t ON a.team_id = t.id
+WHERE t.course_phase_id = $1;
```
niclasheun, line 5: please adjust these accordingly after the allocation db migration updates

servers/team_allocation/tease/service.go
```diff
@@ -172,3 +172,59 @@ func getTeaseSkillLevel(ctx context.Context, coursePhaseID uuid.UUID, coursePart
...
 	return teaseDTO.GetTeaseStudentSkillResponseFromDBModel(skills), nil
 }
+
+func GetAllocationsByCoursePhase(ctx context.Context, coursePhaseID uuid.UUID) ([]db.Allocation, error) {
+	dbAllocations, err := TeaseServiceSingleton.queries.GetAllocationsByCoursePhase(ctx, coursePhaseID)
+	if err != nil {
+		log.Error("could not get the allocations from the database: ", err)
+		return nil, fmt.Errorf("could not get the allocations from the database: %w", err)
+	}
+	return dbAllocations, nil
```
niclasheun, line 182: Why do you return db.Allocation as type here? This does not align with the TEASE API spec. Please transform it in a suited DTO

servers/team_allocation/tease/service.go
```diff
@@ -172,3 +172,59 @@ func getTeaseSkillLevel(ctx context.Context, coursePhaseID uuid.UUID, coursePart
...
+					if err != nil {
+						log.Errorf("failed to create allocation for student %s: %v", studentID, err)
+						return fmt.Errorf("failed to create allocation: %w", err)
+					}
+				} else {
+					log.Errorf("error checking existing allocation for student %s: %v", studentID, err)
+					return fmt.Errorf("error fetching existing allocation: %w", err)
+				}
+			} else {
+				err = qtx.CreateOrUpdateAllocation(ctx, db.CreateOrUpdateAllocationParams{
```
niclasheun, line 212: why do you have a create or update method if you differentiate between these cases? The idea of a createOrUpdate is that you do not net a Get Allocation and a case differentiation

niclasheun (changes requested):

servers/team_allocation/tease/service.go
mathildeshagl: Would you rather have a createOrUpdate and not differenciate or differenciate and have two seperate methods?

niclasheun: both has it pros and cons. In general, I'm a fan of moving stuff to the DB because it makes the code more performant. In this particular case, it is not as critical because it is not called so often.  But if you have such endpoint, especially with for loops, you should rather use a CreateOrUpdate or even better a Batch DB Operation. Otherwise you will have ~ 60 students x 2 database operations.  (-> UNNEST function)

mathildeshagl: I made it a createorupdate function👍🏻

servers/team_allocation/tease/service.go
mathildeshagl: Oops, have a DTO defined, but didn't use it here.🙈

servers/team_allocation/tease/service.go
mathildeshagl: @niclasheun could you maybe take a look at the CreateOrUpdate function in `allocation.sql`, if it makes sense to do it like this (I wanted to keep the original `created_at` value, when updating an allocation) or should we omit the `updated_at` value of an allocation in the DB completely and always overwrite the `created_at` value?

servers/team_allocation/tease/service.go
```diff
@@ -198,3 +198,54 @@
...
+			log.Error("could not get students for team: ", err)
+			return nil, fmt.Errorf("could not get students for team: %w", err)
+		}
+		teaseAllocations = append(teaseAllocations, teaseDTO.Allocation{
+			ProjectID: dbAllocation.TeamID,
+			Students:  students,
+		})
+	}
+	return teaseAllocations, nil
+}
```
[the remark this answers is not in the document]
mathildeshagl: this is the intended behaviour. these endpoints are for another service (TEASE) and that service requires the teaseDTO.Allocation exactly as it is defined.

servers/team_allocation/tease/service.go
niclasheun: I get what you did here (bc. of the code rabbit corrections) but I do not really like the way to update it. 
I would much more prefer this solution (from chatGPT)

-- ────────────────────────────────────────────────────────────────────────────────
-- 1.  Ensure we can target the conflict (not from me -> add this to the migration)
-- ───────────────────────────────────────────────────────────────────────────────
ALTER TABLE allocations
    ADD CONSTRAINT allocations_participation_phase_uk
        UNIQUE (course_participation_id, course_phase_id);

-- ────────────────────────────────────────────────────────────────────────────────
-- 2.  Create‑or‑update (upsert): keeps id & created_at, refreshes updated_at
-- ────────────────────────────────────────────────────────────────────────────────
-- name: CreateOrUpdateAllocation :exec
INSERT INTO allocations AS a (
        id,
        course_participation_id,
        team_id,
        course_phase_id,
        created_at,
        updated_at
) VALUES (
        $1,         -- id to use on first insert only
        $2,         -- course_participation_id
        $3,         -- team_id
        $4,         -- course_phase_id
        CURRENT_TIMESTAMP,   -- created_at (first insert only)
        CURRENT_TIMESTAMP    -- updated_at
)
ON CONFLICT ON CONSTRAINT allocations_participation_phase_uk
DO UPDATE
   SET team_id    = EXCLUDED.team_id,
       updated_at = CURRENT_TIMESTAMP;   -- keep original id & created_at

servers/team_allocation/db/query/allocation.sql
```diff
@@ -0,0 +1,48 @@
...
+
+-- name: CreateOrUpdateAllocation :exec
+WITH deleted AS (
```
niclasheun, line 8: see comment from conversation

servers/team_allocation/db/migration/0003_allocations.up.sql
```diff
@@ -0,0 +1,18 @@
...
+        course_participation_id UUID NOT NULL,
+        team_id UUID NOT NULL,
+        course_phase_id UUID NOT NULL,
+        created_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+        updated_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+        -- still enforce that team_id itself exists...
+        FOREIGN KEY (team_id) REFERENCES team (id) ON DELETE CASCADE,
+        -- …and also enforce that the phase matches the one on the team:
+        FOREIGN KEY (team_id, course_phase_id) REFERENCES team (id, course_phase_id) ON DELETE CASCADE,
+        UNIQUE (course_participation_id, team_id)
```
niclasheun, line 15:
```suggestion
CONSTRAINT allocations_participation_phase_uk UNIQUE (course_participation_id, course_phase_id)


```

servers/team_allocation/tease/service.go
```diff
@@ -198,3 +198,64 @@ func GetTeaseTeamsByCoursePhase(ctx context.Context, coursePhaseID uuid.UUID) ([
 
 	return teaseDTO.GetTeaseTeamResponseFromDBModel(teams), nil
 }
+
+func GetAllocationsByCoursePhase(ctx context.Context, coursePhaseID uuid.UUID) ([]teaseDTO.Allocation, error) {
+	dbAllocations, err := TeaseServiceSingleton.queries.GetAllocationsByCoursePhase(ctx, coursePhaseID)
```
niclasheun, line 203: I think there is no reason to have a for loop and separate queries. A simple agg function does the job for this very simple form or agg.

SELECT
  team_id,
  array_agg(course_participation_id::uuid ORDER BY course_participation_id)::uuid[] AS course_participation_ids
FROM allocations
WHERE course_phase_id = $1
GROUP BY team_id
ORDER BY team_id;

niclasheun (changes requested):
Most stuff looks fine. I added concrete suggestions on how to improve the createOrUpdate function and the retrieval of the team allocations

servers/team_allocation/tease/service.go
mathildeshagl: that's a good point

servers/team_allocation/tease/service.go
mathildeshagl: Thanks for looking into it 😊

servers/team_allocation/db/migration/0003_allocations.up.sql
```diff
@@ -0,0 +1,18 @@
+BEGIN;
+
```
niclasheun, line 2:
```suggestion
ALTER TABLE team
  ADD CONSTRAINT team_id_course_phase_uk UNIQUE (id, course_phase_id);
  ```

**Review by niclasheun — CHANGES_REQUESTED** [review]
Mostly looks good to me. But please add the fix for the DB Migration such that we can test it before merging

File: servers/team_allocation/db/migration/0003_allocations.up.sql (hunk 22 above)
Comment (mathildeshagl) [reply servers/team_allocation/db/migration/0003_allocations.up.sql:2]: Implemented that, it runs locally now as well.

File: servers/team_allocation/tease/service.go [hunk 23]
```diff
@@ -198,3 +198,61 @@ func GetTeaseTeamsByCoursePhase(ctx context.Context, coursePhaseID uuid.UUID) ([
...
+			continue
+		}
+		for _, studentID := range allocation.Students {
+			if studentID == uuid.Nil {
+				return fmt.Errorf("invalid student ID in allocation")
+			}
+			err = qtx.CreateOrUpdateAllocation(ctx, db.CreateOrUpdateAllocationParams{
+				ID:                    uuid.New(),
+				CourseParticipationID: studentID,
+				TeamID:                allocation.ProjectID,
```
niclasheun, line 243: I think you are missing the coursePhaseID here

niclasheun (changes requested):

niclasheun (approved):

niclasheun:
Scheint alles zu klappen. Vielen Danke dir!

rappm (approved):

rappm:
Looks good to me!

PR #434, opened 2025-04-22
`Core`: Improving Application Auth

niclasheun:
Currently,  we grabbed an existing student by email. This leads to an error if a student changes their email in Tum Online. Thats why this approach tries to fetch the user by matriculationNumber and universityLogin.

## Summary by CodeRabbit

- **New Features**
  - Added support for retrieving student records and application forms using matriculation number and university login instead of email.
- **Bug Fixes**
  - Improved authentication checks by requiring both matriculation number and university login for access.
- **Refactor**
  - Updated internal logic to use new student lookup methods based on matriculation number and university login.


rappm (approved):

rappm:
Nice!

PR #448, opened 2025-04-23
`Team Allocation`: First Version of the Self Team Allocation Server

robertjndw:


niclasheun:
On a high level looks good to me. I'll merge it into my branch such that I can add some adjustments and test it.

PR #799, opened 2025-08-07
`Core`: Show a preview of recipient count before manually sending out email

### LINKED ISSUE #202: Show the number of mails that would be send
When I send the accept/reject emails, a pop-up window appears asking for confirmation.
It would be great if it showed the actual number of emails to be sent so that I can verify this number:

Image

JGStyle:
## ✨ What is the change?

When manually sending out acceptance or rejection mails, the user will now see a preview of how many people will receive that email.

fixes #202 

## 🧪 How to Test

1. clone into this branch, start the database, server and client
2. create a course that has an application phase
3. add some members (and accept or reject some of them)
4. Go to Course Phases > Your application phase > Mailing  and click on "Send Acceptance Mails" or "Send Rejection Mails" to see the preview

comparison

## What I changed / added
- The ConfirmSendEmailDialog component, which handles manually sending out emails, will now, upon triggering fetch the recipients count from a new REST route: /api/coure_phases/:uuid/participation_status_counts
- There is a new corresponding endpoint in the go main server that handles the request, sums up the amounts and returns them

_Reasoning:_
Backend
- While some coursephases have shared state that could be used to leverage existing data and reduce the amount of requests, the ManualMailSending component is being used by multiple course phases.
- therefore the counter that shows the amount of recipients must work for all course phases and can not make use of a zustand (like the one from the Application) part or other state for a specific coursephase.
- Since every course phase has the table course_phase_participation, this is the best place to get the info for every course phase, that's why adding a query here makes sense
Frontend
- I put the acutal trigger for the request inside of the ConfirmSendEmailDialog, because (1): if the user does not  not trigger the dialog, no request will be performed, keeping the system minimal (2): inside of this dialog, the current coursephase id is already being used for the other requests (like the ones requesting the mail send) anyways, allowing me to reuse it

## ✅ PR Checklist

- No Documentation added, deemed irrelevant

### DIFF
--- a/servers/core/coursePhase/router.go
+++ b/servers/core/coursePhase/router.go
@@ -22,6 +22,7 @@ func setupCoursePhaseRouter(router *gin.RouterGroup, authMiddleware func() gin.H
 	coursePhase := router.Group("/course_phases", authMiddleware())
 	coursePhase.GET("/:uuid", permissionIDMiddleware(permissionValidation.PromptAdmin, permissionValidation.CourseLecturer, permissionValidation.CourseEditor, permissionValidation.CourseStudent), getCoursePhaseByID)
 	coursePhase.GET("/:uuid/course_phase_data", permissionIDMiddleware(permissionValidation.PromptAdmin, permissionValidation.CourseLecturer, permissionValidation.CourseEditor, permissionValidation.CourseStudent), getPrevPhaseDataByCoursePhaseID)
+	coursePhase.GET("/:uuid/participation_status_counts", permissionIDMiddleware(permissionValidation.PromptAdmin, permissionValidation.CourseLecturer, permissionValidation.CourseEditor), getCoursePhaseParticipationStatusCounts)
 
 	// getting the course ID here to do correct rights management
 	coursePhase.POST("/course/:courseID", permissionCourseIDMiddleware(permissionValidation.PromptAdmin, permissionValidation.CourseLecturer), createCoursePhase)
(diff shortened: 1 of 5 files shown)


servers/core/coursePhase/router.go
```diff
@@ -218,3 +219,36 @@ func handleError(c *gin.Context, statusCode int, err error) {
...
+		handleError(c, http.StatusInternalServerError, err)
+		return
+	}
+
+	// Convert the slice of structs to a map
+	countsMap := make(map[string]int)
+	for _, count := range counts {
+		status := string(count.PassStatus.PassStatus)
+		countsMap[status] = int(count.Count)
+	}
```
rappm, line 251: Let’s move this logic into a service method called `GetCoursePhaseParticipationStatusCounts` in `servers/core/coursePhase/service.go`. That way the router just handles requests and responses, and the service takes care of the logic. This is how we usually keep our servers organised in the Prompt Project.

To be fair - this isn't mentioned in the server guidelines yet.

rappm (changes requested):

servers/core/coursePhase/router.go
JGStyle: Great catch, i didn't notice it the first time. Fixed it in [5524004](https://github.com/ls1intum/prompt2/pull/799/commits/55240047220e01b66608ccb2b56df372452c670e)

JGStyle:
Lastly, if we go by principle 7 of the contributor guidelines 

> Eliminate Dead Code Immediately

I could also go ahead and remove that `course_phase_data` route - what do you think?

again, searching through the entire project, there is no usage
image

Also lsp_references show no usage except for the definition

niclasheun:
First of all, I'm deeply impressed that someone is actually reading the docs 😍👏. 

I, personally, would not call this a dead code. You are right, that it is currently not used, but the core API is designed to offer a stable, rarely changing interface for all other phases. Thus, we designed it with additional functionality. 

I do, however, see the argument, that it is unlikely that an external course phase just accesses the course phase data and not directly the entire course phase (where the course phase data is included). So I think there are valid arguments for keeping but also for removing it.

rappm:
So I would strongly argue that we shouldn't remove it 😂

But really nice catch, I also had to double check if we use it anywhere. This is one of the disadvantages of the SDK, rg doesn't always help.

rappm:
What's the status of this? Is it still a draft?

PR #1205, opened 2026-02-19
`Interview`: Fix interview question race condition with optimistic updates

### LINKED ISSUE #1204: Interview: after writing notes, adding another question is wonky / doesn't work

mathildeshagl:
## ✨ What is the change?

Fixed a critical race condition bug where adding a new interview question after writing notes would cause the new question to disappear or behave unpredictably. The fix includes:

- Removed circular dependency in QuestionConfiguration auto-save pattern using ref-based change tracking
- Added 500ms debouncing to prevent rapid-fire mutations
- Implemented optimistic updates for both questions and notes with automatic rollback on error
- Added cross-cache invalidation to ensure UI consistency across all interview views
- Added loading indicators and error handling for better UX

Fixes #1204

The bug occurred because the auto-save useEffect in QuestionConfiguration depended on the entire `coursePhase` object, creating a circular update loop: save → invalidate cache → refetch → coursePhase updates → triggers save again with potentially stale data → new questions get lost.

## 🧪 How to Test

1. Navigate to Interview Component → Settings → Question Configuration
2. Add a new question and verify "Saving..." indicator appears
3. Navigate to Manage Interviews tab → Select an interviewee
4. Verify the new question appears in the interview card
5. Write notes in an existing question field → blur the input
6. **Immediately add another new question** (this was the failing scenario)
7. Verify both the notes are saved AND the new question appears without disappearing
8. Test rapid additions: Add 3-5 questions quickly in succession
9. Verify all questions appear correctly (debouncing should group them)
10. Test error scenario: Disconnect network → add question → verify error message and rollback

N/A - Bug fix with minor UI improvement (saving indicator)

## ✅ PR Checklist

## Summary by CodeRabbit

* **New Features**
  * Debounced auto-save for interview questions (500ms) with a visible saving indicator and inline error message.
  * Optimistic UI updates for phase and participation changes with success/error toasts and automatic resync.

* **Bug Fixes**
  * Prevents spurious saves during initial load to avoid unintended persistence.
  * Rolls back and surfaces errors when updates fail.

* **Stability**
  * Preserves add/delete/reorder question behavior with improved load/save handling.

### DIFF
--- a/clients/interview_component/src/interview/pages/Settings/QuestionConfiguration.tsx
+++ b/clients/interview_component/src/interview/pages/Settings/QuestionConfiguration.tsx
@@ -4,7 +4,7 @@ import { useDebouncedCallback } from 'use-debounce'
 import type { InterviewQuestion } from '../../interfaces/InterviewQuestion'
 import { useCoursePhaseStore } from '../../zustand/useCoursePhaseStore'
 import { useUpdateCoursePhaseMetaData } from '@/hooks/useUpdateCoursePhaseMetaData'
-import { DragDropContext, Droppable, Draggable } from '@hello-pangea/dnd'
+import { DragDropContext, Droppable, Draggable, type DropResult } from '@hello-pangea/dnd'
 import {
   DeleteConfirmation,
   Input,
@@ -72,14 +72,16 @@ export const QuestionConfiguration = () => {
   }
 
   const deleteQuestion = () => {
-    if (!toBeDeletedQuestionID) return
-    const newQuestions = interviewQuestions.filter((q) => q.id !== toBeDeletedQuestionID)
+    if (toBeDeletedQuestionID == null) return
+    const newQuestions = interviewQuestions
+      .filter((q) => q.id !== toBeDeletedQuestionID)
+      .map((q, idx) => ({ ...q, orderNum: idx }))
     setInterviewQuestions(newQuestions)
     debouncedSave(newQuestions)
     setToBeDeletedQuestionID(undefined)
   }
 
-  const onDragEnd = (result: any) => {
+  const onDragEnd = (result: DropResult) => {
     if (!result.destination) return
 
     const newQuestions = Array.from(interviewQuestions)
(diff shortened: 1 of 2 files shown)


mathildeshagl:
@coderabbitai please resolve all the conversations that have been addressed.

## Resolved

The following comments have been implemented in commit `a5b104d0`:

1. **`deleteQuestion` falsy-check bug + missing `orderNum` reindex** — Changed `if (!toBeDeletedQuestionID)` to `if (toBeDeletedQuestionID == null)` to correctly handle edge-case ID `0`, and added `.map((q, idx) => ({ ...q, orderNum: idx }))` after the filter so question indices are always sequential after deletion.

2. **Type `onDragEnd` result with `DropResult`** — Replaced `result: any` with `result: DropResult` (imported from `@hello-pangea/dnd`) for full type safety on drag event fields.

3. **Optimistic update matching wrong field** — Changed `participant.id === newParticipationData.courseParticipationID` to `participant.courseParticipationID === newParticipationData.courseParticipationID`, which is the correct field on `CoursePhaseParticipationWithStudent`.

4. **Replace loose `ParticipantData` interface** — Removed the ad-hoc `ParticipantData` interface (with `[key: string]: any`) and replaced it with `CoursePhaseParticipationWithStudent` from `@tumaet/prompt-shared-state`, restoring compile-time type safety.

5. **Add `onSettled` for cache re-sync** — Moved `queryClient.invalidateQueries` from `onSuccess` to `onSettled` so the cache is re-synced with the server after both success and error (ensuring a stale rollback snapshot does not persist).

## Not implemented (don't make sense / no longer applicable)

- **`isInitialLoadRef` / `queueMicrotask` guard (repeated across reviews)** — The current implementation does *not* use a second `useEffect` that watches `interviewQuestions`. Saves are triggered *explicitly* by user action handlers (`addQuestion`, `deleteQuestion`, `onDragEnd`), not by state changes. There is therefore no save-on-load race condition in the current code, and the `queueMicrotask` / ref-inversion suggestion does not apply.

- **`eslint-disable-next-line` comment (repeated across all review rounds)** — Already present in the code before these commits: the `// eslint-disable-next-line react-hooks/exhaustive-deps` comment is on the line above `}, [coursePhase?.id])`.

- **Remove unused `useCallback` import** — `useCallback` is not imported in the current file; this was already resolved in an earlier commit before these reviews.

PR #1439, opened 2026-03-28
`Core`: External Data export integration + Deletion routine 

JGStyle:
## ✨ What is the change?

This PR is the second step of the full gdpr data export.

On the core server:

- It implements the core functionality for calling the course phase microservices to let them export the requesting subject's data. Every request is performed in int's own go routine.
- Status is tracked in the corresponding databse entries. here a new state for no_data was added to support representing the state where services handled the request fine but there was no data stored about the requested subject on the server
- Adds a go deletion routine that runs once a day. It checks for exports that are past the expiry date and deletes the corresponding S3 Entries. the records of the actual export itself are never deleted for admin and audit reasons. This includes all meta data information such as whether the user has downloaded an inidivudal item and which one (if any) failed
- adds the `/api/auth/subject_identifier` route, which returns the corresponding subject identifiers for the requester identified by it's token. This is used by the Subject Identifier middleware and central to ensuring that you can only request the export for yourself.

This pr depended on https://github.com/prompt-edu/prompt-sdk/pull/65, which introduced the Subject Identifiers Middleware. This dependency is now resolved → prompt-sdk is being bumped to 1.1.8

https://github.com/orgs/prompt-edu/discussions/683

| Description      | Screenshot         |
| A finished export with only successful items. the new blue state indicates a successful export with no data found |  2026-03-28 at 11 26 31@2x  |
| An export with failed items | 2026-03-28 at 11 23 35@2x |
| Admin Privacy Audit Dashboard, will also show export requests in the future | 2026-03-28 at 11 27 52@2x |

Full export workflow in action:
2026-05-24 at 14 10 09

---
Note: This pr used to also contain the implementation of the individual export steps on the course phase microservices, but they have now been split off into their own PR's to make it more easy to review:
#1651, #1652, #1653, #1654, #1655, #1656

## ✅ PR Checklist

- [!] Tests added or updated (if needed)
- [!] Documentation updated (if relevant)

## Summary by CodeRabbit

* **New Features**
  * Privacy Audit: tabbed Deletion/Export UI, inline export request flow with clearer confirmation and availability status popover.
  * Export list: status filter, loading state, documents grouped into collapsible sections (Ready / No data / Failed), improved document cards and “Download All” showing count/size and adjusted download behavior.
  * Background cleanup: expired/archived exports are removed automatically; new status states supported.

* **Bug Fixes**
  * Export polling, request/button states, and download access now handle no-data, completion, failure, and archived cases more reliably.


JGStyle:
updated export confirmation dialogue

- Added status indicator to show right away if some services are unavailable
- added some spacing, icon, separator to make it more readable.

 image
2026-05-17 at 14 04 02@2x

JGStyle:
example exports with the latest implementations from this PR (all data from local setup, only test data):

- Assessment: [prompt_61147f_Assessment.zip](https://github.com/user-attachments/files/27903494/prompt_61147f_Assessment.zip)
- Certificate: [prompt_61147f_Certificate.zip](https://github.com/user-attachments/files/27903495/prompt_61147f_Certificate.zip)
- Team Allocation: [prompt_61147f_Team Allocation.zip](https://github.com/user-attachments/files/27903497/prompt_61147f_Team.Allocation.zip)

JGStyle:
@coderabbitai review

rappm:
@JGStyle I would strongly prefer having this in several pull request for the different phases / modules. Otherwise this PR will grow more and more in size.

JGStyle:
@rappm  that's fair. i'll split it up

clients/core/src/managementConsole/pages/AdminPrivacyPage.tsx
```diff
@@ -1,46 +1,89 @@
...
               columns={adminExportColumns}
+              actions={[
+                {
+                  label: 'Delete',
+                  icon: <Trash2 className='w-4 h-4' />,
+                  onAction: async (rows) => {
+                    await Promise.all(rows.map((r) => deleteExport(r.id)))
+                    queryClient.invalidateQueries({ queryKey: ['privacy', 'admin', 'exports'] })
+                  },
+                  hide: (rows) => rows.every((r) => r.status === ExportStatus.archived),
```
rappm: AI finding: 
hide: (rows) => rows.every((r) => r.status === ExportStatus.archived), 
If a mixed selection of archived and non-archived rows is made, "Delete" is shown and clicked — the backend returns 400 for already-archived rows. The Promise.all swallows individual failures silently. Either filter out archived rows before calling deleteExport, or surface partial failures.

clients/core/src/managementConsole/shared/components/Privacy/PrivacyServiceAvailability.tsx
```diff
@@ -0,0 +1,120 @@
...
+  Good = 'good',
+  Problem = 'problem',
+}
+
+const GreenDot = <span className='h-2 w-2 rounded-full shrink-0 bg-green-500' />
+const RedDot = <span className='h-2 w-2 rounded-full shrink-0 bg-destructive' />
+const OrangeDot = <span className='h-2 w-2 rounded-full bg-orange-500 shrink-0' />
+
+function ServiceList({
+  eles,
```
rappm: non-descriptive param name, give this a variable name (e.g. services) matching this components purpose

servers/core/privacy/privacyDTO/privacy_export.go
```diff
@@ -19,48 +19,72 @@ type PrivacyExportDocument struct {
...
-	if model.StudentID.Valid {
-		id, _ := uuid.FromBytes(model.StudentID.Bytes[:])
-		studentID = &id
+func GetAdminPrivacyExportDTOFromDBModel(model db.GetAllExportsRow) (AdminPrivacyExport, error) {
+	rawDocs, err := json.Marshal(model.Docs)
+	if err != nil {
+		return AdminPrivacyExport{}, fmt.Errorf("failed to serialize docs field: %w", err)
+	}
+	var docs []AdminExportDoc
+	if err := json.Unmarshal(rawDocs, &docs); err != nil {
```
rappm: rawDocs, err := json.Marshal(model.Docs)       // interface{} → []byte
json.Unmarshal(rawDocs, &docs)                  // []byte → []AdminExportDoc

Even though this is a valid workaround with sqlc returing interface{} for JSON columns, I would try to improve it. Maybe you can adapt the underlaying SQL to get a more meaningful outcome for sqlc. 
This way any JSON structure that doesn't match AdminExportDoc would silently produce zero-value structs rather than an error. Consider using pgx's scanning with a concrete type directly, or at least add field validation after unmarshal.

servers/core/privacy/router_test.go
```diff
@@ -0,0 +1,429 @@
...
+}
+
+func (suite *RouterTestSuite) TestRouterPostDataExport_ValidExportAlreadyExists() {
+	// Alice already has a valid export (valid_until: 2030)
+	router := routerForUser(aliceUserID)
+	req, _ := http.NewRequest("POST", "/api/privacy/data-export", nil)
+	w := httptest.NewRecorder()
+	router.ServeHTTP(w, req)
+
+	assert.Equal(suite.T(), http.StatusInternalServerError, w.Code)
```
rappm: I wouldn't look for an InternalServerError if the Export already exists, I would prefer maybe 409.

servers/core/privacy/router_test.go
```diff
@@ -0,0 +1,429 @@
...
+}
+
+func (suite *RouterTestSuite) TestRouterPostDataExport_RateLimited() {
+	// Dave's export expired but he is still within the 30-day rate limit window
+	router := routerForUser(daveUserID)
+	req, _ := http.NewRequest("POST", "/api/privacy/data-export", nil)
+	w := httptest.NewRecorder()
+	router.ServeHTTP(w, req)
+
+	assert.Equal(suite.T(), http.StatusInternalServerError, w.Code)
```
rappm: Same here, use an appropriate status code like 429

servers/core/db/query/privacy_export.sql
```diff
@@ -37,11 +40,42 @@ SELECT * FROM privacy_export WHERE user_id = $1 ORDER BY date_created DESC LIMIT
...
+SELECT * FROM privacy_export WHERE now() >= valid_until AND status != 'archived' LIMIT $1;
+
+-- name: GetExportDocObjectKeysByExportID :many
+SELECT object_key FROM privacy_export_document WHERE export_id = $1;
+
+-- name: SetExportDocStatusByExportID :exec
+UPDATE privacy_export_document SET status = $2 WHERE export_id = $1;
+
+-- name: ArchiveCompletedExportDocs :exec
+UPDATE privacy_export_document SET status = 'archived' WHERE export_id = $1 AND status = 'complete';
```
rappm, line 77: AI Comment: When the deletion routine archives an export, no_data documents stay in no_data status. This is likely unintentional — an archived export should have all its docs in a terminal state. This makes the getExportDocObjectKeysByExportID/S3 cleanup logic correct (those docs have no files), but the admin view will show some archived exports with mixed statuses.

rappm (changes requested):
Already looks really good, especially the high test coverage.  Mostly some minor changes with an archiving overlook and incorrect status codes. The rest is mostly polish items.

JGStyle:
> Already looks really good, especially the high test coverage. Mostly some minor changes with an archiving overlook and incorrect status codes. The rest is mostly polish items.

Thanks! I'm working on getting the feedback ready

clients/core/src/managementConsole/pages/AdminPrivacyPage.tsx
JGStyle: fixed in latest commit, i switched from Promise.all to Promise.allSettled, i still want to keep the bulk deletion, but that way it will simply wait for all requests to finish, regardless of status code, so those already deleted won't block the rest (visible once pushed, will do with the others)

clients/core/src/managementConsole/shared/components/Privacy/PrivacyServiceAvailability.tsx
JGStyle: fair, not a good name. fixed in latest commit, visible once pushed

servers/core/privacy/privacyDTO/privacy_export.go
JGStyle: I agree that's the better fix. i even realized that but i had already deployed to dev once which makes it annoying for me because i don't have access. but longterm speaking it's cleaner to just have the right type 'annotation' right away. I added it in the commit, but i don't want to add another migration file for that → Maybe you could run this sql code on the dev server once? :
(just drops the old view and inserts the new one with the correct type checks)
```sql
  DROP VIEW IF EXISTS privacy_export_with_docs;

  CREATE VIEW privacy_export_with_docs AS
  SELECT
    e.id,
    e.user_id,
    e.student_id,
    e.status,
    e.date_created,
    e.valid_until,
    e.next_request_allowed_at,
    COALESCE(
      jsonb_agg(
        jsonb_build_object(
          'id', ed.id,
          'date_created', ed.date_created,
          'source_name', ed.source_name,
          'status', ed.status,
          'file_size', ed.file_size,
          'downloaded_at', ed.downloaded_at
        ) ORDER BY ed.date_created ASC
      ) FILTER (WHERE ed.id IS NOT NULL),
      '[]'::jsonb
    )::jsonb AS documents
  FROM privacy_export e
  LEFT JOIN privacy_export_document ed ON ed.export_id = e.id
  GROUP BY e.id;

```

edit nevermind seems to work anyways because it's still jsonb, just now known at compilation

servers/core/privacy/router_test.go
JGStyle: Fair, i adjusted it. in the same go i made the validations more explicit, they are no longer in the creation method but separate. I think this is better because the router method now clearly reads what the conditions are under which the request is rejected. also simplifies the creation method.

servers/core/privacy/router_test.go
JGStyle: adjusted it together with the other one

servers/core/db/query/privacy_export.sql
JGStyle: this is actually intentional: originally i also set all of them to archived but then i would lose the info on which parts of the export failed and succeded, which is good to have because otherwise you can't f.e. give a good answer for the downloaded column (because of course a user could have never downloaded a failed export, but then we don't know anymore which one failed)

rappm (approved):

PR #1663, opened 2026-06-05
`CI`: Add sqlc generation check to server quality workflow

### LINKED ISSUE #1660: Include SQLC check in server quality pipeline

rappm:
# 📝 Pull Request Template

## ✨ What is the change?

Adds a sqlc generated-code drift check to the reusable server quality workflow, and fixes a change-detection bug in the CI pipeline that caused `team_allocation` to be silently excluded from quality and test stages.

**sqlc drift check:**
The workflow now installs `sqlc` pinned to `v1.31.1`, runs `sqlc generate` for each server matrix module, and fails when generated files under `db/sqlc` would change.

**Change detection fix (`detect-changes.yml`):**
The force-all path (triggered when workflow files change) used `for dir in servers/*/` + `[ -f "$dir/go.mod" ]` to enumerate server modules. The trailing slash in the glob expansion (`servers/team_allocation/`) produced a double-slash path (`servers/team_allocation//go.mod`), which silently failed the `[ -f ]` check on the GitHub Actions runner. Because bash's `set -e` does not abort on the left-hand side of `&&`, the loop skipped `team_allocation` and continued — leaving it out of the matrix for every CI run on this PR.

Fixed by replacing both force-all loops (`server_modules` and `api_spec_directories`) with `find servers -maxdepth 2 -name "go.mod" | sort`, which directly discovers modules without relying on glob expansion or a separate file-existence check.

Related to #1660.

Prevents SQL, migration, or sqlc configuration changes from being merged without committing the corresponding generated Go files. The detection fix ensures all server modules (including `team_allocation`) are correctly included in quality and test stages when CI workflow files change.

## 🧪 How to Test

1. Inspect `.github/workflows/quality-servers.yml` — confirms `sqlc@v1.31.1` install and drift check.
2. Inspect `.github/workflows/detect-changes.yml` — confirms force-all path now uses `find` to enumerate server modules.
3. Confirm all 7 server modules (including `team_allocation`) appear in the quality and test matrix on the next CI run.

Local validation: `git diff --check`.

N/A - CI workflow change only.

## ✅ PR Checklist

## Summary by CodeRabbit

* **Chores**
  * CI now pins the sqlc tool per server module and verifies generated DB code matches committed files to prevent drift.

* **Changes**
  * Template lookup now returns richer template records including course-phase identifiers.
  * Several team-allocation DB access operations were removed, altering some team retrieval behaviors and removing a legacy test query.


bensofficial (dismissed):
The changes look good to me. Just one question about your CI setup: You only updated the sqlc files for all server modules. That triggered tests and linting for all modules except `team_allocation`, even though that one was also updated. I took a quick look at the workflow files and I don't think that's actually how it's supposed to work, is it?

rappm:
@bensofficial thanks for the review! Really good catch - wasn't aware the detect changes was broken before. "team-allocation" was missing in the module input. Fixed now.

PR #1685, opened 2026-06-15
fix alertdialog fly in animation under tailwind v4

### LINKED ISSUE #1684: Fix Data Export Animation
The data export dialog flies in weirdly. We do not have this animation anywhere else.

https://github.com/user-attachments/assets/c8f4a36f-90a6-4776-a8fe-2b5e45cc934e

JGStyle:
## ✨ What is the change?

the Tailwind v4 upgrade is the root cause. Tailwind v3's translate-x-* utilities emitted legacy
  transform: translate(...), which lived in the same property as the animate plugin's keyframes, so
  the animation simply overrode the static centering and everything composed correctly. v4 switched
  those utilities to the modern translate: property, but the animation library wasn't updated in
  lockstep — so the static centering and the entry animation now write to different properties,
  compose instead of overriding, and that's the off-screen fly-in.

Claude found this fix for #1684, however I think it's rather ugly

This fixes #1684

## 🧪 How to Test

Open any alertdialog from promp-ui-components and notice the animation is back to normal (no fly in from top left)
For example open the export request

only animation is changed. check #1684 to see faulty animation

## ✅ PR Checklist

- [!] Code is clean, readable, and documented


JGStyle:
I think it's a really ugly fix, but it works. not sure if this is worth it - we would also have to remove it again later if it is fixed. maybe we just wait until either radix introduces a fix themselves or any of the other packages in this dependency chain?
@mathildeshagl

mathildeshagl:
puh, weiß nicht ob 12.000 Zeilen code change den animation fix wert sind 😅

PR #1710, opened 2026-06-22
`Core`: UI for managing Lecturers and Editors

Mtze:
## ✨ What is the change?

Adds a new **User Management** item to the course sidebar that lets a `CourseLecturer` or `PROMPT_Admin` list, add, and remove members of the Lecturer and Editor Keycloak groups for their course — without needing direct Keycloak admin access.

Keycloak remains the source of truth; PROMPT stores no group state in its own database. No new microservice, no new tables.

### Backend (`servers/core/keycloakRealmManager`)
- New routes:
  - `GET    /api/keycloak/:courseID/group/team`
  - `PUT    /api/keycloak/:courseID/group/:groupName/members/:userID`
  - `DELETE /api/keycloak/:courseID/group/:groupName/members/:userID`
  - `GET    /api/keycloak/users/search` (realm-wide, gated by `CheckAccessControlByRole`)
- `:groupName` is validated against a `{Lecturer, Editor}` allow-list **before** any Keycloak call. The Keycloak group path is derived server-side from the course UUID alone — the client never sends a group name, path, or ID.
- Target users are verified via `GetUserByID` before any mutation; 404 maps to a clean client `404`.
- Self-removal is rejected at the server with `400` (the only mechanism we need to prevent emptying the Lecturer group — the final remover can never delete themselves, so no advisory lock is required).
- New generic `GetCourseSubgroup` helper replaces the Lecturer/Editor pair; the existing `GetCourseEditorGroup` delegates to it.
- Structured audit log line on every mutating call (caller, action, target, group, course).
- Pure-logic unit tests for the allow-list, self-removal, and short-query guards.

### Latent bug fix (affects existing code paths too)

gocloak's `GetGroupByPath` joins URL segments with \`/\`, so a leading-slash \`groupPath\` produced \`group-by-path//Prompt/...\` which recent Keycloak versions reject under strict path normalization (\`missingNormalization: Request path not normalized\`). The wrapper now strips the leading slash before delegating — Keycloak itself strips it internally (\`KeycloakModelUtils.findGroupByPath\`) so behaviour is unchanged at the business-logic layer. Same pattern fixed in \`GetOrCreateCustomTopLevelGroup\` and \`GetOrCreateCustomGroup\`.

### Frontend (\`clients/core\`)
- New route \`/management/course/:courseId/user-management\` gated by \`PROMPT_ADMIN + COURSE_LECTURER\`.
- New sidebar entry next to **Settings**.
- \`CourseUserManagementPage\` with two tables (Lecturers, Editors) and an \`AddUserDialog\` with debounced Keycloak user search, \"already a member\" hints, and a \"more results available\" indicator when the response is truncated.
- The self-row's remove button is disabled with a tooltip (defence in depth — the server still enforces the same rule via the JWT \`sub\`).
- TanStack Query invalidates both the team and the user-search caches after every mutation, so membership badges update immediately.

Implements the MVP scope of discussion [#700](https://github.com/orgs/prompt-edu/discussions/700). Today, adding or removing course Lecturers/Editors requires direct Keycloak admin console access — this blocks routine course-team management on anyone without Keycloak privileges and creates an unnecessary operational dependency.

Out of scope (deliberately deferred to future PRs):
- Custom-group management under \`CustomGroups/*\` (discussion section 2). The \`:groupName\` allow-list is the only choke point that would need to change.
- Token caching in the realm manager (discussion section 6).
- Restricting the searchable user pool to \"users who have logged into PROMPT at least once\" (discussion section 3).
- Persisting the audit log to PROMPT's DB (currently only logged via logrus).

## 🧪 How to Test

1. \`make db && make server && make clients\`.
2. Log in as a course Lecturer; open any course you're a Lecturer of.
3. Confirm the new **User Management** entry appears in the course sidebar next to **Settings**.
4. Navigate to it: both Lecturer and Editor tables should render with the current Keycloak state.
5. Click **Add Editor**, type at least 2 characters into the search input, and verify results come back. Add a user; verify the table refreshes and Keycloak admin console reflects the change.
6. Repeat with **Add Lecturer**.
7. Remove a non-self Editor and Lecturer; confirm the UI updates and the toast shows.
8. Try to remove yourself: the trash button should be disabled with a tooltip. Optional: \`curl -X DELETE\` the API directly with your own \`userID\` to confirm the server rejects with \`400\`.
9. Authorization check: log in as a \`CourseEditor\` of the same course — sidebar item is hidden, and direct URL access shows the unauthorized page. Try a curl from a Lecturer of a **different** course — expect \`403\`.

## ✅ PR Checklist

- [x] Tested locally
- [x] Tests added (pure-logic unit tests; Keycloak interactions verified manually since there's no testcontainers Keycloak harness in this repo)
- [ ] Screenshots attached for UI changes
- [ ] Documentation updated (none needed — the docs site doesn't currently describe instructor-management flows)

## Summary by CodeRabbit

* **New Features**
  * Added a Course **User Management** page for managing **Lecturers/Editors**, including Keycloak user search, add, and remove (with confirmation), with a new permission-restricted route and sidebar navigation.
  * Added a **Keycloak** status/probe card on the System Status page with per-capability checks and manual refresh.
* **Bug Fixes**
  * Fixed Keycloak group path normalization to avoid double-slash path issues.
* **Documentation**
  * Updated Keycloak service-account setup guidance and added user management + course team API/security documentation.
* **Tests**
  * Added backend unit tests covering staff management validation and Keycloak search query rules.


servers/core/keycloakRealmManager/service.go
```diff
@@ -151,6 +177,175 @@ func GetStudentsInGroup(ctx context.Context, courseID uuid.UUID, groupName strin
...
+
+	group, err := GetCourseSubgroup(ctx, token.AccessToken, courseID, groupName)
+	if err != nil {
+		return fmt.Errorf("failed to resolve %s group: %w", groupName, err)
+	}
+
+	if err := KeycloakRealmSingleton.client.DeleteUserFromGroup(ctx, token.AccessToken, KeycloakRealmSingleton.Realm, targetUserID, *group.ID); err != nil {
+		log.Error("failed to remove user from group: ", err)
+		return fmt.Errorf("failed to remove user from group: %w", err)
+	}
```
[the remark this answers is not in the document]
Mtze: Not applying this one. The add path verifies the target exists via `GetUserByID` because adding a non-existent user would silently create a stale group membership. Remove is the inverse - it unsticks whatever's there - and the cost vs. benefit is different:

- Doubles the Keycloak round-trips for every remove just to translate a missing-user case into a clean `ErrUserNotFound` instead of letting the `DeleteUserFromGroup` error surface naturally.
- The edge case (someone tries to remove a user whose Keycloak account was deleted between the page load and the click) is rare and recoverable - the user clicks refresh and the row is gone.

If we ever start seeing actual 500s from this path in the audit logs, we can revisit.

Mtze:
Triaged the CI failure and the bot feedback. Summary of what landed (commits `c81c9dac` and `f5887eb5`):

### CI failure: `quality-servers / quality-server (core)` — fixed
Root cause: this branch was based on a snapshot that still had `gocloak/v13`. Commit ec1d8d7f on main migrated everything to `gocloak/v14` and dropped v13 from `go.mod`. When CI merges this PR into main, v13 disappears from `go.mod` but the new code still imports it -> `golangci-lint` typecheck failure. Migrated all imports in this package to v14 and dropped v13 from `go.mod`. `go build`, `go vet`, and the package tests are green.

### CodeRabbit feedback
- `TeamMemberTable.tsx`: added an `sr-only "Actions"` label to the empty `TableHead`. ✅
- `router.go`: added a defence-in-depth empty-string check on `callerUserID` for both the add and remove handlers. Returns `401` if the auth middleware didn't populate it. ✅
- `CourseUserManagementSidebar.tsx`: extracted the inline prop type literal into a named `interface`. ✅
- `service.go` (remove-path pre-flight `GetUserByID`): not applying — reasoning posted as an inline reply on that thread.

### `github-code-quality`
- `CourseUserManagementPage.tsx`: dropped the redundant `open={dialogGroup !== null}` inside the `{dialogGroup && (...)}` guard. ✅

### Pre-merge: docstring coverage (warning, 62.07% vs 80%)
Added missing doc comments to the exported `TeamMember`, `CourseTeam`, `UserSearchResults`, `GetTeamMemberFromKeycloakUser`, `GetCourseEditorGroup`, and the `Capability*` constants. New code should now be above the threshold.

servers/core/docs/swagger.json
```diff
@@ -3172,6 +3172,71 @@
...
+                    },
+                    "500": {
+                        "description": "Internal Server Error",
+                        "schema": {
+                            "$ref": "#/definitions/utils.ErrorResponse"
+                        }
+                    }
+                }
+            }
+        },
```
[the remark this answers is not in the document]
rappm: this is done nowhere in prompt, save that in your memories

servers/core/keycloakRealmManager/service.go
```diff
@@ -151,6 +177,239 @@ func GetStudentsInGroup(ctx context.Context, courseID uuid.UUID, groupName strin
 	}, nil
 }
 
+// GetCourseStaff returns the Lecturer and Editor members of a course.
+func GetCourseStaff(ctx context.Context, courseID uuid.UUID) (keycloakRealmDTO.CourseStaff, error) {
```
rappm, line 185: AI Finding of my error on prompt-dev: 
This is where the deployed "couldn't load staff entries" failure surfaces.**

`GetCourseStaff` returns a hard error (→ HTTP 500) whenever *either* `getCourseGroupMembers` call fails, and the client only renders a generic `Failed to load course staff.` with no detail. On prompt-dev this is almost certainly the `prompt-server` service account missing `view-users` / `view-groups` (see `docs/admin/keycloak-prod.md` §9 — diagnosable via the new **System Status → Keycloak** card this PR adds).

Two things worth hardening:
1. Propagate the underlying Keycloak error to the client so a misconfigured service account is self-explanatory instead of a blank page.
2. Consider treating a *group-not-found* from `GetCourseSubgroup` as an empty member list, so a course missing one subgroup (legacy / manually-created) still renders the group that does exist instead of 500-ing both tables. Keep genuine permission/transport errors as 500.

clients/core/src/managementConsole/courseUserManagement/pages/CourseUserManagementPage.tsx
```diff
@@ -0,0 +1,95 @@
...
+import { useRemoveCourseStaffMember } from '../hooks/useRemoveCourseStaffMember'
+import { StaffMemberTable } from '../components/StaffMemberTable'
+import { AddUserDialog } from '../components/AddUserDialog'
+import { CourseGroupName, StaffMember } from '../interfaces/StaffMember'
+
+export const CourseUserManagementPage = () => {
+  const { courseId } = useParams<{ courseId: string }>()
+  const { user } = useAuthStore()
+
+  const { data, isLoading, isError, refetch } = useCourseStaff(courseId)
```
rappm: Show the real error to the user. Capture `error` from the query so the ErrorPage below can show the server's message.

```suggestion
  const { data, isLoading, isError, error, refetch } = useCourseStaff(courseId)
```

clients/core/src/managementConsole/courseUserManagement/pages/CourseUserManagementPage.tsx
```diff
@@ -0,0 +1,95 @@
...
+  )
+  const editorIDs = useMemo(
+    () => new Set((data?.editors ?? []).map((m) => m.keycloakUserID)),
+    [data],
+  )
+
+  if (!courseId) return <ErrorPage onRetry={() => undefined} description='No course selected.' />
+  if (isLoading) return <LoadingPage />
+  if (isError || !data) {
+    return <ErrorPage onRetry={() => refetch()} description='Failed to load course staff.' />
```
rappm: A misconfigured service account currently shows only `Failed to load course staff.`, also reproducable on prompt-dev once deployed; passing the backend error through would have made the deploy failure self-diagnosing.

```suggestion
    return (
      <ErrorPage
        onRetry={() => refetch()}
        description={
          (error as { response?: { data?: { error?: string } } })?.response?.data?.error ??
          'Failed to load course staff.'
        }
      />
    )
```

servers/core/keycloakRealmManager/router.go
```diff
@@ -174,3 +197,164 @@ func handleError(c *gin.Context, statusCode int, err error) {
...
+// @Router /keycloak/users/search [get]
+func searchKeycloakUsers(c *gin.Context) {
+	query := c.Query("q")
+
+	limit := 20
+	if raw := c.Query("limit"); raw != "" {
+		if parsed, err := strconv.Atoi(raw); err == nil {
+			limit = parsed
+		}
+	}
```
rappm, line 332: **[pending] Minor:** a malformed `limit` (e.g. `?limit=abc`) is silently swallowed and falls back to 20, unlike the 400s the rest of this API returns. Committable suggestion to reject it (`errors` and `net/http` are already imported):

```suggestion
	limit := 20
	if raw := c.Query("limit"); raw != "" {
		parsed, err := strconv.Atoi(raw)
		if err != nil {
			handleError(c, http.StatusBadRequest, errors.New("limit must be an integer"))
			return
		}
		limit = parsed
	}
```

servers/core/keycloakRealmManager/service.go
```diff
@@ -151,6 +177,239 @@ func GetStudentsInGroup(ctx context.Context, courseID uuid.UUID, groupName strin
...
+		return keycloakRealmDTO.CourseStaff{}, err
+	}
+
+	return keycloakRealmDTO.CourseStaff{
+		Lecturers: lecturers,
+		Editors:   editors,
+	}, nil
+}
+
+func getCourseGroupMembers(ctx context.Context, accessToken string, courseID uuid.UUID, groupName string) ([]keycloakRealmDTO.StaffMember, error) {
```
rappm: **[pending] Cleanup (non-blocking):** Lecturers and Editors are fetched sequentially, and each call re-resolves the course group name via `GetCourseGroupName` (a DB round-trip) under the same login. For a 2-group page this is minor, but resolving the parent course group once and/or fetching the two member lists concurrently would halve the Keycloak round-trips. Noting, not blocking.

rappm (changes requested):

rappm:
And a merge conflict @mtze

servers/core/keycloakRealmManager/service.go
Mtze: Both addressed in a43831f5 - `getCourseGroupMembers` treats a missing Lecturer/Editor subgroup (404) as an empty list and lets the other table render; the underlying error bubbles up through `handleError` and the page now surfaces it (see comment #3497127870).

clients/core/src/managementConsole/courseUserManagement/pages/CourseUserManagementPage.tsx
Mtze: Applied in a43831f5.

clients/core/src/managementConsole/courseUserManagement/pages/CourseUserManagementPage.tsx
Mtze: Applied in a43831f5 - `ErrorPage description` now falls through to `response.data.error` when the server sends one.

servers/core/keycloakRealmManager/router.go
Mtze: Applied in a43831f5 - `?limit=abc` now returns 400 with `limit must be an integer`.

servers/core/keycloakRealmManager/service.go
Mtze: Done in 4d5bf366 - `GetCourseStaff` now resolves the course group name once and fans the Lecturer + Editor member lookups out with `errgroup`.

rappm (approved):

PR #1736, opened 2026-06-24
`Core`: Archived courses can't be accessed by their instructors

### LINKED ISSUE #1264: Archived Courses cant be accessed by their Instructors
As soon as a course is archived the instructors can 
1. no longer access the course via the link 
2. not see the course in the archive 

both should be possible for them so they can see older results

mathildeshagl:
## ✨ What is the change?

Archived courses are now visible to their instructors (Course Lecturers/Editors) again — they can open an archived course via its link and find it in the course archive view. Student-only users still do not see archived courses (unchanged behavior).

The fix is a single SQL change in `GetAllActiveCoursesRestricted`:
- Removed the `WHERE c.archived = FALSE` filter from the `user_course_roles` CTE.
- Added a `HAVING` clause that hides archived courses **only** from student-only users, reusing the query's existing student-detection logic.

No frontend changes were needed — the archive page, the link-access gate in `ManagementRoot`, and the sidebar all behave correctly once the backend returns the data.

Closes #1264.

When a course was archived, `GetAllActiveCoursesRestricted` (the course-list query for all non-admin users) excluded it with `WHERE c.archived = FALSE`. This single omission caused both reported symptoms:
- **Not in archive:** `ArchivedCoursesPage` filters the course store for `archived === true`, but the store never contained archived courses for instructors.
- **Link broken:** `ManagementRoot` gates access on the course being in the fetched list, so archived links rendered `CourseNotFound`.

Admins were unaffected because they use a different, unfiltered query — which is why the issue couldn't be reproduced by the (admin) reporter.

## 🧪 How to Test

1. `cd servers/core && go test ./course/...` — passes, including new tests `TestGetAllCoursesArchivedVisibleToInstructor` and `TestGetAllCoursesArchivedHiddenFromStudent`.
2. Manual (optional): as a Course Lecturer (non-admin), archive a course, then:
   - Open `…/management/course-archive` → the course appears.
   - Open `…/management/course/<archived-course-id>` directly → the course loads instead of "Course not found".
   - As a student of that course, confirm it does **not** appear.

## ✅ PR Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated course listing behavior so archived courses are visible to instructors and other eligible roles, while remaining hidden from student-only access.
  * Adjusted the “Template Courses” sidebar visibility rules to match the intended role combinations.
* **Tests**
  * Updated course listing expectations to include the newly added archived course.
  * Added test coverage for archived visibility for instructor vs. student-only users.
* **Data**
  * Updated course database dump to include an additional archived course record.


Mtze:
I tested this change and noticed that we have a second problem with our current implementation. Instructors can only see the Course Archive if the have the "Prompt Lecturer" role in keycloak - I think we also have to change that for this feature to make sense. All instructors of courses should see their old/archived courses - independent of the ability to create new ones.

Mtze (changes requested):
See comment above

Mtze (approved):

PR #1774, opened 2026-07-01
`General`: Migrate to biome

### LINKED ISSUE #1667: Switch from eslint to biome

magkue:
# Switch from eslint to biome

## ✨ What is the change?

We use biome instead of eslint

Biome is better.
## 🧪 How to Test

1.  Run `cd clients && yarn install && yarn biome check .`

## ✅ PR Checklist


rappm (approved):
Reviewed the ESLint→Biome migration. Rule mapping is faithful (noShadow, hooks rules, exhaustive-deps, prettier→formatter), and Biome's stricter noExplicitAny is an improvement over the old `no-explicit-any: off`. The unsafe autofixes that were applied (`==`→`===`, `!!x`→`x` in boolean contexts, useless-ternary, literal keys, template literals) are all semantically equivalent on the operands involved, and side-effect/CSS import ordering is preserved. Resolved the merge conflict in AssessmentHeader.tsx: kept main's `md:inline` breakpoint but used optional chaining instead of the non-null assertions to avoid new Biome warnings (equivalent at runtime, since the block is guarded by hasNavigation). Typecheck and `biome check` pass on the resolved file.

PR #1835, opened 2026-07-03
`Assessment`: Validate grade suggestion range on the server and in the database

### LINKED ISSUE #1778: Grade suggestion range validated only client-side (no server/DB bounds check)
## Summary

The grade suggestion range (1.0-5.0) is enforced only in the React client. The server maps any `float64` straight into the column with no validation and no DB `CHECK` constraint, so a direct API call can store out-of-range grades.

## Location

- Service: `servers/assessment/assessments/assessmentCompletion/service.go:137` maps `req.GradeSuggestion` via `utils.MapFloat64ToNumeric` with no bounds check.
- DTO: `assessmentCompletion/assessmentCompletionDTO/assessmentCompletion.go` declares `GradeSuggestion float64` with only a `json` tag (no binding validation).
- Client-only validation: `clients/assessment_component/src/assessment/pages/utils/gradeConfig.ts:44`.
- No `CHECK` constraint exists in any migration.

## Impact

Any direct API call (or a second consumer such as the export / inter-phase communication path) can store `gradeSuggestion = -3`, `0`, or `9.9`. Downstream grade statistics, the core phase-communication endpoint, and transcripts consume it unchecked.

Note: the column is `numeric(2,1)` (`db/migration/0010_adapt_assessment_schema.up.sql:8`), so values that exceed precision (e.g. `42`) currently fail as a `500` rather than being stored, but any in-precision out-of-range value (`-3`, `0`, `9.9`) is accepted.

## Suggested fix

Validate the grade server-side (reject values outside the allowed discrete set / range in the DTO binding or service) and add a DB `CHECK` constraint as defense in depth. Consider centralizing the allowed-grade definition so client and server agree.

rappm:
# 📝 Pull Request Template

## ✨ What is the change?

The grade suggestion range was enforced only in the React client. This adds two server-side guards:

- **Service validation:** `CreateOrUpdateAssessmentCompletion` now rejects any `gradeSuggestion` outside `1.0`–`5.0`, returning `400 Bad Request` (via a new `ErrInvalidGradeSuggestion`).
- **DB `CHECK` constraint (defense in depth):** new migration `0025_grade_suggestion_bounds` adds `CHECK (grade_suggestion >= 1.0 AND grade_suggestion <= 6.0)` on `assessment_completion`. The upper bound is `6.0` (not `5.0`) so the constraint tolerates the legacy `6.0` "no grade" default/backfill introduced in migration `0010` and cannot fail on existing rows. The same constraint is added to the test dump (`database_dumps/assessmentCompletions.sql`).

Closes #1778.

The server mapped any `float64` straight into the `numeric(2,1)` column with no bounds check and no DB constraint, so a direct API call could store out-of-range grades such as `-3`, `0`, or `9.9`. Downstream grade statistics, the core phase-communication endpoint, and transcripts consumed the value unchecked. `CreateOrUpdateAssessmentCompletion` is the only path that writes `grade_suggestion`, so validating it there plus the `CHECK` constraint closes the gap regardless of the caller.

## 🧪 How to Test

1. `cd servers/assessment && go test ./...` — all suites pass, including the new `TestCreateOrUpdateAssessmentCompletionRejectsOutOfRangeGrade` (rejects `-3`, `0`, `0.9`, `5.1`, `9.9` and stores nothing).
2. `POST /api/course_phase/{coursePhaseID}/student-assessment/completed` with `"gradeSuggestion": 9.9` → `400 Bad Request` with `grade suggestion must be between 1.0 and 5.0`.
3. A valid in-range value (e.g. `2.3`) still succeeds.

N/A — backend and database only.

## ✅ PR Checklist


servers/assessment/db/migration/0025_grade_suggestion_bounds.up.sql
```diff
@@ -0,0 +1,8 @@
+BEGIN;
+
+-- ponytail: API rejects >5.0; DB tolerates the legacy 6.0 "no grade" default/backfill from 0010 so the constraint can't fail on existing rows
+ALTER TABLE assessment_completion
+    ADD CONSTRAINT assessment_completion_grade_suggestion_check
```
mathildeshagl, line 6: `ADD CONSTRAINT … CHECK` (no `NOT VALID`) validates every existing row at apply time. Since the whole premise is that the old path could store `-3`/`0`/`9.9`, one such row will fail `migrate up` and block server startup. Use `NOT VALID` (+ later `VALIDATE CONSTRAINT`) or a pre-clamp `UPDATE`, unless you can confirm no out-of-range rows exist.

servers/assessment/db/migration/0025_grade_suggestion_bounds.up.sql
```diff
@@ -0,0 +1,8 @@
+BEGIN;
+
+-- ponytail: API rejects >5.0; DB tolerates the legacy 6.0 "no grade" default/backfill from 0010 so the constraint can't fail on existing rows
```
mathildeshagl: Stray `-- ponytail:` marker — drop the prefix, keep the explanation.

servers/assessment/database_dumps/assessmentCompletions.sql
```diff
@@ -92,7 +92,8 @@ CREATE TABLE public.assessment_completion (
     author text NOT NULL,
     comment text DEFAULT '' NOT NULL,
     grade_suggestion numeric(2, 1) DEFAULT 4.0 NOT NULL,
-    completed boolean DEFAULT false NOT NULL
+    completed boolean DEFAULT false NOT NULL,
+    CONSTRAINT assessment_completion_grade_suggestion_check CHECK (grade_suggestion >= 1.0 AND grade_suggestion <= 6.0)
```
mathildeshagl, line 96: Constraint added only here; `assessment_completion` is also defined in `actionItem.sql` and `assessments.sql`, which now diverge from the prod schema and don't exercise the check. Mirror it there for fidelity.

mathildeshagl (changes requested):
Solid two-layer guard, and the client already restricts to 1.0–5.0 so no UI regression. One blocking concern on the migration plus two minor nits inline.

mathildeshagl (changes requested):
Thanks for addressing the earlier points — `NOT VALID`, the comment cleanup, and mirroring the CHECK into all three dumps all look good, and the dump/db-level seeds are all in range.

One follow-up (couldn't anchor inline — the line is unchanged, outside the diff):

**`servers/assessment/assessments/assessmentCompletion/service_test.go:215` — `TestCreateOrUpdateAssessmentCompletionWithInvalidData` is now vacuous.** It seeds `GradeSuggestion: 0.0` to exercise the nil-UUID / DB-constraint path, but the new grade guard at the top of `CreateOrUpdateAssessmentCompletion` now returns `ErrInvalidGradeSuggestion` before that path is reached. The tolerant assertion keeps it green, so it no longer tests what its name claims. Bump the seed to an in-range value (e.g. `4.0`) so the nil-UUID scenario is actually exercised.

mathildeshagl (approved):

PR #1839, opened 2026-07-05
`Assessment`: Read data from React Query instead of mirror zustand stores

### LINKED ISSUE #1780: DataShell components mirror React Query into zustand stores (duplicate source of truth)
## Summary

`EvaluationDataShell` and `AssessmentDataShell` mirror React Query results into ~7 zustand stores via ~12 hand-maintained `useEffect`s, creating a second source of truth that lags refetches and requires a store + setter + effect per dataset.

## Location

- `clients/assessment_component/src/assessment/pages/EvaluationDataShell.tsx:163-245` — 12 `useEffect` blocks feeding 7 stores (`useCoursePhaseConfigStore`, `useTeamStore`, `useMyParticipationStore`, `useSelfEvaluationCategoryStore`, `usePeerEvaluationCategoryStore`, `useTutorEvaluationCategoryStore`, `useEvaluationStore`).
- `clients/assessment_component/src/assessment/pages/AssessmentDataShell.tsx:135-185` — same pattern, 8 effects.

## Cost

- Two sources of truth per dataset. After a mutation invalidates a query, the store holds stale data until the shell re-renders and the effect re-runs (a one-render lag while mounted; indefinite if a consumer reads the store while the shell is unmounted).
- Every new dataset requires a new store, a setter, and a `useEffect` with a hand-maintained dependency array; drift there silently freezes UI data.

## Suggested fix

Have consumers read the React Query hooks directly (the query key already dedupes fetches), or expose the shell's loaded data through a single React context, removing the per-dataset stores and their sync effects. This pairs naturally with the self/peer/tutor store consolidation.

rappm:
# 📝 Pull Request Template

## ✨ What is the change?

Removes the duplicate source of truth in the assessment component. `EvaluationDataShell` and `AssessmentDataShell` used to fetch every dataset with React Query and then mirror each result into ~10 zustand stores via ~22 hand-maintained `useEffect`s. Consumers read the stores, so after a mutation invalidated a query the store held stale data until the shell re-rendered and the effect re-ran.

- Consumers now read the React Query hooks directly. The data is already cached by the shells, so reads hit the same cache with no extra fetch and reflect mutations immediately.
- The shells stay as pure loading/error gates: they still fetch to gate rendering, but no longer own store state or sync effects.
- The 10 mirror zustand stores are deleted (`useCoursePhaseConfigStore`, `useTeamStore`, `useParticipationStore`, `useCategoryStore`, `useScoreLevelStore`, `useMyParticipationStore`, `useEvaluationStore`, and the self/peer/tutor category stores). `useStudentAssessmentStore` and `useStudentEvaluationStore` (genuine client-side UI state) are kept.
- The shell-shared query hooks were hardened to make the direct-read pattern correct: a shared `staleTime` so a consumer observer mounting under a shell reuses the shell's fetch instead of triggering a redundant refetch (which would otherwise reset in-progress settings forms); stable empty-array defaults so call sites don't rebuild `[]` each render (avoids breaking memo dependencies); feature gating baked into the category hooks so a disabled evaluation type yields an empty list even against warmed cache; and required `enabled` on the student-only hooks (`useGetMyEvaluations`, `useGetMyEvaluationCompletions`, and the extracted `useGetMyParticipation`) so staff-accessible evaluation routes cannot accidentally call the `my-*` endpoints.
- The three identical `useGet{Self,Peer,Tutor}EvaluationCategoriesWithCompetencies` hooks were collapsed into one `useGetEvaluationCategoriesWithCompetencies(type, enabled)` factory. Query keys are preserved exactly, so existing invalidations keep working.

Net result: 59 files changed, roughly 320 fewer lines.

The mirror stores were a second source of truth that lagged refetches: a store stayed stale for a render after a mutation (indefinitely if a consumer read it while the shell was unmounted), and every new dataset needed a store + setter + effect with a hand-maintained dependency array that could silently freeze the UI. The repo's own state-management rule (`.claude/rules/react-typescript/state-management.md`) states: "Server-derived data → React Query — this is cache, not application state. Don't mirror it in Zustand."

## 🧪 How to Test

1. `cd clients/assessment_component && yarn build` (or `tsc --noEmit` + `biome check src`) — passes.
2. Open an evaluation route as a student (self-evaluation) and confirm data renders; open it as a lecturer (non-student) and confirm the "not a student" banner shows and no `my-participation` / `my-evaluations` network calls fire.
3. Open an assessment route (participants, statistics) and a settings route; edit a settings field and confirm the form does not reset a moment after load (no stray config refetch).
4. Toggle an evaluation type off in settings and confirm its categories render empty even if that schema was viewed while enabled (gating, not warmed cache).
5. Perform a mutation (save an evaluation, edit settings) and confirm the UI reflects the change immediately without the prior one-render lag.

No visual changes; this is an internal state-management refactor.

## ✅ PR Checklist


mathildeshagl (approved):

JGStyle (approved):
went through the steps in how to test, worked fine. I couldn't see a 'no student' banner but i think that might just be due to how i set up my local assessment course

PR #1563, opened 2026-04-23
`Team Allocation`: Display team popularity and skill distribution statistics after survey

mathildeshagl:
## ✨ What is the change?

Adds a **Survey Statistics** page for instructors/lecturers showing survey results:
- **Summary cards**: survey responses with response rate progress bar (respondents vs. phase participants), most popular team, and least popular team (with first-choice counts and average rank).
- **Team Popularity**: stacked bar chart of how many students ranked each team as their 1st/2nd/3rd choice (teal/sky/indigo) — taller bars mean higher demand, teams ordered by average rank. Tooltip shows the average rank, response count, and the full per-rank breakdown (1st choice: N, …, Nth choice: N).
- **Skill Distribution**: stacked bar chart per skill using the assessment pastel palette (red→orange→amber→green→blue), with a legend and a tooltip showing counts and percentages per level.

Statistics are **intentionally visible while the survey is still running** (explicit product decision, diverging from the issue's original "after the survey closes"): the page shows a "Survey still open" alert with the deadline so lecturers know the results are live and partial.

The charts are responsive: below the mobile breakpoint the axis labels shrink and angle, the y-axis narrows, and a common prefix shared by all team names (e.g. "Team ") is stripped from the labels. Chart legends are explicitly ordered (recharts 3 defaults to alphabetical).

Changes the skill level scale from 4-level (`novice/intermediate/advanced/expert`) to 5-level (`very_bad/bad/ok/good/very_good`) matching the assessment component. TEASE export maps the two worst levels → Novice, others 1:1.

Also fixes the breadcrumb label on the TEASE Configuration page ("Tease-config" → "TEASE Configuration").

## 📌 Reason for the change / Link to issue

Closes #1552

## 🧪 How to Test

1. Navigate to a team allocation course phase — "Survey Statistics" appears in the sidebar
2. While the survey is still open, verify the page shows the "Survey still open" alert with the deadline and live statistics below it
3. Verify the summary cards show the response count with a response-rate progress bar plus the most and least popular teams
4. Verify the team popularity chart renders stacked 1st/2nd/3rd-choice bars ordered by average rank ascending, with a legend below
5. Hover or tap a bar and confirm the tooltip shows the average rank, responses, and all choice ranks (1st–Nth) including zeros
6. Verify the skill distribution chart uses the correct pastel colors per level (legend ordered Very Bad → Very Good) and its tooltip shows counts with percentages
[...]

PR #1853, opened 2026-07-07
`Team Allocation`: Scope tutor access to assigned team

### LINKED ISSUE #1698: Improve Editor Rights Management
As tutor and editor sometimes overlap and tutors should only have access to their own teams / assigned students in some cases, improve the authentication schema with the following: 

* map Tutors directly to Teams
* once this is activated, tutors only see teams they are assigned to
* Allow Team Allocation Edit (e.g. by giving Team Allocation an input team)
* Middleware to allow access based on university login

rappm:
# 📝 Pull Request Template

## ✨ What is the change?

Scopes tutor (`CourseEditor`) access in the Team Allocation service to the team a tutor is assigned to. A tutor's university login is now stored on the `tutor` record so the backend can resolve which team a signed-in editor belongs to and filter results accordingly.

- Adds a `university_login` column to the `tutor` table with a unique index per `(course_phase_id, university_login)` (migration `0008`).
- Sends the tutor university login on import (`TutorImportDialog`, `tutor.ts` interface).
- Adds `tutorScopingMiddleware` to the `team` and `allocation` routers: for a non-lecturer editor with a matching tutor login, the request is scoped to their assigned team; allocations are filtered by that team.
- Regenerates `sqlc` (`GetTutorTeamByUniversityLogin`, tutor models) and Swagger docs.
- Extends the e2e realm, seed, and constants to cover the tutor-scoping setup.

Improves editor rights management: tutors should only see and act on their own team's allocation data, not the whole course phase. Closes #1698.

## 🧪 How to Test

1. Run `make test` in `servers/team_allocation` (router and scoping tests).
2. Apply migrations and import a tutor with a university login via the Tutor Import dialog.
3. Sign in as that tutor (`CourseEditor`, non-lecturer) and confirm the team allocation views only expose their assigned team; a lecturer still sees everything.

| before | after |

## ✅ PR Checklist

## Summary by CodeRabbit

* **New Features**
  * Tutors can now be identified by university login during import and reassigned to a team using that login.
  * Team and allocation views now respect tutor-specific access, showing only the data a tutor is allowed to see.

* **Bug Fixes**
  * Duplicate tutor logins during import now return a clear conflict error.
  * Requests for unauthorized teams or allocations now return forbidden responses instead of succeeding silently.

* **Documentation**
  * Updated API docs to reflect the new tutor login field, reassignment flow, and additional error responses.


servers/team_allocation/team/router.go
```diff
@@ -8,24 +8,29 @@ import (
...
 	teamRouter.POST("/student-names", authMiddleware(promptSDK.PromptAdmin, promptSDK.CourseLecturer), addStudentNamesToTeams)
 
 	teamRouter.POST("/tutors", authMiddleware(promptSDK.PromptAdmin, promptSDK.CourseLecturer), importTutors)
+	teamRouter.PUT("/tutors/:universityLogin", authMiddleware(promptSDK.PromptAdmin, promptSDK.CourseLecturer), updateTutorTeam)
 
-	// this is required to comply with the inter phase communication protocol
-	teamRouter.GET("/:teamID", authMiddleware(promptSDK.PromptAdmin, promptSDK.CourseLecturer, promptSDK.CourseEditor, promptSDK.CourseStudent), getTeamByID)
+	// required for inter-phase communication protocol
+	teamRouter.GET("/:teamID", authMiddleware(promptSDK.PromptAdmin, promptSDK.CourseLecturer, promptSDK.CourseEditor, promptSDK.CourseStudent), scopingMW, getTeamByID)
 }
```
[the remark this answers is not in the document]
rappm: Verified against the current code and this is a false positive. Gin v1.12.0 (this module) supports a static segment and a wildcard as siblings at the same tree depth: `PUT /team/tutors/:universityLogin` and `PUT /team/:teamID` register without panicking, and the static `tutors` branch takes priority over `:teamID` at match time.

Confirmed with an isolated registration test (`gin.New()` + both PUT routes + the GET/POST siblings), which passes with no panic. The existing `router_test.go` also boots the full router via `setupTeamRouter` and passes. No change needed.

servers/team_allocation/db/query/tutor.sql
```diff
@@ -1,6 +1,11 @@
--- name: CreateTutor :exec
-INSERT INTO tutor (course_phase_id, course_participation_id, first_name, last_name, team_id)
-VALUES ($1, $2, $3, $4, $5);
+-- name: UpsertTutor :exec
+INSERT INTO tutor (course_phase_id, course_participation_id, first_name, last_name, team_id, university_login)
+VALUES ($1, $2, $3, $4, $5, $6)
+ON CONFLICT (course_phase_id, course_participation_id) DO UPDATE
+    SET team_id          = EXCLUDED.team_id,
+        university_login = EXCLUDED.university_login,
```
mathildeshagl: Upsert unconditionally overwrites `university_login`. The import dialog sends `''` when a login is absent, which `UniversityLoginParam` maps to NULL — so re-importing tutors from a source without logins wipes previously-set logins and (via the middleware's fail-open path) silently un-scopes those tutors. Consider preserving the existing value: `university_login = COALESCE(EXCLUDED.university_login, tutor.university_login)`.

servers/team_allocation/tutorscope/tutorscope.go
```diff
@@ -0,0 +1,63 @@
...
+		if err != nil {
+			c.AbortWithStatusJSON(http.StatusBadRequest, gin.H{"error": "invalid course phase id"})
+			return
+		}
+		teamID, err := q.GetTutorTeamByUniversityLogin(c.Request.Context(), db.GetTutorTeamByUniversityLoginParams{
+			CoursePhaseID:   coursePhaseID,
+			UniversityLogin: teamDTO.UniversityLoginParam(login),
+		})
+		if errors.Is(err, pgx.ErrNoRows) {
+			c.Next()
```
mathildeshagl: @rappm Fail-open: an editor whose login resolves to no tutor row passes through fully unscoped. I assume this is deliberate so a genuine non-tutor `CourseEditor`/organizer keeps full access — but it means a tutor whose `university_login` isn't recorded (e.g. imported before this migration, or cleared by a re-import) silently gets full access too. Is that the intended trust model? If so, worth a short comment documenting it.

servers/team_allocation/team/router_test.go
```diff
@@ -43,7 +43,7 @@ func (suite *TeamRouterTestSuite) SetupSuite() {
 	testMiddleware := func(allowedRoles ...string) gin.HandlerFunc {
 		return sdkTestUtils.MockAuthMiddlewareWithEmail(allowedRoles, "lecturer@example.com", "03711111", "ab12cde")
```
mathildeshagl, line 74: Both router suites mock lecturer/admin tokens, which take the middleware's pass-through branch — so none of the new scoping behavior (team resolution, `filterTeamsByID`/`filterAllocationsByTeam`, the 403 paths, `UpsertTutor`, `UpdateTutorTeam`, duplicate-login 409) is exercised, though the checklist marks tests as added. Please add a tutor-token case: tutor sees only their team, 403 on another team's resource, non-tutor editor still sees all, duplicate login → 409.

servers/team_allocation/allocation/router.go
```diff
@@ -54,6 +62,7 @@ func getAllAllocations(c *gin.Context) {
 // @Param courseParticipationID path string true "Course Participation UUID"
 // @Success 200 {object} allocationDTO.AllocationWithParticipation
```
mathildeshagl: This annotation says `@Success 200 {object} allocationDTO.AllocationWithParticipation`, but the handler returns a bare `uuid.UUID` (the team ID). Pre-existing inaccuracy, but you're editing this handler — worth correcting the Swagger while here.

servers/team_allocation/team/router.go
```diff
@@ -270,10 +295,66 @@ func importTutors(c *gin.Context) {
...
+// @Router /course_phase/{coursePhaseID}/team/tutors/{universityLogin} [put]
+func updateTutorTeam(c *gin.Context) {
+	coursePhaseID, err := uuid.Parse(c.Param("coursePhaseID"))
+	if err != nil {
+		log.Error("Error parsing coursePhaseID: ", err)
+		handleError(c, http.StatusBadRequest, err)
+		return
+	}
+
+	universityLogin := teamDTO.NormalizeUniversityLogin(c.Param("universityLogin"))
```
mathildeshagl, line 333: @rappm The doc string promises "TUM format aa00aaa" but only trim+lowercase is applied to the path param — no format validation. Intentional, or should this reject malformed logins?

mathildeshagl (changes requested):
Solid, well-structured change — the middleware is cleanly isolated, the `ON CONFLICT` target matches the tutor PK, the partial unique index and pg error-code handling are correct, and login normalization is centralized. A few things to resolve before merge, mostly around the fail-open trust model, the upsert clobbering logins, and missing test coverage for the new scoping paths. Left inline comments; tagged you on the ones where I'm unsure whether the behavior is intentional.

servers/team_allocation/db/query/tutor.sql
rappm: Fixed in d3099314 — the upsert now does `university_login = COALESCE(EXCLUDED.university_login, tutor.university_login)`, so a login-less re-import preserves the existing value.

servers/team_allocation/tutorscope/tutorscope.go
rappm: Deliberate, and now documented in 6d79b541 with a comment on the `ErrNoRows` branch: a login with no tutor row (genuine editor/organizer, or a tutor whose `university_login` is not recorded) keeps full, unscoped access by design. Combined with the COALESCE fix, the "cleared by re-import" case no longer happens silently.

servers/team_allocation/team/router_test.go
rappm: Added in 03cb9447. Both router suites now have tutor-token cases (via a middleware that sets an editor/non-lecturer `TokenUser`): tutor sees only their team/allocations, 403 on another team resource, non-tutor editor still sees all, and duplicate import login → 409. Seed dumps gained a tutor `university_login` + the unique index so these paths are reachable.

servers/team_allocation/allocation/router.go
rappm: Corrected in 6d79b541 to `@Success 200 {string} string "Assigned team UUID"`, matching the bare `uuid.UUID` the handler returns.

servers/team_allocation/team/router.go
rappm: Intentional — the login is only a lookup key (a malformed value simply matches no tutor → 404), and import does not format-validate either. Rather than add inconsistent validation, I dropped the misleading "TUM format aa00aaa" promise from the doc string in 6d79b541.

servers/team_allocation/db/migration/0009_tutor_university_login.up.sql
```diff
@@ -0,0 +1,9 @@
+BEGIN;
```
mathildeshagl, line 1: **Migration version collision — must renumber.** `main` already has `0008_skill_level_five_options.up.sql` (landed 2026-07-02 in #1563), so merging this adds a second migration at version `0008`. `golang-migrate` errors on duplicate versions and the repo's `detect-migrations` CI check will flag it.

Please rebase onto `main` and renumber both files to `0009_tutor_university_login.{up,down}.sql`. The branch is behind `main`, so a rebase is needed regardless — the collision only surfaces once rebased.

mathildeshagl (changes requested):
Overall a well-constructed, security-conscious change — the tutor-scoping model is sound (SDK middleware normalizes the login the same way storage does and fails closed on non-`ErrNoRows` errors), scope coverage is complete (the four scoped GETs are the only `CourseEditor`-reachable endpoints in this service), and the router tests are good.

One blocking issue before merge: the migration version collides with `0008_skill_level_five_options` already on `main` (see inline comment). Please rebase and renumber to `0009`. After that this looks ready.

A couple of non-blocking notes I'll leave to your discretion: the per-ID endpoints return 403 vs 200 which is a mild existence oracle (404 for out-of-scope would avoid confirming existence), and `ImportTutors` now upserts instead of insert-only — just confirm the silent update-on-reimport is the intended UX.

servers/team_allocation/db/migration/0009_tutor_university_login.up.sql
rappm: Renumbered both files to `0009_tutor_university_login.{up,down}.sql` in 948a7881. The branch already carries `0008_skill_level_five_options` from an earlier merge of main, so the collision is resolved. Note: still behind latest main; will fold in a fresh main sync.

mathildeshagl (approved):

PR #1880, opened 2026-07-09
`Interview`: become unified and fully independent of core

### LINKED ISSUE #1705: Interview: move to become a fully independent course phase microservice
For historic reason, the interview course phase microservice stores the interview score and comment inside the restricted data column of the course participation table.

The interview course phase however has grown overtime and now does have it's own database and backend. It makes sense to also move that data now into that storage to have a centralized place where everything is stored

JGStyle:
## ✨ What is the change?

make interview cpm fully indepdent of core and unified:
- reviews (revierwer + score) are now stored in the interview db
- the score info is now transferred to the next phase via REST (as standard in other phases, such as assessment), the core graph record has been updated accordingly
- client side aggregation updated to work with the rest api but still provide to the client

at the same time, core now no longer has a special case for aggregating the participations data from the database for interview: same codebase for all now.

fixes #1705 

## 🚀 Deployment steps (required, in this order)

Existing reviews live in core's `course_phase_participation.restricted_data` and nothing moves them
automatically, so an instance with a finished interview phase must run the backfill or every
conducted interview opens as an empty form.

1. Deploy core and the interview service so both apply their migrations
   (core `0029_interview_rest_provided_outputs`, interview `0002_interview_review`).
2. Run the one-off backfill once, against the two databases:

   ```bash
   CORE_DB_URL=postgres://user:pass@host:5432/prompt \
   INTERVIEW_DB_URL=postgres://user:pass@host:5432/prompt_interview \
   scripts/backfill-interview-reviews.sh
   ```

   It is idempotent: existing `interview_review` rows are never overwritten, so re-running it is
   safe and it will not clobber a review written after the cutover.
3. Core migration `0029` also rewrites the Interview type's `base_url` from `core` to
   `{CORE_HOST}/interview/api`. Without that the resolution URL stays relative and successor phases
   (matching) receive no score.

## 🧪 How to Test

ensure the interview phase works exactly as expected before, including passing the score to the next phase

no UI changes, only work behind the scenes

## ✅ PR Checklist

## Summary by CodeRabbit

* **New Features**
  * Added interview review management with interviewer details, answers, scores, score levels, and Accept/Reject actions.
  * Reviews save automatically and remain available when revisiting interviews.
  * Added interview score display and sorting in interview and matching views.
  * Added interview review APIs and persistent storage.
* **Privacy**
  * Included interview reviews in privacy exports and deletion.
* **Bug Fixes**
  * Improved participation and interview data loading, retry handling, and score synchronization.


JGStyle:
rebased on main

JGStyle:
@coderabbitai review

servers/interview/interviewReview/interviewReviewDTO/interviewReview.go
```diff
@@ -0,0 +1,121 @@
...
+func decodeInterviewAnswers(raw []byte) []InterviewAnswer {
+	answers := make([]InterviewAnswer, 0)
+	if len(raw) == 0 {
+		return answers
+	}
+	if err := json.Unmarshal(raw, &answers); err != nil {
+		return make([]InterviewAnswer, 0)
+	}
+	return answers
+}
```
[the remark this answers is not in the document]
JGStyle: now logging.

servers/core/db/migration/0029_interview_rest_provided_outputs.up.sql
```diff
@@ -0,0 +1,19 @@
+BEGIN;
+
+UPDATE course_phase_type_participation_provided_output_dto po
+SET endpoint_path = '/interview-review/score'
```
rappm, line 9: **`endpoint_path` moves to REST, but `base_url` for the Interview type is never updated, so the resolution URL is unusable on every database that already has the type.**

`0011_inter_phase_communication.up.sql:190` sets `base_url = 'core'` for the Interview type, and `initInterview()` only writes `{CORE_HOST}/interview/api` in the `if !exists` branch, so on an instance that already has the type it returns early and the column keeps the old value. Nothing else ever updates it: across `servers/core/db/query/`, `base_url` appears only in the `SELECT`s of `course_phase.sql` and in the `INSERT` of `course_phase_type.sql`. Our own core test dump agrees that `'core'` is the real-world value (`full_db.sql:555`).

Failure case: after this migration `GetResolutionsForCoursePhase` hands out `{baseURL: "core", endpointPath: "/interview-review/score"}`. `ReplaceResolutionURLs` leaves `core` untouched because it carries no `{CORE_HOST}` placeholder, so both `promptSDK.buildURL` and the new client-side `buildResolutionURL` produce the relative path `core/course_phase/<id>/interview-review/score`, which 404s. Because branch (II) is deleted from `course_phase_participation.sql` in the same PR, matching and every other successor phase lose the interview score outright. Nothing local catches it: `e2e/seed/e2e_seed.sql:752` already seeds the Interview type with `{CORE_HOST}/interview/api`, so e2e passes on a state production is not in.

Suggested fix: add the `base_url` update to this migration, with the inverse in the down file.

```sql
UPDATE course_phase_type
   SET base_url = '{CORE_HOST}/interview/api'
 WHERE name = 'Interview'
   AND base_url = 'core';
```

Worth double-checking against the dev override in `initInterview()` too, which uses `http://localhost:8087/interview/api`. A migration cannot know which of the two applies, so the placeholder form is the only one that works in both.

servers/interview/db/migration/0002_interview_review.up.sql
```diff
@@ -0,0 +1,16 @@
+BEGIN;
+
+CREATE TABLE interview_review (
```
rappm, line 3: **Nothing backfills the existing reviews, so every interview conducted so far disappears from the UI.**

Until this PR, score, interviewer and interviewAnswers lived in core's `course_phase_participation.restricted_data` for Interview phases. This migration creates `interview_review` empty, and the PR points every reader at it (`InterviewCard`, `StudentCard`, `useSorting`, and the new `/score` + `/scoreLevel` outputs), but no step copies the old values across.

Failure case: on the first deploy to an instance with a finished interview phase, every already-conducted interview opens as an empty form (no answers, no score, no interviewer) and downstream matching sees no scores at all. The data is not destroyed, it just stays orphaned in core's `restricted_data` where nothing reads it anymore, so it is recoverable but only by hand.

This one is awkward because the source and the target live in different databases, so it cannot be a plain migration in either service. Options I see: a one-off script run at deploy time, or a core endpoint that hands out the old `restricted_data` for a phase and an interview-side import that calls it once. Either is fine, but something has to exist before this merges, and it would be good to have it written down in the PR description so whoever deploys knows the order.

servers/interview/db/migration/0002_interview_review.up.sql
```diff
@@ -0,0 +1,16 @@
...
+    interviewer varchar(255),
+    interview_answers jsonb NOT NULL DEFAULT '[]'::jsonb,
+    created_at timestamp with time zone DEFAULT now(),
+    updated_at timestamp with time zone DEFAULT now(),
+    PRIMARY KEY (course_phase_id, course_participation_id)
+);
+
+CREATE INDEX idx_interview_review_course_phase ON interview_review(course_phase_id);
+
+COMMIT;
```
rappm, line 16: There is no `0002_interview_review.down.sql`, so `migrate down` against the interview service fails with "no down migration found for version 2" and this deploy has no rollback path. Core's `0027` in this same PR ships its down file, so this reads as an oversight rather than a decision.

servers/interview/privacy/service.go
```diff
@@ -17,9 +20,25 @@ func PrivacyDataExportHandler(c *gin.Context, exp *utils.Export, subject sdkAuth
 	exp.AddJSON("Interview Assignments", "interview_assignments.json", func() (any, error) {
 		return PrivacyServiceSingleton.Queries.GetInterviewAssignmentsByParticipationIDs(c, subject.CourseParticipationIDs)
 	})
+	exp.AddJSON("Interview Reviews", "interview_reviews.json", func() (any, error) {
+		return PrivacyServiceSingleton.Queries.GetInterviewReviewsByParticipationIDs(c, subject.CourseParticipationIDs)
+	})
```
rappm, line 65: **The export writes the interview answers as base64.**

`GetInterviewReviewsByParticipationIDs` returns `[]db.InterviewReview`, whose `InterviewAnswers` field is `[]byte`: `servers/interview/sqlc.yaml` overrides only `uuid`, so `jsonb` falls back to sqlc's default. `AddJSON` marshals with `encoding/json`, which base64-encodes any byte slice.

Failure case: a student exercises their access request and gets `"interview_answers": "W3sicXVlc3Rpb25JRCI6..."` instead of the answers. Formally complete, practically unreadable, which is the one property a subject access response cannot have.

Suggested fix: `interviewReviewDTO.GetInterviewReviewsFromDB` already decodes this via `decodeInterviewAnswers`, so wrapping the call is enough. The deletion handler below is fine.

clients/matching_component/src/matching/network/resolveParticipations.ts
```diff
@@ -0,0 +1,93 @@
...
+          { headers: authHeaders(), timeout: RESOLUTION_TIMEOUT_MS },
+        )
+        for (const item of response.data ?? []) {
+          const id = item.courseParticipationID
+          if (typeof id === 'string') {
+            valuesByParticipation[id] = item[resolution.dtoName]
+          }
+        }
+      } catch (err) {
+        console.error(`Failed to resolve "${resolution.dtoName}" from ${resolution.baseURL}`, err)
```
rappm, line 96: **A failed resolution is only logged, and the matching export silently loses every rank.**

The catch swallows everything (a 403, a timeout, the URL problem from my comment on migration `0027`, a plain network blip) and the participation simply ends up with no `prevData.score`. Since the failure is handled inside the queryFn, the `['participants', phaseId]` query still resolves successfully, so `isError` never trips and nothing upstream can distinguish "this student has no score" from "we could not load the scores".

Failure case: `useStudentMatching.tsx:32` builds each student's `rank` from `matchedApp.prevData.score`. One transient failure while a lecturer has the page open produces a matching export in which every student has `rank: undefined`, and the only trace of it is an "N/A" in the participants table, which reads like an unscored student rather than a broken fetch. That export is what the matching is run on.

The docstring's reasoning is right for rendering: one dead provider should not blank the page. It stops being right once the same data is exported. Please return the failed `dtoName`s out of `resolveParticipations` so the page can surface them, and ideally block the export while a resolution is missing rather than exporting a silently incomplete file.

clients/interview_component/src/interview/network/hooks/useUpdateInterviewReview.ts
```diff
@@ -0,0 +1,24 @@
...
+  return useMutation({
+    mutationFn: ({ courseParticipationID, review }: UpdateInterviewReviewVariables) => {
+      if (!coursePhaseID) {
+        throw new Error('Cannot update interview review without a course phase ID')
+      }
+      return updateInterviewReview(coursePhaseID, courseParticipationID, review)
+    },
+    onSuccess: () => {
+      queryClient.invalidateQueries({ queryKey: ['interviewReviews', coursePhaseID] })
+    },
```
rappm, line 29: No `onError`, so a rejected save leaves the edited score and answers on screen as if they had been persisted. The card saves on blur, so the reviewer has no reason to look for a result: the `score must be between 1 and 5` 400, a 500, and a dropped connection all look exactly like success until the page is reloaded and the work is gone. `useUpdateCoursePhaseParticipation`, which this replaces, at least rolled its optimistic cache entry back. A toast plus an invalidate would be enough here.

servers/interview/db/query/interview_review.sql
```diff
@@ -0,0 +1,38 @@
...
+WHERE course_phase_id = $1
+ORDER BY course_participation_id ASC;
+
+-- name: GetScoredInterviewReviewsByCoursePhase :many
+SELECT * FROM interview_review
+WHERE course_phase_id = $1
+  AND score IS NOT NULL
+ORDER BY course_participation_id ASC;
+
+-- name: DeleteInterviewReviewByParticipation :exec
```
rappm: `DeleteInterviewReviewByParticipation` is byte-for-byte identical to `DeleteInterviewReviewsByParticipationIDs` in `privacy.sql`, and only the latter has a caller. `GetInterviewReview` above is generated but never used either. Two copies of the same delete is the kind of thing where a later fix lands on one of them, and this one sits on the GDPR path. Please drop both.

Same area: `InterviewReviewService.conn` (`interviewReview/service.go:16`) is assigned in `InitInterviewReviewModule` and never read.

servers/core/db/query/course_phase_participation.sql
```diff
@@ -223,53 +223,14 @@ SELECT
                        (jsonb_each(pcpp.restricted_data)).value AS value
               ) sub
             ) AS each
-            WHERE dpm.from_course_phase_type_name NOT IN ('Application', 'Interview')
+            WHERE dpm.from_course_phase_type_name NOT IN ('Application')
```
rappm, line 226: Removing the special case is the right direction, but nothing in core now covers what replaces it. `full_db.sql:555` still seeds the Interview phase type with `base_url = 'core'` and `full_db.sql:566` its `score` output with `endpoint_path = 'core'`, so the core suite keeps exercising the pre-migration shape, and `TestGetParticipationsWithPrevData` targets the Interview phase itself rather than a successor of it. The result is that branch (II) can be deleted with the suite staying green either way.

Please bring the dump in line with `0027` and add a case that reads `score` through a successor phase. As it stands the fixture happens to encode exactly the production state that my comment on `0027` is about, and the tests still pass, which is a good illustration of the gap.

rappm (changes requested):

rappm:
@JGStyle in case you missed this

JGStyle:
> @JGStyle in case you missed this

thanks

servers/core/db/migration/0029_interview_rest_provided_outputs.up.sql
rappm: Added the base_url update to the migration (now 0029, it collided with 0027 on main) plus the inverse in the down file, using the {CORE_HOST} placeholder form so it works in dev and prod. Covered by a new core test that asserts the resolution resolves to https://localhost:8080/interview/api; reverting just this UPDATE makes it fail with actual: "core".

servers/interview/db/migration/0002_interview_review.up.sql
rappm: Added scripts/backfill-interview-reviews.sh, a one-off psql copy from core's restricted_data into interview_review. It is idempotent (ON CONFLICT DO NOTHING), rounds legacy fractional scores to the integer column, and skips non-Interview phases. Verified end to end against two throwaway Postgres containers, including that tabs/newlines/backslashes/quotes inside the answer HTML round-trip and that a row written after the cutover is not clobbered. Deploy order is now written up in the PR description.

servers/interview/db/migration/0002_interview_review.up.sql
rappm: Added 0002_interview_review.down.sql (drops the index and the table).

servers/interview/privacy/service.go
rappm: Fixed. The export now maps through interviewReviewDTO.GetInterviewReviewFromDB so the answers come out as JSON, and keeps coursePhaseID plus the created/updated timestamps that the raw row carried.

clients/matching_component/src/matching/network/resolveParticipations.ts
rappm: resolveParticipations now returns { participations, failedResolutions }. Both matching pages show a destructive alert naming the failed dtoNames, and the TUM matching export upload is disabled while any resolution is missing, so a transient failure can no longer produce an export with every rank undefined.

clients/interview_component/src/interview/network/hooks/useUpdateInterviewReview.ts
rappm: Added an onError that invalidates the interviewReviews query and raises a destructive toast with the server's error message, so the 400/500/dropped-connection cases no longer look like a successful save.

servers/interview/db/query/interview_review.sql
rappm: Dropped both DeleteInterviewReviewByParticipation and GetInterviewReview, regenerated sqlc, and removed the unused conn field from InterviewReviewService (and its assignment in InitInterviewReviewModule and the test setup).

servers/core/db/query/course_phase_participation.sql
rappm: Brought full_db.sql in line with 0029 (Interview base_url and the score endpoint_path) and added TestInterviewScoreIsResolvedNotInlined, which reads through the successor Matching phase: it asserts score is absent from prevData and present as a REST resolution with the resolved base URL. The interview participations it needs carry a stale score in restricted_data on purpose, so re-adding the inlining would fail the test.

rappm (approved):

PR #1889, opened 2026-07-11
`Assessment`: Fix N+1 queries in `ListCompetenciesForCoursePhase`

### LINKED ISSUE #1794: N+1 queries in ListCompetenciesForCoursePhase
## Summary

`ListCompetenciesForCoursePhase` loads every competency in the database and then issues two queries per competency, producing `1 + 2N` sequential round-trips per request.

## Location

`servers/assessment/competencies/service.go` `ListCompetenciesForCoursePhase`:

- `ListCompetencies(ctx)` is `SELECT * FROM competency` with no `WHERE` (`db/query/competency.sql`), so it loads all competencies platform-wide.
- The loop then calls `GetAssessmentSchemaIDByCompetency` and `CheckSchemaAccessibleForCoursePhase` (a real DB query) for each competency.

## Impact

With, say, 50 schemas x 30 competencies platform-wide, a single `GET /competency` fires roughly `1 + 2 x 1500 = 3001` sequential DB round-trips to return the ~30 competencies of one phase. Every lecturer page load pays this, and it grows with total platform history rather than the phase size.

## Suggested fix

Replace the load-all-then-filter loop with a single query that selects competencies whose schema is accessible to the phase (a JOIN against the accessible-schema subquery). The sibling `ListCategoriesForCoursePhase` already fetches the accessible schemas once and filters, which proves the data model supports it; ideally push the filter fully into SQL.

rappm:
## ✨ What is the change?

Replaces the load-all-then-filter loop in `ListCompetenciesForCoursePhase` (`servers/assessment/competencies/service.go`) with a single sqlc query that JOINs `competency -> category -> assessment_schema` and filters by the schemas accessible to the given course phase. The service now issues one query instead of `1 + 2N`.

The old implementation was an N+1:
1. `ListCompetencies(ctx)` = `SELECT * FROM competency` with no `WHERE`, loading every competency platform-wide.
2. For each competency, `GetAssessmentSchemaIDByCompetency` (1 query) + `CheckSchemaAccessibleForCoursePhase` (1 query).

That is `1 + 2N` sequential round-trips per request, where `N` is the total number of competencies across all courses — it grows with total platform history rather than phase size, and every lecturer page load pays it.

The new sqlc query `ListCompetenciesForCoursePhase` reuses the same accessible-schema subquery pattern as `ListAssessmentSchemasForCoursePhase` (phase-owned, source-null, or referenced by the phase config's assessment/self/peer/tutor schemas). No migration is required (no schema change) — only a new query plus regenerated `db/sqlc/`. Per-request query count drops from `1 + 2N` to **1**.

Closes #1794

## 🧪 How to Test

1. `cd servers/assessment`.
2. `go build ./...` and `go vet ./competencies/...` — clean.
3. `go test ./competencies/...` (testcontainers, requires Docker) — passes, including the new `TestListCompetenciesForCoursePhase`, which asserts the returned set is non-empty, a subset of all competencies, and that every returned competency belongs to a schema accessible for the phase (equivalence with the old per-item filter).
4. `GET /competency` for a phase returns the same competencies as before, now via a single DB round-trip.

N/A — backend-only change.

## ✅ PR Checklist

- [ ] Screenshots attached for UI changes (if any) — N/A
- [ ] Documentation updated (if relevant) — N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved course-phase competency listings to include only competencies applicable to the selected phase.
  * Ensured phase-independent competencies and competencies linked through configured assessment types are included correctly.
  * Results are consistently ordered by competency name.


servers/assessment/db/query/competency.sql
```diff
@@ -34,6 +34,24 @@ WHERE comp.id = $1;
...
+FROM competency comp
+INNER JOIN category cat ON comp.category_id = cat.id
+INNER JOIN assessment_schema s ON cat.assessment_schema_id = s.id
+LEFT JOIN phase_config pc ON TRUE
+WHERE s.source_phase_id IS NULL
+   OR s.source_phase_id = sqlc.arg(course_phase_id)
+   OR s.id = pc.assessment_schema_id
+   OR s.id = pc.self_evaluation_schema
+   OR s.id = pc.peer_evaluation_schema
+   OR s.id = pc.tutor_evaluation_schema;
```
mathildeshagl: No `ORDER BY`, so result order is unspecified. Not a regression (the old `ListCompetencies` was unordered too), but the sibling `ListAssessmentSchemasForCoursePhase` sorts by `name ASC`. Add `ORDER BY comp.name` for deterministic output and consistency with that query.

servers/assessment/competencies/service_test.go
```diff
@@ -78,6 +78,27 @@ func (suite *CompetencyServiceTestSuite) TestListCompetencies() {
...
+
+	competencies, err := ListCompetenciesForCoursePhase(suite.ctx, coursePhaseID)
+	assert.NoError(suite.T(), err)
+	assert.Greater(suite.T(), len(competencies), 0, "Expected at least one accessible competency")
+
+	all, err := ListCompetencies(suite.ctx)
+	assert.NoError(suite.T(), err)
+	assert.LessOrEqual(suite.T(), len(competencies), len(all), "Phase competencies should be a subset of all competencies")
+
+	for _, competency := range competencies {
```
mathildeshagl: This checks no false positives (every returned competency is accessible) but not completeness — a regression that wrongly *drops* accessible competencies would still pass. Since the old per-item path still exists, consider asserting exact-set equality: build the expected set by filtering `all` through `CheckSchemaAccessibleForCoursePhase`, then compare membership. That makes this a true equivalence guard for the refactor.

mathildeshagl (changes requested):
Correct, well-scoped fix — the single query is logically equivalent to the old `1 + 2N` filter (join path and WHERE predicate match `CheckSchemaAccessibleForCoursePhase`, and omitting `DISTINCT` is safe since `course_phase_config.course_phase_id` is the PK). Two non-blocking cleanups inline before merge.

rappm:
Addressed both cleanups: added `ORDER BY comp.name` (regenerated sqlc) and changed the test to assert exact-set equality against the per-item `CheckSchemaAccessibleForCoursePhase` filter.

Issue #1740, opened 2026-06-25
Fix performance problems found by test suite


Mtze:
In https://github.com/prompt-edu/prompt/pull/1739 we added a stress test for the server sides. The suite found several performance issues add error paths. These need fixing. 

[report.md](https://github.com/user-attachments/files/29329008/report.md)

Mtze:
Filed per-phase reports for the findings from the `api-stress` suite (PR #1739). Two deterministic runs produced identical results.

### Public issues (robustness / error paths)
- [ ] #1745 — core: 500 instead of 404/400 (missing not-found handling + file-upload input validation)
- [ ] #1741 — team_allocation: 500 instead of 404 on valid GETs (unhandled `pgx.ErrNoRows`)
- [ ] #1742 — self_team_allocation: 500 on allocation-by-participation
- [ ] #1743 — assessment: 500 on assessment-schema update (PUT)
- [ ] #1744 — certificate: 500 on download/preview when template not configured
- [ ] #1746 — setup: Keycloak 26.6.3 fails to import the realm on a fresh DB

### Private (security)
A **cross-course broken-access-control (BOLA/IDOR)** finding affecting the phase microservices (assessment, certificate, team-allocation, self-team-allocation) was filed as a **private draft security advisory** rather than a public issue, since it is exploitable against the live instance. Maintainers can view it under **Security → Advisories** (`GHSA-4wgm-2wvm-fphm`). It has one systemic root cause (per-course-phase membership not enforced in the shared auth path) and should be fixed before the public robustness issues if prioritizing by risk.

Common pattern across the public issues: unhandled `pgx.ErrNoRows` (and unvalidated request bodies on the file endpoints) returning **500** where **404/400** is expected. Each issue links the repro via `api-stress/run.sh`.

PR #1890, opened 2026-07-11
`Self Team Allocation`: Return 404 instead of 500 for missing allocation

### LINKED ISSUE #1742: 500 instead of 404 on allocation-by-participation (unhandled ErrNoRows)
`GET /self-team-allocation/api/course_phase/{coursePhaseID}/allocation/{courseParticipationID}`
returns **HTTP 500** when the participation has no allocation, instead of `404`.
Same class as the team-allocation issue: an unhandled `pgx.ErrNoRows` is mapped to
500.

### Expected
`404 Not Found` (or empty) when there is no allocation for the participation;
reserve 500 for real DB failures (`errors.Is(err, pgx.ErrNoRows)`).

### Where to look
`servers/self_team_allocation/allocation/` (handler `getAllocationByCourseParticipationID`).

---
### How to reproduce / verify
Found by the `api-stress` stress & fuzz suite (PR #1739), deterministic across 2 runs.
```bash
cp api-stress/stress.env.example api-stress/stress.env
docker compose --env-file api-stress/stress.env -f docker-compose.yml -f api-stress/docker-compose.stress.yml -p prompt-stress up -d   # see api-stress/README.md for the full service list
api-stress/run.sh --smoke-only        # reachability + per-endpoint status/latency
```
Open `api-stress/reports/<ts>/report.md` and `findings.json`; each finding has the endpoint, observed vs expected, a `servers/<service>/` pointer, and a `curl` repro. Re-run after a fix and confirm the endpoint's `5xx` count drops to 0.

Part of #1740.

rappm:
## ✨ What is the change?

`GET /self-team-allocation/api/course_phase/{coursePhaseID}/allocation/{courseParticipationID}` now returns `404 Not Found` when the participation has no allocation, instead of `500 Internal Server Error`.

The handler `getAllocationByCourseParticipationID` now branches on `errors.Is(err, pgx.ErrNoRows)` and maps that case to `404`, while all other errors continue to return `500`. This matches the existing pattern already used in the sibling `team_allocation` service (`servers/team_allocation/allocation/router.go`).

Previously an unhandled `pgx.ErrNoRows` was mapped to `500`, so a participation without an allocation produced a server error instead of a proper "not found" response. `500` should be reserved for real DB failures.

Closes #1742
Part of #1740

## 🧪 How to Test

1. Start the self_team_allocation service.
2. Call `GET /self-team-allocation/api/course_phase/{coursePhaseID}/allocation/{courseParticipationID}` with a `courseParticipationID` that has no allocation.
3. Confirm the response is `404 Not Found` (previously `500`).
4. Call the same endpoint with a participation that has an allocation and confirm it still returns `200` with the team ID.
5. Automated: from `servers/self_team_allocation`, run `go test ./allocation/...` (includes `TestGetAllocationByCourseParticipationIDNotFound`, now asserting `404`).

N/A - backend-only change.

## ✅ PR Checklist


servers/self_team_allocation/allocation/router.go
```diff
@@ -70,7 +72,11 @@ func getAllocationByCourseParticipationID(c *gin.Context) {
 
 	allocation, err := GetAllocationByCourseParticipationID(c, courseParticipationID, coursePhaseID)
 	if err != nil {
-		handleError(c, http.StatusInternalServerError, err)
+		if errors.Is(err, pgx.ErrNoRows) {
```
mathildeshagl, line 76: Swagger godoc for this handler still lists only `@Failure 400` / `@Failure 500`. Now that 404 is a documented outcome, add `// @Failure 404 {object} map[string]string`. Same annotation is also omitted in the sibling `servers/team_allocation/allocation/router.go` — worth fixing in both so the docs match the behavior.

servers/self_team_allocation/allocation/router.go
```diff
@@ -70,7 +72,11 @@ func getAllocationByCourseParticipationID(c *gin.Context) {
...
+			handleError(c, http.StatusNotFound, err)
```
mathildeshagl, line 77: A missing allocation is now an expected response, but it's still logged twice at ERROR level — once in `service.go:41` (`log.Error("Error fetching assignment...")`) and again in `handleError` (`router.go:81`). Consider demoting the `pgx.ErrNoRows` case to Debug so a normal 404 doesn't produce error-log noise.

mathildeshagl (changes requested):
The 404 mapping is correct and matches the sibling `team_allocation` pattern. Two small things before merge — see inline comments: the Swagger godoc should document the new 404 (also missing in the sibling service), and the not-found path now logs at ERROR level twice for what is a normal response.

rappm:
Added the `@Failure 404` swagger annotation (and the same on the sibling `team_allocation` route), and demoted the missing-allocation case to Debug in both `service.go` and `handleError` so a normal 404 no longer produces error-log noise.

PR #1892, opened 2026-07-11
`Certificate`: Fix 500 on download/preview when template not configured

### LINKED ISSUE #1744: certificate: 500 on certificate download/preview when template not configured
`GET /certificate/api/course_phase/{coursePhaseID}/certificate/download` (and the
preview/download-student variants) returns **HTTP 500** when no certificate
template/data is configured for the phase, instead of `404`/`409`.

### Expected
A `404`/`409` with a clear "certificate not configured / not available" message -
not 500. (Note the sibling `/certificate/status` endpoint already returns a clean
`{"available":false,...}` 200, so the download path should fail gracefully too.)

### Where to look
`servers/certificate/generator/` (handlers `downloadOwnCertificate` / preview).

---
### How to reproduce / verify
Found by the `api-stress` stress & fuzz suite (PR #1739), deterministic across 2 runs.
```bash
cp api-stress/stress.env.example api-stress/stress.env
docker compose --env-file api-stress/stress.env -f docker-compose.yml -f api-stress/docker-compose.stress.yml -p prompt-stress up -d   # see api-stress/README.md for the full service list
api-stress/run.sh --smoke-only        # reachability + per-endpoint status/latency
```
Open `api-stress/reports/<ts>/report.md` and `findings.json`; each finding has the endpoint, observed vs expected, a `servers/<service>/` pointer, and a `curl` repro. Re-run after a fix and confirm the endpoint's `5xx` count drops to 0.

Part of #1740.

rappm:
## ✨ What is the change?

The certificate download and preview endpoints now return a clean `404 Not Found` with a `"Certificate template not configured"` message when no certificate template is configured for the course phase, instead of an opaque `500 Internal Server Error`.

Affected endpoints:
- `GET /certificate/api/course_phase/{coursePhaseID}/certificate/download` (student self-download)
- `GET /certificate/api/course_phase/{coursePhaseID}/certificate/download/{studentID}` (instructor download for a student)
- `GET /certificate/api/course_phase/{coursePhaseID}/certificate/preview` (instructor preview)

Each handler now performs the same "is a template configured?" check that the sibling `/certificate/status` endpoint already uses (`getTemplateStatus`), and short-circuits with a 404 before attempting generation. This makes the download path fail gracefully and consistently with `/status`, which already returns a clean `{"available":false,...}` 200.

**Why 404 (not 409):** The service has no existing 409 usage; it signals "not available" states with clear codes (400 bad input, 403 not yet released, 422 template compilation error). A missing/unconfigured certificate template is best modeled as "the requested resource does not exist yet," so 404 is the natural fit and keeps clients simple (404 = nothing to download).

Requesting a certificate download or preview for a phase without a configured template surfaced a generic 500, which reads as a server fault rather than an expected "not configured" state. This aligns the download/preview behavior with the already-graceful `/status` endpoint.

Closes #1744
Part of #1740

## 🧪 How to Test

1. Use a course phase that has no certificate template configured (e.g. the seeded phase with a NULL template).
2. `GET /certificate/api/course_phase/{coursePhaseID}/certificate/download` → expect `404` with `{"error":"Certificate template not configured"}`.
3. `GET /certificate/api/course_phase/{coursePhaseID}/certificate/download/{studentID}` → expect `404` with the same message.
4. `GET /certificate/api/course_phase/{coursePhaseID}/certificate/preview` → expect `404` with the same message.
5. Confirm a phase **with** a configured template still returns the PDF (200) as before.
6. Automated: `cd servers/certificate && go test ./generator/...` (testcontainers).

N/A - backend-only change.

## ✅ PR Checklist


servers/certificate/generator/router.go
```diff
@@ -40,6 +40,12 @@ func downloadOwnCertificate(c *gin.Context) {
 		return
 	}
 
+	if _, err := getTemplateStatus(c, coursePhaseID); err != nil {
```
mathildeshagl: `getTemplateStatus` returns an error for two different cases: template genuinely unconfigured, *and* `GetCoursePhaseConfig` itself failing (transient DB error, context cancellation, etc.). This guard collapses both into a 404 "not configured", so a real DB outage on a template-configured phase is now masked as "nothing to download" and hidden from monitoring. Consider distinguishing a not-configured sentinel from genuine query errors and letting the latter surface as 500.

servers/certificate/generator/router.go
```diff
@@ -109,6 +115,12 @@ func downloadStudentCertificate(c *gin.Context) {
...
```
mathildeshagl: This identical 5-line guard now appears in all three handlers. Worth extracting a small helper (e.g. `ensureTemplateConfigured(c, coursePhaseID) bool`) so the error-vs-sentinel handling from the note above lives in one place.

servers/certificate/generator/router.go
mathildeshagl: Redundant DB read: `getTemplateStatus` calls `GetCoursePhaseConfig`, and the student release-date branch below (line ~58) calls it again. If you extract a helper, consider returning the fetched `config` and reusing it for the release-date check.

servers/certificate/generator/router.go
```diff
@@ -232,6 +244,12 @@ func previewCertificate(c *gin.Context) {
...
```
mathildeshagl: Confirm `GeneratePreviewCertificate` has no distinct "no template" branch that this guard now makes dead code.

mathildeshagl (changes requested):
Solid, well-tested fix that achieves its goal and follows conventions. One thing to resolve before merge: the guard turns *any* `GetCoursePhaseConfig` failure into a 404 "template not configured", so a real DB/infra error on a valid phase would be masked as "nothing to download". Splitting the not-configured sentinel from genuine query errors (and extracting the repeated block) would make this robust rather than only consistent-with-`/status`. Inline notes below.

Note: `TestPreviewCertificate_NonExistentPhase` now asserts 404, which locks in the "any config-lookup failure = 404" semantics — worth a conscious decision alongside the above.

rappm:
Reworked the guard: introduced an `errTemplateNotConfigured` sentinel so a genuine `GetCoursePhaseConfig` failure (DB outage, cancellation) now surfaces as 500 instead of being masked as a 404, while a missing config row or empty template stays a 404 (200 available:false for the status endpoint). Extracted the shared check into `ensureTemplateConfigured`, which returns the fetched config so the release-date branch reuses it instead of re-querying. Confirmed `GeneratePreviewCertificate`'s no-template branch is not dead: it reads via `config.GetTemplateContent`, a different query than the guard's `GetCoursePhaseConfig`, so it remains a valid defensive check.

JGStyle (commented):
I can see a 404 and the correct log instead of 500.
Works 👍

Image

JGStyle (approved):

PR #1899, opened 2026-07-11
`Application`: Fix infinite loading spinner when authenticated user has no email

### LINKED ISSUE #1764: Application form stuck on "Loading application form…" when the authenticated user has no email
## Summary

On the authenticated application page (`/apply/{phaseId}/authenticated`), the form is stuck forever on "Loading application form…" for any logged-in student whose Keycloak account has no email address. There is no error and no timeout, just a permanent spinner.

## Root cause

In [`ApplicationAuthenticated.tsx`](clients/core/src/publicPages/application/pages/ApplicationAuthenticated/ApplicationAuthenticated.tsx), the `application` query is gated on the user's email:

```ts
const { data: application, isPending: isApplicationPending, ... } = useQuery<GetApplication>({
  queryKey: ['application', phaseId, user?.email],
  queryFn: () => getApplication(phaseId ?? ''),
  enabled: !!phaseId && !!user?.email && !!localStorage.getItem('jwt_token'),
})
```

`user.email` comes from the `email` claim in the Keycloak token and is defaulted to `''` in `useKeycloak.tsx` when the claim is absent. If the Keycloak account has no email, `enabled` is `false`, the query never runs, and in react-query v5 a disabled query stays `isPending: true`. The render guard `if (isPending || isApplicationPending)` therefore shows `LoadingState` indefinitely. The request to `GET /api/apply/authenticated/{phaseId}` is never sent.

## Reproduction

1. Have a TUM student Keycloak account with no email set.
2. Open a course application, choose "Log in with TUM-ID", authenticate.
3. Page lands on `/apply/{phaseId}/authenticated` and hangs on "Loading application form…". Network trace shows `/api/apply/{phaseId}` (200) but never `/api/apply/authenticated/{phaseId}`.

## Suggested fix

Authorization is via the JWT, not the email, so the query should not depend on `user.email`. Drop `!!user?.email` from the `enabled` condition (gate on `phaseId` + the jwt token only). Consider also surfacing an error/timeout instead of an indefinite spinner if a required query stays disabled.

## Workaround applied

On the affected instance, an email was set on the affected accounts in Keycloak so the token carries the `email` claim. This is a data workaround; the frontend should not require the claim.

rappm:
# 📝 Pull Request Template

## ✨ What is the change?

The authenticated application page (`/apply/{phaseId}/authenticated`) no longer gates its `application` query on `user.email`. The `enabled` condition and the query key now depend only on the phase and the JWT token, and the mutation's invalidation key was updated to match.

Fixes #1764.

`user.email` comes from the Keycloak `email` claim and is defaulted to `''` in `useKeycloak` when the claim is absent. When a logged-in student's Keycloak account has no email, `enabled` was `false`, so the query never ran. In react-query v5 a disabled query stays `isPending: true`, and the render guard `if (isPending || isApplicationPending)` left the page stuck forever on "Loading application form…" with no error and no timeout.

Authorization is via the JWT, not the email, so the query should not depend on `user.email`.

## 🧪 How to Test

1. Use a Keycloak student account with **no** email address set.
2. Open a course application, choose "Log in with TUM-ID", and authenticate.
3. Land on `/apply/{phaseId}/authenticated` and confirm the application form loads (previously it hung on the loading spinner). The network trace should now show `GET /api/apply/authenticated/{phaseId}`.

## ✅ PR Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved authenticated application loading by using the active session token and phase information.
  * Updated application data refresh behavior after submissions to ensure the latest information is displayed.

### DIFF
--- a/clients/core/src/publicPages/application/pages/ApplicationAuthenticated/ApplicationAuthenticated.tsx
+++ b/clients/core/src/publicPages/application/pages/ApplicationAuthenticated/ApplicationAuthenticated.tsx
@@ -43,9 +43,9 @@ export const ApplicationAuthenticated = () => {
     isError: isApplicationError,
     error: applicationError,
   } = useQuery<GetApplication>({
-    queryKey: ['application', phaseId, user?.email],
+    queryKey: ['application', phaseId],
     queryFn: () => getApplication(phaseId ?? ''),
-    enabled: !!phaseId && !!user?.email && !!localStorage.getItem('jwt_token'),
+    enabled: !!phaseId && !!localStorage.getItem('jwt_token'),
   })
 
   const { mutate: mutateSendApplication, error: mutateError } = useMutation({
@@ -53,7 +53,7 @@ export const ApplicationAuthenticated = () => {
       return postNewApplicationAuthenticated(phaseId ?? 'undefined', modifiedApplication)
     },
     onSuccess: (data) => {
-      queryClient.invalidateQueries({ queryKey: ['application', phaseId, user?.email] })
+      queryClient.invalidateQueries({ queryKey: ['application', phaseId] })
       setConfirmationMailSent(data.confirmationMailSent)
       setShowDialog('success')
     },


JGStyle (approved):
there's no infinite spinner:

Image

(application not possible since it's a required field, that might be intentional tho)

PR #1906, opened 2026-07-12
`Assessment`: implement privacy deletion

JGStyle:
## ✨ What is the change?

This PR implements the privacy deletion feature for the assessment CPM

It registers the deletion handler using the sdk functinality.

These records are deleted after approval:

As part of the bigger effort described in https://github.com/orgs/prompt-edu/discussions/683

## 🧪 How to Test

As a user with self team allocation data, request deletion:
1. click your user profile in the top right corner
2. click privacy
3. click deletion
4. request a deletion

5. approve the deletion as an admin from the privacy dashboard (Home → Privacy)
Then, verify that the data from assessment cpm was deleted
(f.e. by inspecting the db or the phase in the course sidebar)

## 🖼️  Demo

Deletion flow working, Requesting student on the left, admin on the right:

https://github.com/user-attachments/assets/8e7e0c3f-7087-47cd-b498-dccb65252671

## ✅ PR Checklist

## New Features
- Added privacy data deletion support for assessment-related records.
- Enables removal of assessments, evaluations, completions, action items, feedback, and category assessments.
- Added category assessment data to privacy exports.
- Registered a privacy deletion endpoint alongside existing data export functionality.
- Processes related deletions together to help ensure complete and consistent removal of assessment data.
- Privacy tools now provide broader coverage of assessment information for export and deletion requests.


servers/assessment/db/query/privacy.sql
```diff
@@ -29,3 +29,31 @@ WHERE course_participation_id = ANY($1::uuid[]);
...
+-- name: DeleteAssessmentCompletionsByCourseParticipationIDs :exec
+DELETE FROM assessment_completion
+WHERE course_participation_id = ANY($1::uuid[]);
+
+-- name: DeleteCategoryAssessmentsByCourseParticipationIDs :exec
+DELETE FROM category_assessment
+WHERE course_participation_id = ANY($1::uuid[]);
+
+-- name: DeleteEvaluationsByCourseParticipationIDs :exec
+DELETE FROM evaluation
```
rappm, line 58: You filter only on `course_participation_id`, but all of our schemata also carry the author data.

Just a small heads up because those rows also hold the recipient's data: deleting them erases feedback that belongs to another student. 

Was this a deliberate decision? I'm not 100% sure anymore if we already discussed this.

servers/assessment/privacy/service.go
```diff
@@ -42,3 +43,38 @@ func PrivacyDataExportHandler(c *gin.Context, exp *utils.Export, subject sdkAuth
...
+	defer promptSDK.DeferDBRollback(tx, c)
+	qtx := PrivacyServiceSingleton.Queries.WithTx(tx)
+
+	if err := qtx.DeleteAssessmentsByCourseParticipationIDs(c, ids); err != nil {
+		return err
+	}
+	if err := qtx.DeleteAssessmentCompletionsByCourseParticipationIDs(c, ids); err != nil {
+		return err
+	}
+	if err := qtx.DeleteCategoryAssessmentsByCourseParticipationIDs(c, ids); err != nil {
```
rappm: AI finding:  Adding this delete classifies `category_assessment` as subject-scoped personal data, but the export path does not agree: `PrivacyDataExportHandler` above has no `category_assessment` entry and there is no `GetAllCategoryAssessmentsByCourseParticipationIDs` in `privacy.sql`.

The result is that the per-category free-text `comment` written about a student is erased on a deletion request but never disclosed on an access request. The two lists should be aligned, and doing it in this PR is cheap since the query file is already open.

servers/assessment/privacy/service.go
```diff
@@ -42,3 +43,38 @@ func PrivacyDataExportHandler(c *gin.Context, exp *utils.Export, subject sdkAuth
...
+	ids := subject.CourseParticipationIDs
+
+	tx, err := PrivacyServiceSingleton.Conn.Begin(c)
+	if err != nil {
+		return err
+	}
+	defer promptSDK.DeferDBRollback(tx, c)
+	qtx := PrivacyServiceSingleton.Queries.WithTx(tx)
+
+	if err := qtx.DeleteAssessmentsByCourseParticipationIDs(c, ids); err != nil {
```
rappm: AI finding: All seven errors are returned bare, so a failed erasure logs one generic SDK line (`deletion was not successful, error: ...`) with no indication of which table aborted. Wrapping with `%w` and the table name would make a failed GDPR deletion diagnosable from the logs, which matters here because the admin has to decide whether to retry.

servers/assessment/privacy/service.go
```diff
@@ -42,3 +43,38 @@ func PrivacyDataExportHandler(c *gin.Context, exp *utils.Export, subject sdkAuth
...
+		return err
+	}
+	if err := qtx.DeleteActionItemsByCourseParticipationIDs(c, ids); err != nil {
+		return err
+	}
+	if err := qtx.DeleteFeedbackItemsByCourseParticipationIDs(c, ids); err != nil {
+		return err
+	}
+
+	return tx.Commit(c)
```
rappm: AI Finding: 

Worth thinking through the failure path of this all-or-nothing transaction. Core's `RunDataDeletion` runs `ExecuteCoreDeletion` and `RequestDeletionFromCPM` in parallel `wg.Go` calls, so if any statement here errors or the chain exceeds `cpmDeletionRequestTimeout` (5 minutes), `DeferDBRollback` discards all seven deletes while core has already committed `DeleteStudentByID` and cascaded away the `course_participation` rows.

A retry is then impossible: `PrepareDataDeletion` gets an empty phase list from `GetCoursePhaseTypesForStudent`, and `AssembleSubjectIdentifiers` returns an empty participation list, so no code path can target the data anymore. The assessment history and peer feedback stay in the database permanently, and the subject gets a failure mail with no remediation.

That is arguably a core-side ordering problem rather than something this PR introduced, but this is the first service that can hit it. At minimum it deserves a note; ideally core should not destroy the identifiers until the fan-out has succeeded.

servers/assessment/privacy/service.go
```diff
@@ -42,3 +43,38 @@ func PrivacyDataExportHandler(c *gin.Context, exp *utils.Export, subject sdkAuth
 
 	return nil
 }
+
+func PrivacyDataDeletionHandler(c *gin.Context, subject sdkAuth.SubjectIdentifiers) error {
+	ids := subject.CourseParticipationIDs
+
+	tx, err := PrivacyServiceSingleton.Conn.Begin(c)
```
rappm: AI Finding: The `*gin.Context` is passed as the `context.Context` for `Begin`, all seven `Exec` calls, `Commit` and the deferred rollback. `main.go:122` uses `gin.Default()` without setting `ContextWithFallback = true`, so `c.Done()` returns nil and no cancellation or deadline ever reaches pgx.

If core times out and retries the deletion, the first transaction keeps running and holding row locks on `assessment`, `evaluation` and `feedback_items`, and both the retry and any concurrent lecturer save block on them. Deriving a `context.WithTimeout` from `c.Request.Context()` here would bound it.

servers/assessment/privacy/service.go
rappm: I think this emerged due to the category free text comment being added after the access request was already there, but we did not add it.

rappm (changes requested):

servers/assessment/db/query/privacy.sql
JGStyle: You're right - maybe we anonymize the author ?
Easy option of course would be to also delete by author

What's your call?

JGStyle: The first option would require more changes - author_course_participation_id is currently NOT NULL and the UNIQUE (course_participation_id, course_phase_id, competency_id, author_course_participation_id) could also become problematic then

JGStyle: I would say the second option would be the better bet (also is more safely compliant with GDPR)

servers/assessment/privacy/service.go
JGStyle: fair, so we just add it to the export as well? I could quickly also do this in this PR

JGStyle: quickly added in a762fdd7, only ~6 lines

servers/assessment/db/query/privacy.sql
rappm: sounds good to me, I think this also makes more sense from a users perspective. Then "Delete" really deletes all associated data with this user

servers/assessment/privacy/service.go
JGStyle: Fair, I improved the error messages in the latest commit 0eb0b58e

servers/assessment/privacy/service.go
JGStyle: It's a good finding, this has been a bit ignored until now. I though through it and i think the best fix is to store the subject identifiers on the deletion record. This would best need a core change, tracked in #2024

servers/assessment/privacy/service.go
JGStyle: better to fix it then to leave it in, fixed it in 2b59cd6e (eventhough should normally not happen )

servers/assessment/db/query/privacy.sql
JGStyle: okay great, implemented in e8286551

servers/assessment/db/sqlc/privacy.sql.go
```diff
@@ -138,6 +208,50 @@ func (q *Queries) GetAllAssessmentsByCourseParticipationIDs(ctx context.Context,
 	return items, nil
 }
 
+const getAllCategoryAssessmentsByCourseParticipationIDs = `-- name: GetAllCategoryAssessmentsByCourseParticipationIDs :many
+SELECT id, category_id, course_phase_id, course_participation_id, comment, created_at, updated_at
+FROM category_assessment
+WHERE course_participation_id = ANY($1::uuid[])
```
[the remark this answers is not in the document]
JGStyle: now implmemented in 2eaa3abdd423ce394616964b2258fb63adb94308

JGStyle:
if present, the query for all category assessments used by the export query now removes the (by: authorname)
text from the category assessment, if it's present. 
This was apparently added in an older migration from old comments and would otherwise leak the name. I verified it works

rappm:
Would look good to me now, but the pipeline fails

rappm:
@JGStyle

rappm (approved):

PR #1924, opened 2026-07-15
`Core`: Delete Keycloak groups and roles on course deletion

### LINKED ISSUE #184: Course Deletion Keycloak Group Deletion
On deleting a course, it should also delete all associates keycloak roles and groups.

rappm:
# 📝 Pull Request Template

## ✨ What is the change?

Deleting a course now also removes the Keycloak artifacts that were created for it. `CreateCourseGroupsAndRoles` builds a course group (`{semester}-{name}`) with `Lecturer`/`Editor` subgroups and the matching client roles, plus custom-group roles (`{group}-cg-{name}`), but none of these were ever cleaned up, so deleted courses left orphaned groups and roles behind in the realm.

`DeleteCourse` now calls a new `keycloakRealmManager.DeleteCourseGroupsAndRoles`, which:

- deletes the course group (Keycloak cascades its subgroups and group-role mappings), and
- deletes the course's client roles: the exact `-Lecturer`/`-Editor` staff roles plus any `-cg-` custom-group roles.

The Keycloak cleanup runs **before** the DB row is removed (the group name is derived from the course record) and is idempotent: a group that is already gone is treated as success. If cleanup fails, the course is kept so it stays deletable on a later retry instead of half-deleting.

Closes #184. Course deletion previously removed only the database rows, leaving the associated Keycloak roles and groups in place forever.

## 🧪 How to Test

1. Create a course, then delete it (as a `PromptAdmin`).
2. In the Keycloak admin console, confirm the `Prompt/{semester}-{name}` group (and its `Lecturer`/`Editor`/`CustomGroups` subgroups) is gone.
3. Confirm the `{semester}-{name}-Lecturer`, `-Editor`, and any `-cg-*` client roles are gone.
4. Automated: `cd servers/core && go test ./course/ ./keycloakRealmManager/`. `TestIsCourseClientRole` covers the role-matching logic, including that a course whose name is a prefix of another's is not caught by mistake.

## ✅ PR Checklist

### DIFF
--- a/servers/core/course/main.go
+++ b/servers/core/course/main.go
@@ -16,6 +16,7 @@ func InitCourseModule(routerGroup *gin.RouterGroup, queries db.Queries, conn *pg
 		queries:                    queries,
 		conn:                       conn,
 		createCourseGroupsAndRoles: keycloakRealmManager.CreateCourseGroupsAndRoles,
+		deleteCourseGroupsAndRoles: keycloakRealmManager.DeleteCourseGroupsAndRoles,
 	}
 
 	// possibly more setup tasks
--- a/servers/core/course/router_test.go
+++ b/servers/core/course/router_test.go
@@ -45,12 +45,16 @@ func (suite *CourseRouterTestSuite) SetupSuite() {
 		// No-op or add assertions for test
 		return nil
 	}
+	mockDeleteGroupsAndRoles := func(ctx context.Context, courseID uuid.UUID) error {
+		return nil
+	}
 
 	suite.cleanup = cleanup
 	suite.courseService = CourseService{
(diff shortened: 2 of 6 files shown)


JGStyle (approved):
Nice.

worked through the code and tested locally.

Deleting course test2 with semestertag test2

| what | before | after |
| --- | --- | --- |
| Keycloak groups now missing test2-test2 | Image | Image |
| Keycloack user roles now missing test2 | Image | Image|

JGStyle (commented):
Minor: during deletion the delay is sometimes a bit higher than expected and there is currently no loading state displayed → we could consider adding it here or separately

PR #1925, opened 2026-07-15
`Interview`: add slot series creation and cloning

### LINKED ISSUE #1186: Allow interview slot series and slot cloning
In the Interview module, you have to create each interview slot by hand. This is tedious; for 10 slots you need to type the target date 20 times, and enter the times for each slot manually. You have to paste the location 10 times as well.

### Quick Wins
- The "end date" should only be a time (hh:mm). It's not a real use case to have an interview that spans two or more days (imo). That saves us from entering 50% of the dates
- Add a "clone" button on each slot that just takes the data of an existing slot and adds it to the modal

### More complex
- Add a series feature where you can create N slots in a specific time range with a specific duration at once

rappm:
# 📝 Pull Request Template

## ✨ What is the change?

Speeds up creating interview slots in the Interview Schedule Management view:

- **End time is now a time-of-day (`HH:mm`)** combined with the start date on submit, so the date only has to be entered once per slot. An end at or before the start is resolved as the next day, and the dialog shows the resolved end (`Ends Tue, Mar 4 at 01:00 (next day)`) so overnight slots stay visible and remain editable.
- **Clone action** on each slot pre-fills the create dialog with that slot's data (time, location, capacity).
- **Create multiple slots** checkbox in the create dialog divides the time range into back-to-back slots of a fixed duration, with an optional break. The dialog previews how many slots will be created and warns when the range exceeds the limit of 100.

The series is generated client-side and created through a new **transactional batch endpoint** (`POST /interview-slots/batch`), so a rejected slot never leaves a partial series behind and creating 100 slots is one request rather than 100 parallel ones.

Creating slots one by one is tedious: for 10 slots you re-enter the date 20 times and paste the location 10 times. This addresses all three suggestions from the issue (time-only end, clone, series).

Closes #1186

## 🧪 How to Test

1. Open a course's Interview phase → **Interview Schedule Management**.
2. **Create Slots**: the End Time field is now a time picker and the slot is created on the start date. Verify the resolved end line below the field, and that entering an end before the start shows `(next day)`.
3. Click the **clone** (copy) icon on an existing slot → the create dialog opens pre-filled with its data.
4. Tick **Create multiple slots**: set a slot duration and optional break. The dialog previews how many slots will be created; confirm and verify they all appear in the table.
5. Verify the batch is atomic: with an invalid range the request is rejected and no partial series is created.

| before | after |

## ✅ PR Checklist

## Summary by CodeRabbit

* **New Features**
  * Create a single interview slot or generate multiple slots from a time range.
  * Configure slot duration and breaks, preview generated slots, and create up to 100 slots at once.
  * Clone existing interview slots with prefilled details.
  * Support overnight scheduling and clearer time-of-day handling.

* **Bug Fixes**
  * Batch creation validates all slots and prevents partial creation when errors occur.

* **Tests**
  * Added coverage for slot series creation, cloning, validation, limits, and persistence.


JGStyle (changes requested):
Image

works as expected and seems like a nice feature, code also looks fine to me


just for the naming : To me Create Series would feel like i'm creating a different kind of 'thing' in the system, like a repeating interview or something

I would maybe find it easier if we just call it 'Create Slots' and 'Create Slot' or maybe even better:

Just have a single create slots button and a single form
The form could just have a check mark at the bottom: 

'create multiple slots'
which just makes the two extra fields needed appear for the slot series


If you disagree, feel free to close this idea, happy to approve 👍

mathildeshagl (changes requested):
Useful feature, and the typed error handling in here is a nice cleanup. Requesting changes on two correctness issues and the missing test coverage for both new features.

Line references are to the PR head, in `clients/interview_component/src/interview/pages/ScheduleManagement/InterviewScheduleManagement.tsx` unless noted.

**1. `:381` / `:83` — slots ending on a different day become permanently uneditable.** `handleEditClick` truncates the stored end to `HH:mm`, and `buildSlotTimes` re-anchors it to the start's date. For any existing slot that crosses midnight — the API and DB still allow them, and the old `datetime-local` UI could create them — end is now before start, so the validation message shows and Update stays disabled forever. The lecturer cannot even fix that slot's location or capacity. The same trap applies to clone at `:393`. A fallback that rolls the end to the next day when end <= start, or an explicit guard, would cover it.

**2. `:233` — total failure runs the success path.** `Promise.allSettled` never rejects, so even when `created === 0` the `onSuccess` path fires: the dialog closes, `resetForm()` wipes the input, and the rejection reasons are discarded (only counts are surfaced). There is no way to retry just the failed slots, and the mutation's `onError` at `:245` is effectively dead code.

**3. `:220` — up to 100 unthrottled parallel POSTs.** There is no bulk endpoint, no concurrency limit, and no abort on unmount, so each series is a burst against the interview service. Chunking, running sequentially, or adding a server-side bulk create would all work.

**4. No tests for series generation or cloning.** `generateMultipleSlots` at `:98` is pure and trivially unit-testable (count math, break handling, cap) but untested, and the e2e diff only relabels the existing single-slot test. Related: `InterviewPage.ts:42-49` embeds checkbox show/hide assertions inside the shared `createSlot()` page-object helper, where they do not belong — those deserve their own test.

**5. `:358`** — `generateMultipleSlots(formData)` runs unmemoized on every render, including on every location and capacity keystroke and while the checkbox is off. `useMemo` would fix it.

**6. `:106`** — `MAX_MULTIPLE_SLOTS` truncates silently, so a 150-slot range quietly yields 100 with no warning to the user.

**7. `:104-108`** — millisecond stepping drifts across DST transitions, shifting wall-clock times by an hour.

**8. `:59`** — `SlotFormData` conflates three different forms, which is why the dummy `durationMinutes: 30, breakMinutes: 0` is repeated at three call sites.

**9.** The PR description is stale: it describes a separate "Create Series" dialog, but the implementation is a checkbox, so "How to Test" step 4 cannot be followed as written. Also `:477` leaves a vestigial `flex gap-2` wrapper around a single button.

Genuinely good: five `onError: (error: any)` handlers replaced with typed `unknown` plus `isAxiosError<ErrorResponse>`, matching `TutorImportDialog.tsx` and satisfying the no-`any` rule; `slotRequestBody` de-duplicates the create and update payloads; the clone button carries both `aria-label` and `title`; and the backend `CreateInterviewSlot` already validates end > start as a real backstop.

mathildeshagl (approved):
Both correctness issues are fixed, and the missing coverage is now there.

1. `buildSlotTimes` rolls the end to the next day when it lands at or before the start, so an existing cross-midnight slot is editable again, and `formatResolvedEnd` shows the resolved end with an explicit `(next day)` marker in both the create and edit dialogs. Clone goes through the same `slotFormFromSlot` helper, so it is covered too.
2. The `Promise.allSettled` fan-out is replaced by `POST /interview-slots/batch`, which runs the whole series in one transaction with `DeferDBRollback`. A rejected slot leaves nothing behind, and `onError` is real code now. This also resolves (3): one request instead of up to 100.
4. `slot-series.spec.ts` covers series creation, cloning, and the checkbox visibility as its own test rather than buried in the `createSlot()` page object, and the Go suite covers the happy path, the atomic rejection, and the batch cap.
5. `seriesPreview` is memoized and skipped entirely while the checkbox is off.
6. `truncated` is surfaced with a destructive warning naming the limit.
7. `addLocalMinutes` steps in wall-clock minutes, so a DST transition inside the range no longer shifts the slot times.
8. `SERIES_DEFAULTS` plus `emptySlotForm` / `slotFormFromSlot` removes the repeated dummy values; keeping one form type is fine at this size.
9. The description now matches the checkbox implementation and the vestigial wrapper is gone.

I checked that `CreateInterviewSlotsBatch` validates end-after-start exactly as `CreateInterviewSlot` does, and that the batch route carries the same roles as the single-slot route.

One thing to be aware of rather than change: with the next-day rollover, `isTimeRangeValid` can never be false once both fields are set, so a lecturer who mistypes the end time gets a ~23-hour slot instead of a validation message. The resolved-end line is what catches that now, which I think is the right trade for keeping overnight slots usable.

CI is green.

PR #1926, opened 2026-07-15
`Core`: Add down migrations to the core database

### LINKED ISSUE #887: Add Down Migrations to Core DB

rappm:
# 📝 Pull Request Template

## ✨ What is the change?

Adds the missing down-migrations to the core service database. `servers/core/db/migration/` had up-migrations `0001`–`0025` but only `0020` shipped a `.down.sql`, so the core DB could not be rolled back the way the other services (`assessment`, `team_allocation`) can.

This PR adds a `.down.sql` for every incremental migration `0002`–`0019` and `0021`–`0025`. Following the existing convention, the base schema `0001` gets no down (matching `assessment`/`team_allocation`), and the one genuinely irreversible structural migration, `0012` (which permanently drops the `course_phase_participation` surrogate key), also gets no down — documented in `0012_course_phase_participation_id.NOT_REVERSIBLE.md`, the same way `team_allocation/0008` is left without one. The practical rollback floor is therefore version 12, which is not a limitation for real databases (all far past it).

Schema-only migrations get faithful reversals. Data/structural migrations (`0005`, `0010`, `0011`, `0024`, `0025`) revert the schema shape with an inline comment noting that data is not restored. `0023` (an `ALTER TYPE ... ADD VALUE`) is a documented no-op because PostgreSQL cannot remove enum values safely, and the type is dropped wholesale by `0022.down` on a full rollback. A few reversals (`0004`, `0015`, `0025`) can fail on pre-existing incompatible data; each says so in a comment.

Closes #887. Without down-migrations, core migrations cannot be rolled back during development or after a bad deploy, unlike every other service.

## 🧪 How to Test

Verified with `migrate` (golang-migrate v4) against a throwaway Postgres in two independent sub-ranges (they bracket the intentional `0012` gap):

- **High range:** `migrate up` (to 25) → `migrate goto 12` (runs `0025.down`…`0013.down`) → `migrate up` (back to 25).
- **Low range:** fresh DB → `migrate goto 11` → `migrate goto 1` (runs `0011.down`…`0002.down`) → `migrate goto 11`.

Both cycles complete without error, confirming every new down reverses its up and the schema re-applies cleanly.

## ✅ PR Checklist


JGStyle (approved):
manually verified the latest 4 migrations (privacy migrations) + another AI verification of the first migrations up to that.

PR #1929, opened 2026-07-15
`Assessment`: Add sticky evaluation headers

### LINKED ISSUE #1915: Use the sticky assessment header also in the Self/Peer/Tutor Evaluation
Currently the header in the Self, Peer and Tutorevaluation is static and stays at the top, not showing critical information. 
The same sticky header from the assessment page should also be used: 
- on the self evaluation page (when the students fill it out)
- on the self evaluation "details" page (when an instructor reviews the self evaluation of a student)
- on the peer evaluation page (when the students fill it out)
- on the peer evaluation "details" page (when an instructor reviews the self evaluation of a student)
- on the tutor evaluation page (when the students fill it out)
- on the tutor evaluation overview page => magnified on a specific pariticipant (when an instructor reviews the results of a tutor in their evaluations from the students)

rappm:
# 📝 Pull Request Template

## ✨ What is the change?

- Extract the assessment page's fixed-position docking behavior into a reusable `StickyHeader`.
- Add a compact `EvaluationHeader` and use it on self, peer, and tutor evaluation forms and instructor result views.
- Keep tutor result navigation available in the docked header and preserve the existing assessment participant header behavior.

Closes #1915.

Evaluation titles currently scroll out of view on the self, peer, and tutor evaluation pages. Docking them beneath the global management header keeps the active evaluation context visible while reviewing or entering longer evaluations.

## 🧪 How to Test

1. Open a self, peer, or tutor evaluation as a student and scroll the evaluation content.
2. Confirm the large page title condenses into a centered sticky title directly below the global header.
3. Open the self/peer instructor participant details and the magnified tutor results, then confirm the same behavior; tutor navigation buttons should remain usable.
4. Resize to a narrow viewport and verify the docked title remains contained and readable.
5. Run `yarn tsc -p assessment_component/tsconfig.json --noEmit --pretty false`, `yarn biome check assessment_component/src`, and the assessment production build from `clients`.
6. Run `make lint` and `docker compose -f docker-compose.e2e.yml --env-file e2e/.env.e2e run --rm e2e-runner npx playwright test tests/assessment/self-evaluation.spec.ts`.

`make test` was also run. All suites before `servers/self_team_allocation/team` passed; that unrelated existing suite reproducibly fails three router status assertions (expected 200/400, received 500) without any files in that service changed by this PR.

Verified locally in authenticated Chromium at 1280×720 and 390×700. The fixed header position, 56 px shell offset, compact title class, and responsive width were asserted during the E2E run; screenshots were kept as local verification artifacts rather than added to the source diff.

## ✅ PR Checklist

## Summary by CodeRabbit

* **New Features**
  * Added consistent sticky headers across assessment and evaluation pages.
  * Evaluation headers now support optional previous and next navigation.
* **UI Improvements**
  * Standardized page title layouts across evaluation views.
  * Assessment student details collapse when the header is docked.
  * Navigation controls appear independently when a previous or next participant is available.
* **Refactor**
  * Consolidated header docking behavior into a shared component.


magkue (commented):
Nice refactor — pulling the fixed-position docking out of `AssessmentHeader` into a reusable `StickyHeader` (render-prop) plus an `EvaluationHeader` wrapper reads well, and the expanded title keeps the standard `text-4xl font-bold` styling so it stays consistent with `ManagementPageHeader`. The `print:!static print:!w-auto` guard for the tutor-results print is a nice touch. I ran it locally on the self-evaluation page; the docking works. Two things:

### 1. `ResizeObserver` feedback loop (should-fix)

`StickyHeader.tsx` newly adds a `ResizeObserver` on the bar:

```js
const resizeObserver = new ResizeObserver(update)
resizeObserver.observe(bar)
```

…but `update()` mutates that same observed element inside the callback (`bar.style.position/left/width`, and `bar.style.cssText = ''` on undock). Changing the bar's size/position reflows it, which re-fires the observer, so the browser emits:

> ResizeObserver loop completed with undelivered notifications.

In dev this surfaces as an rspack error overlay on the evaluation pages. Note this is new behavior in this PR — the old `AssessmentHeader` used only `scroll`/`resize` listeners, no `ResizeObserver`.

Suggested fix: break the synchronous cycle, e.g.

```js
const resizeObserver = new ResizeObserver(() => requestAnimationFrame(update))
```

or observe the `placeholder` (which `update` doesn't resize) rather than the `bar`.

### 2. Nav-button visibility — behavior change worth confirming

In `AssessmentHeader`, the previous code gated both nav buttons behind `hasNavigation = Boolean(prevMember && nextMember)` (all-or-nothing). This PR renders `prevMember && (...)` and `nextMember && (...)` independently, so on the first/last participant the single available direction now shows. This looks like an intentional improvement, but since it's a functional change bundled into a sticky-header refactor it'd be good to confirm and maybe call it out in the PR description.

(Separately, while testing as a lecturer I hit an unrelated `my-feedback` 401 → "Error loading feedback items"; that's a pre-existing issue with viewing a student page as a non-student, not caused by this PR. Filing it separately.)

JGStyle:
Works for me as expected

2026-07-23 at 08 16 48

mathildeshagl (commented):
Clean extraction of logic that was genuinely duplicated, and the `prevMember &&` / `nextMember &&` simplification is behavior-preserving since `useParticipantNavigation` returns both or neither. A few things worth a look, none of them blocking.

**1. `clients/assessment_component/src/assessment/pages/components/StickyHeader.tsx:40` + `EvaluationHeader.tsx:26-38` — docking causes a ~24px content jump and can oscillate.** The header shrinks on dock (`text-4xl` + `mb-6`, roughly 64px, down to `h-10` = 40px), but `placeholder.style.height = bar.offsetHeight` is written before React applies the docked classes. The new `ResizeObserver` then corrects it a frame later and everything below shifts up. Because that shift (24px) exceeds `UNDOCK_HYSTERESIS_PX` (8), a scroll container near its bottom can clamp `scrollTop`, undock, re-dock, and flicker. Freezing the placeholder at the measured undocked height would remove both the shift and the loop risk. Worth noting the original `AssessmentHeader` did not have this problem because its bar height was constant.

**2. `StickyHeader.tsx:33-58` — `update()` runs unthrottled on every capture-phase scroll event**, doing a `getBoundingClientRect` read followed by inline style writes and an unconditional `setDocked(...)` each frame. That now applies to long evaluation forms rather than just the assessment page. Guarding with `if (dockedRef.current !== shouldDock)` and rAF-throttling scroll the way resize already is would keep it cheap.

**3. `EvaluationHeader.tsx:31-38`** — re-implements `ManagementPageHeader` (`text-4xl font-bold mb-6`) locally, so these pages silently diverge if the shared component changes (`.claude/rules/react-typescript/shared-libraries.md` says reuse first). Also `mb-6` is applied only when there are no nav actions, so vertical rhythm ends up depending on whether prev/next buttons exist.

**4. No tests in the diff** although the PR checklist marks "Tests added or updated"; `e2e/tests/assessment/self-evaluation.spec.ts` is untouched, so nothing guards the docking or print behavior.

**5. `StickyHeader.tsx:81`** — `print:!static print:!w-auto` uses Tailwind v3 prefix-important syntax. I compiled it against the repo's Tailwind 4.3.3 and it does still emit, so this is not a bug, but repo convention is the v4 suffix form (for example `w-3!` in core). Separately, the placeholder's inline `height` survives into print, leaving a stray gap.

**6. `AssessmentHeader.tsx`** — the relocated "Name + badges" comment kept its old continuation indentation.

ResizeObserver plus rAF is a real improvement over the original, aria-labels are preserved, and the props typing (no `any`, `interface`, `cn`, correct relative imports) all follow the rules.

mathildeshagl (approved):
All six points addressed in `b5a97371`.

1. `undockedHeightRef` freezes the placeholder at the undocked height, so the content below no longer jumps by the shrink delta and the shift can no longer exceed `UNDOCK_HYSTERESIS_PX` and drive the dock/undock oscillation. The height is captured on every undocked pass and on mount, so a genuine resize while undocked is still picked up.
2. Scroll and resize now share one rAF-coalesced `scheduleUpdate`, and `setDocked` only fires on an actual transition, so the capture-phase scroll handler is no longer doing a layout read plus a state write per event.
3. Fair enough on not reusing `ManagementPageHeader` — it takes no `className` and this title needs a condensed docked variant, and the comment says so. Moving the gap to `pb-6` on the row also fixes the part I actually cared about: the vertical rhythm no longer depends on whether prev/next buttons exist.
4. The e2e test is a good one. Asserting the placeholder height is unchanged across dock and undock tests the actual fix rather than just that the header sticks, and the guard that something really overflowed keeps it from passing vacuously.
5. Converted to the v4 suffix form, and `print:h-auto!` on the placeholder drops the inline height so the printed page has no stray gap.
6. Comment indentation fixed.

CI is green.

PR #1939, opened 2026-07-16
`New Phase`: Add presentation scheduling, materials, and instructor feedback

### LINKED ISSUE #1931: Presentation phase (scheduling, material upload, multi-instructor feedback)
## 🚀 Feature Request Description

A new course phase for **presentations**, supporting either **team** or **single-student** presentations. It would cover the full presentation lifecycle: scheduling slots, uploading slides / supporting material, and collecting structured instructor feedback.

## 🧩 Problem Case

Presentations are a common part of project-based courses, but PROMPT 2.0 has no dedicated phase for them. Scheduling, material handling, and feedback are currently done ad-hoc outside the platform, so there is no single place where students submit materials and instructors record and share evaluations.

## 💡 Proposed Solution

A presentation course phase that provides:

- **Scheduling** — assign presentation slots to teams or individual students.
- **Material upload** — students upload slides and supporting material for their slot.
- **Instructor feedback** — instructors enter feedback per presentation, with:
  - **Multiple independent instructors / editors** evaluating the same presentation, this should support them editing the same session in parallel and an independent mode, editable in the setting. 
  - A **predefined set of feedback categories** to structure the evaluation.

## 🎨 Alternative Solutions

- Reuse/extend the existing assessment phase for feedback instead of a dedicated phase (loses the scheduling + upload flow).
- Handle scheduling and materials externally and only capture feedback in PROMPT (fragmented UX).

## 📝 Additional Context

Both team and single-student modes should be supported. Feedback from independent instructors/editors should be captured separately (not merged into one shared form) so evaluations stay independent.

rappm:
# 📝 Pull Request Template

## ✨ What is the change?

Adds a dedicated Presentation course phase covering the complete presentation lifecycle:

- staff-managed scheduling for individual students or teams, including overlapping slots and optional locations
- multi-file presentation material uploads through S3-compatible storage, with student deadlines at slot start
- predefined written feedback categories and independent or shared multi-instructor evaluation
- author-private drafts, submitted staff-visible evaluations, optimistic category revisions, and SSE live collaboration
- named feedback release, unrelease, revision, and re-release flows
- optional `teams` and `teamAllocation` phase inputs with dynamic team membership authorization
- Module Federation, Docker, CI, deployment, e2e, and documentation integration

The implementation is split into focused commits for core contracts, the React remote, the Go service, infrastructure, integration testing, and documentation.

Presentation scheduling, materials, and instructor feedback were previously handled outside PROMPT, resulting in fragmented workflows and no central record of evaluations.

Closes #1931.

## 🧪 How to Test

1. Copy `.env.template` and `.env.dev.template` to the local environment files and start the stack with `docker compose up --build`.
2. Add a **Presentation** phase in the Course Configurator. Leave the optional team inputs disconnected for individual mode, or connect `teams` and `teamAllocation` for team mode.
3. As a course lecturer, configure the target mode, feedback mode, and feedback categories.
4. Create overlapping or sequential slots, optionally add locations, and assign students or teams.
5. As an assigned student or team member, upload multiple presentation files before the slot starts and verify download/delete behavior.
6. As multiple editors, enter independent feedback or edit a shared form concurrently; verify live presence and optimistic conflict handling.
7. Submit all feedback, release it with a name, verify student visibility, then unrelease, revise, and release again.
8. Run `go test ./...` in `servers/core`, `servers/team_allocation`, and `servers/presentation`.
9. Run `yarn workspace client typecheck`, `yarn workspace client build`, and the Presentation remote's TypeScript, Biome, and production Rspack checks from `clients/`.
10. Run `docker compose --env-file e2e/.env.e2e -f docker-compose.e2e.yml up --build --abort-on-container-exit --exit-code-from e2e-runner` to exercise the Presentation Module Federation smoke test.

Screenshots are not attached yet; this is intentionally opened as a draft while deployment-level review is pending.

| Before | After |
| ------ | ----- |
| N/A | Presentation overview, schedule, settings, materials, and feedback workspaces |

## ✅ PR Checklist


mathildeshagl (approved):
Approving. This is a lot of surface, and the parts I would expect a new phase to get wrong are right.

What I checked and found sound:

- **Routing and auth.** Everything sits under `/presentation/api/course_phase/:coursePhaseID`, and every single route carries `promptSDK.AuthenticationMiddleware` with a deliberate tier rather than a copy-pasted role list. No unguarded route.
- **Authorization.** `getPresentationAndAuthorize` is a single choke point and is actually used on every material and feedback read and write. Non-staff must match the presentation target; team mode resolves the caller's own team through the SDK rather than trusting a client-supplied id. `DeleteMaterial` additionally restricts presenters to their own uploads, which is the right call and the comment explains why. I went looking for an IDOR on the material endpoints, since they are open to `allRoles`, and did not find one.
- **Upload lifecycle.** The presign binds content type, and `CompleteUpload` then re-verifies the stored object's size and content type and deletes on violation. "The stored object decides the type, not the presign request" is exactly the right instinct, and it is strictly more careful than what core does today.
- **Migrations.** `0027` is the next free number on main, up and down are symmetric, and `db/sqlc/` is regenerated and committed along with the database dumps and swagger. That is the part that usually rots.
- **SSE.** Buffered channel with a non-blocking publish so a slow client cannot block the publisher, unsubscribe deferred on return, keepalive ticker, and exit on request-context cancellation. No leak, and no lock nesting in `EventHub` despite `publishPresence` calling both `ActiveEditors` and `Publish`.
- **Privacy.** Both handlers implemented, and the comment about why shared feedback is exported per record rather than per form, so co-instructors' text is not disclosed, is the kind of reasoning I would rather see written down than rediscovered.
- **`AssignTarget`.** Read-before-write plus unique-constraint fallback mapped to specific 409s, with a comment saying why both exist.
- **Frontend.** No `any` anywhere in the component, and it reuses the shared UI and state packages.
- **CI.** `presentation` added to the quality-servers, test-servers, and quality-clients matrices, so none of this goes unbuilt or untested. DB-backed tests for slots, materials, and feedback.

**The main thing I want to flag is merge order, not code.** This branch predates three cleanup PRs that are removing exactly the patterns it introduces, and none of them contain `presentation_component`:

| In `presentation_component` | Removed by |
| --- | --- |
| `network.ts` hand-rolls `axios.create` plus a JWT request interceptor | #2068, which deletes that same block from six components |
| `routes/index.tsx` and `sidebar/index.tsx` use literal `[Role.PROMPT_ADMIN, Role.COURSE_LECTURER]` arrays | #2069, where that is `LECTURER_ROLES` |
| `rspack.config.mjs` (93 lines), `src/App.tsx`, `src/bootstrap.tsx` | #2071, which extracts all of it into `clients/shared` |

Nobody's mistake, just ordering. The cheapest sequence is those three first, then one commit here adopting all of them. For the #2071 part, note that the shared `mountRemote` does not import styles, so keep the `./styles.css` import when you collapse `bootstrap.tsx`.

Non-blocking, in rough priority order:

- **Presigned PUTs do not bind `ContentLength`.** A presign can therefore be used to PUT an object of any size; the cap only applies at `CompleteUpload`, and simply never calling complete leaves the object in the bucket until the hourly reaper sweeps it. Before writing this up I checked core: `servers/core/storage/s3_adapter.go` has the identical pattern, so this is a pre-existing repo-wide gap that the new service mirrors rather than something this PR introduces, and this service is already stronger than core here. Worth a follow-up issue covering both. The fix is one field on the `PutObjectInput`, threading the requested size through `GetUploadURL`.
- **`streamFeedbackEvents` calls `queries.GetPresentation` inline** instead of going through `getPresentationAndAuthorize`. Equivalent today, because the route is staff-only and the helper's extra check only fires for non-staff, so this is correct rather than buggy. But the invariant is incidental instead of structural: if that role list is ever widened to students, say to let them watch released feedback arrive, the gap becomes real and nothing will complain. Routing it through the same helper would cost nothing and keep the property enforced in one place.
- `EventHub.ActiveEditors` uses `time.Now()` directly while the service has an injectable `s.now()` for testability.
- `HandlePhaseCopy` deletes the target's S3 objects fire-and-forget after commit. A failure there orphans objects the reaper will not reclaim, since their rows are already gone.
- The description still says this is "intentionally opened as a draft while deployment-level review is pending", but it is not a draft. Worth updating, along with the screenshots checkbox, given this adds five new UI pages.

Nice work, particularly the ambiguity-over-guessing and privacy handling.

PR #1961, opened 2026-07-21
`Application`: Support CSV student import

Mtze:
## ✨ What is the change?

Adds a per–application-phase **CSV import mode** as an exclusive alternative to the public application form. A lecturer switches the phase's **Intake Mode** to *CSV Import*, which closes the public form and enables importing students from a CSV instead.

- **Backend:** new `POST /api/applications/:coursePhaseID/import` batch endpoint that upserts students, course + phase participations, and (deduped) text questions/answers in one transaction, with a lecturer-chosen pass status. An `applicationMode` gate closes the public apply flow (open-phase queries + the authenticated GET / file-delete routes) when a phase is in import mode.
- **Login for imported students:** `GetStudentRoleStrings` and the Keycloak middleware role guard were relaxed so an imported student **without a matriculation number** still resolves the course-student role on first login (identity is the university login). Creating the participation during import is exactly what grants access on first login.
- **Frontend:** an *Intake Mode* selector (Application Form / CSV Import); in import mode the application timeframe and public-form options are hidden and a dedicated "CSV Import" status is shown. A 4-step import wizard (upload + preview → column mapping with auto-detection → review with status choice → result summary), a CSV template download, and mapping of leftover columns to text application questions.

Some courses decide their participants elsewhere (central matching, spreadsheets) and need to register them in bulk instead of collecting public applications — while still letting those students log in and reach the course.

## 🧪 How to Test

**Automated**
- Go: `cd servers/core && go test ./applicationAdministration/... ./student/... ./keycloakTokenVerifier/...`
- e2e: `make test-e2e` (adds `tests/api/application-import.api.spec.ts` + a seeded import-mode phase)

**Manual (UI)**
1. Open a course's **Application** phase → **Settings** → set **Intake Mode** to *CSV Import* and save (the form-only options disappear; the overview shows a "CSV Import" status).
2. Go to **Participants** → **Import Students**, download the template or use your own CSV. Required columns: First Name, Last Name, University ID, Email. Extra columns can be mapped to attributes or imported as questions.
3. Map columns → review new/updated counts and pick Accepted / Not-assessed → import. Re-importing the same file updates rather than duplicating.

_UI changes are included (Intake Mode selector + import wizard). Screenshots to be added._

## ✅ PR Checklist

---

## Summary by CodeRabbit

* **New Features**
  * Added **CSV Import Mode** with an “Intake Mode” toggle for **Application Form** or **CSV Import**.
  * Added a multi-step wizard to upload CSVs, map columns, preview records, set pass status, and review results.
  * Unmapped columns can become application questions; re-imports update existing students.
  * Application status badges now display **Import** status.
* **Bug Fixes**
  * Import-mode phases are blocked from public application access.
* **Documentation**
  * Added CSV import requirements and workflow guidance.


servers/core/applicationAdministration/service.go
```diff
@@ -991,6 +992,177 @@ func GetAdditionalScores(ctx context.Context, coursePhaseID uuid.UUID) ([]applic
...
+
+		for _, ans := range row.Answers {
+			questionID, ok := questionIDByColumn[ans.ColumnKey]
+			if !ok || strings.TrimSpace(ans.Answer) == "" {
+				continue
+			}
+			answerDBModel := applicationDTO.CreateAnswerText{
+				ApplicationQuestionID: questionID,
+				Answer:                ans.Answer,
+			}.GetDBModel()
```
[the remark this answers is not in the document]
Mtze: Addressed in 82c450dc: import answers are now validated against the resolved question's `allowedLength` before writing (covers re-imports that reuse a shorter existing question too), returning a clear error instead of letting an oversized value through. Added a Go test (`TestImportApplications_AnswerExceedingAllowedLengthRejected`).

servers/core/keycloakTokenVerifier/oidcVerifier.go
```diff
@@ -27,6 +29,16 @@ func InitKeycloakVerifier() error {
...
 
+	// Local-development escape hatch: when the browser reaches Keycloak on a
+	// different host/port than the containerized server (e.g. http://127.0.0.1
+	// vs the internal service name), the token issuer will not match the
+	// server's provider URL. Signatures are still verified against the realm's
+	// JWKS. NEVER enable this in production.
+	if os.Getenv("KEYCLOAK_INSECURE_SKIP_ISSUER_CHECK") == "true" {
+		config.SkipIssuerCheck = true
+		log.Warn("KEYCLOAK_INSECURE_SKIP_ISSUER_CHECK is enabled - token issuer validation is disabled. Do not use this in production.")
+	}
```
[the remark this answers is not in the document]
Mtze: Addressed in 82c450dc: this dev-only escape hatch was accidentally swept into the branch and has been removed from the PR entirely, so it no longer exists in any production path. It is kept only as a local, uncommitted change for browser testing on 127.0.0.1.

rappm:
@Mtze e2e tests are failing after merging main into this

mathildeshagl (changes requested):
Substantial piece of work, and a lot of it is carefully done. Two things I would like resolved before merge: one bug that defeats the feature's headline promise despite the tests passing, and one authorization regression.

**1. `servers/core/db/query/permission_strings.sql:24` — the relaxed `GetStudentRoleStrings` does not actually cover the import case.** `CreateStudent` always writes `pgtype.Text{Valid: true}` (`servers/core/student/studentDTO/create_student.go:28`), so an imported student without a matriculation number gets `matriculation_number = ''`, never `NULL`. When they log in with a real Keycloak token (`$1 = "01234567"`), all three disjuncts are false and they get no course-student role. The stated goal — an imported student without a matriculation number resolves the role on first login — fails in production. `import_test.go:2026` passes `MatriculationNumber: ""`, which exercises only the trivially-true `$1 = ''` branch, which is why this is green. Something like `COALESCE(s.matriculation_number,'') IN ('', $1) OR $1 = ''` plus a test with a non-empty `$1` would cover it.

**2. `servers/core/applicationAdministration/router.go:34` — authorization regression.** Import grants `CourseEditor`, while `postApplicationManual`, `deleteApplications`, and `updateApplicationsStatus` are `PromptAdmin` + `CourseLecturer` only. Import both creates participations and sets pass status in bulk, so editors gain exactly the bulk-admit power that is deliberately withheld from them everywhere else. Was that intentional?

**3. `servers/core/applicationAdministration/service.go:1078` + `servers/core/student/service.go:167` — silent data loss on re-import.** `CreateOrUpdateStudent` preserves only the matriculation number; gender, nationality, studyProgram, studyDegree, and currentSemester are overwritten unconditionally. Importing a student who already exists (a re-import, or a participant of another course) with a CSV lacking those columns silently resets gender to `prefer_not_to_say`, degree to `bachelor`, and nationality/program to `''`.

**4. `service.go:1122`** — the allowed-length check lives in the service rather than `validateApplicationImport`, so `router.go` maps this user-input error to 500 instead of 400. The test asserts only on the service error, which hides it.

**5. `service.go:1086` vs `ImportStudents.tsx:247`** — `outcome` means "student exists anywhere in PROMPT" while the client preview says "login already in this phase", so preview and result counts disagree for students known from other courses.

**6. `validation.go` (`validateApplicationImport`)** — no cap on `len(req.Rows)`, and roughly 6 queries per row run in a single transaction. A large CSV holds a long transaction open.

**7. `clients/.../utils/parseImportCsv.ts:1`** — `papaparse` is undeclared in `clients/core/package.json` and resolves only via hoisting from `matching_component`; `@types/papaparse` is not installed at all, so `Papa` is untyped `any`, masked by `noImplicitAny: false`.

Minor: batch pass status silently resets already-`failed` participations; `matchImportColumns` / `buildImportRequest` coerce unknown gender and degree values to defaults with no warning; `IsImportModePhase` now turns an unknown phase ID into a 500.

Worth saying explicitly what is strong here: the mode gating is applied consistently across every public-apply entry point (both open-phase listings, `GetOpenApplicationPhase`, `CheckIfCoursePhaseIsOpenApplicationPhase`, the authenticated GET, and file delete); the import is genuinely all-or-nothing with `DeferRollback`; re-import is idempotent via question reuse by title plus `CreateOrOverwriteApplicationAnswerText`; there is a 10-case Go suite on a dedicated container plus an e2e API spec; sqlc and Swagger are regenerated and committed; and the user docs are updated.

servers/core/applicationAdministration/validation.go
```diff
@@ -275,6 +275,99 @@ func validateApplicationManualAdd(ctx context.Context, coursePhaseID uuid.UUID,
...
+	if req.PassStatus != db.PassStatusPassed && req.PassStatus != db.PassStatusNotAssessed {
+		return errors.New("invalid pass status for import")
+	}
+
+	// Validate the imported question columns. Titles must be unique too: the import service
+	// resolves questions by title, so two columns sharing a title would collapse onto one
+	// question and silently overwrite each other's answers.
+	questionColumns := make(map[string]bool, len(req.NewQuestions))
+	questionTitles := make(map[string]bool, len(req.NewQuestions))
+	allowedLengthByColumn := make(map[string]int, len(req.NewQuestions))
```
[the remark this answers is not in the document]
Mtze: Addressed in 7dd58e49. `validateApplicationImport` now loads the persisted question lengths via `GetApplicationQuestionsTextForCoursePhase` and enforces the **stored** `AllowedLength` for any reused title, falling back to the request's length only for genuinely new questions - so validation matches exactly what `PostApplicationImport` enforces. This removes both the false-rejection and the pass-here/fail-in-tx paths. Added `TestImportApplications_ReuseUsesPersistedLength`, which re-imports a title whose request limit (100) exceeds the persisted limit (10) and asserts the persisted limit wins.

Mtze:
@mathildeshagl thanks for the thorough review - genuinely useful catches. All addressed in `fc26611c` / `7dd58e49` (plus `7dd58e49` for a related reuse-length edge). Point by point:

**1. `GetStudentRoleStrings` doesn't cover the import case** - fixed. You're right: `CreateStudent` writes `''`, not `NULL`, so the disjunction failed for a real login token. Changed the guard to `COALESCE(s.matriculation_number,'') = '' OR s.matriculation_number = $1 OR $1 = ''`, regenerated sqlc, and strengthened `TestImport_StudentRoleResolvedWithoutMatriculation` to also assert with a **non-empty** `$1` (the production path that was actually broken).

**2. Authorization regression (`CourseEditor`)** - not intentional; fixed. Import now requires `PromptAdmin` + `CourseLecturer`, matching `postApplicationManual` / `deleteApplications` / `updateApplicationsStatus`.

**3. Silent data loss on re-import** - fixed. The import now resolves the existing student and preserves gender / nationality / studyProgram / studyDegree / currentSemester whenever the CSV omits them; defaults only apply to genuinely new students. Added `TestImportApplications_ReImportPreservesOmittedAttributes`.

**4. Length check maps to 500 not 400** - fixed. The check now runs in `validateApplicationImport` (400 pre-tx); the in-tx check for reused-question limits returns a new `ErrImportAnswerTooLong` sentinel that the handler maps to 400. Also (from CodeRabbit's follow-up) validation now uses the **persisted** length for reused titles so it agrees with the service - see `TestImportApplications_ReuseUsesPersistedLength`.

**5. `outcome` vs preview semantics** - fixed. `outcome` is now based on `GetApplicationExistsForStudent` (this course/phase), matching what the client preview counts, instead of global student existence.

**6. No cap on `len(req.Rows)`** - added `maxImportRows = 2000` in validation (400).

**7. `papaparse` undeclared / `@types/papaparse` missing** - fixed. Both declared in `clients/core/package.json` + lockfile. This surfaced a latent `ParseResult` type error in `matching_component` (previously masked as `any`), which I fixed too.

**Also:** merging `main` created a hard PK collision - the import phase and a newly-merged example phase both used `d0000010-…010`, which breaks seed init (`prompt-e2e-db exited (3)`). Moved the import phase to `d0000011-…011`; verified the full merged seed loads into a fresh Postgres with zero errors. This was almost certainly the e2e failure @rappm hit.

**Consciously not changed** (happy to reconsider):
- *Batch pass-status overwriting a previously-rejected participant on re-import* - this is intended: the lecturer explicitly picks the status for the imported batch.
- *`matchImportColumns` / `buildImportRequest` coercing unknown gender/degree to defaults without a warning* - acceptable for v1, and #3 removes the data-loss consequence.
- *`IsImportModePhase` turning an unknown phase ID into a 500* - not reachable on the import endpoint, which 400s first via the application-phase check.

Thanks again - ready for another look.

servers/core/applicationAdministration/service.go
```diff
@@ -991,6 +994,215 @@ func GetAdditionalScores(ctx context.Context, coursePhaseID uuid.UUID) ([]applic
...
+		}
+
+		studentObj, err := student.CreateOrUpdateStudent(ctx, qtx, studentInput)
+		if err != nil {
+			log.Error(err)
+			var pgErr *pgconn.PgError
+			if errors.As(err, &pgErr) && pgErr.Code == "23505" && pgErr.ConstraintName == "student_email_key" {
+				return applicationDTO.ImportResult{}, ErrEmailAlreadyInUse
+			}
+			return applicationDTO.ImportResult{}, fmt.Errorf("could not save student %s: %w", studentInput.UniversityLogin, err)
```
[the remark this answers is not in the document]
Mtze: Good catch - addressed in 3d747c40. `PostApplicationImport` now detects the 23505 on the `student_university_login_unique` partial index (note: the constraint was renamed from `student_university_login_key` to that partial index in migration 0004) and returns a new `ErrUniversityLoginConflict` sentinel, which the handler maps to **409** with a message naming the offending login - instead of the generic 500.

One note on test coverage: the shared `application_administration.sql` test dump intentionally seeds three students that all share `university_login = 'ab12cde'`, so the partial unique index can't be added to that dump (its own seed data would violate it). The conflict path is therefore only reproducible against the full migration/e2e schema, where the index exists. I verified the handling by construction (it mirrors the existing `student_email_key` path).

Also addressed your suppressed note on `student/validation.go`: `ValidateImport` now validates university data as an account holder (`validateUniversityData(true, ...)`) rather than depending on the caller setting `HasUniversityAccount`, matching its documented contract.

servers/core/db/query/permission_strings.sql
```diff
@@ -20,8 +20,8 @@ SELECT CONCAT(c.semester_tag, '-', c.name, '-Student')::text AS student_role
 FROM course c
 JOIN course_participation cp ON c.id = cp.course_id
 JOIN student s ON cp.student_id = s.id
-WHERE s.matriculation_number = $1
-AND s.university_login = $2;
+WHERE s.university_login = $2
+AND (COALESCE(s.matriculation_number, '') = '' OR s.matriculation_number = $1 OR $1 = '');
```
rappm: **Course student roles can now be granted on the university login alone.**

`main` required both identifiers to match (`s.matriculation_number = $1 AND s.university_login = $2`). The new predicate is satisfied by the login alone as soon as the stored matriculation number is empty, and `keycloakTokenVerifier/middleware.go` no longer returns early when the token carries no matriculation claim.

Failure case: a lecturer imports a row whose University ID is mistyped as `ab12cde`, a real but unrelated student's login, and leaves the matriculation column empty (the import allows that). On that student's next login the `COALESCE(s.matriculation_number, '') = ''` branch matches, so they receive the `<semester>-<course>-Student` role for a course they were never admitted to, see it on their dashboard, and get resolved onto the imported participation. Under the old predicate the mismatching matriculation number blocked this.

Two things make this the most severe item in the PR: the rows most likely to have an empty matriculation number are exactly the imported, least-verified ones, and the relaxation applies to every login in the system rather than only to import-mode phases.

Suggested direction: keep role resolution on the strict two-identifier match and require a matriculation number on import instead. If imports genuinely have to work without one, the relaxed branch should be scoped so it can never match a student row that does carry a matriculation number. Could you also spell out in the PR description why the `OR $1 = ''` branch is needed on top of the `COALESCE` branch?

servers/core/applicationAdministration/service.go
```diff
@@ -991,6 +995,220 @@ func GetAdditionalScores(ctx context.Context, coursePhaseID uuid.UUID) ([]applic
...
+		if outcome == "updated" {
+			result.Updated++
+		} else {
+			result.Created++
+		}
+	}
+
+	// 3. Apply the chosen pass status to the whole batch (including pre-existing participations).
+	if len(participationIDs) > 0 {
+		if _, err := qtx.UpdateCoursePhasePassStatus(ctx, db.UpdateCoursePhasePassStatusParams{
```
rappm, line 1271: **Re-importing a roster silently overwrites pass statuses that were set by hand.**

`participationIDs` collects every row in the file, including the ones this function itself classified as `outcome == "updated"`, and `UpdateCoursePhasePassStatus` sets `pass_status` unconditionally for the whole slice.

Failure case: staff import 200 students, then reject 30 of them in the participants table (`pass_status = failed`). A week later they re-upload the corrected CSV, which still contains those 30 rows, and leave the radio button on its default. All 30 rejections flip back to `passed`/`not_assessed`. There is no confirmation, no report of what changed and no undo, and `failed` is not even offered in the dialog, so the status cannot be restored through the import flow either.

Suggested direction: apply the batch status only to participations the import just created, and either leave pre-existing rows untouched or make overwriting an explicit opt-in. Whichever way it goes, the preview step should say how many existing statuses the import is about to change, since it currently labels the control "Status for imported students".

servers/core/student/validation.go
```diff
@@ -39,6 +39,29 @@ func Validate(c studentDTO.CreateStudent) error {
 	return nil
 }
 
+// ValidateImport is a lenient variant of Validate used for CSV student imports.
+// It only enforces the mandatory identity fields (first name, last name, email and university
+// login) plus their formats. Optional attributes (nationality, study program, current semester)
+// and the enum defaults (gender, study degree) are filled in by the import service, so they are
+// not required here.
+func ValidateImport(c studentDTO.CreateStudent) error {
```
rappm, line 47: **`ValidateImport` drops checks that every other student write path enforces, and the import writes to the globally shared `student` row.**

`Validate` rejects `!c.CurrentSemester.Valid || c.CurrentSemester.Int32 < 1`, an empty nationality, an empty study program, and a study degree outside `{bachelor, master}`. `ValidateImport` enforces none of them, while `buildImportRequest` passes `Number.parseInt(cell, 10)` straight through and `PostApplicationImport` persists the values via `CreateOrUpdateStudent`.

Failure case: a CSV whose `Semester` column contains `0` (common in freshman exports) or a negative number stores `current_semester = 0` on the shared `student` row, and that value then follows the student into every other course, profile view and participant export in PROMPT. The same path lets a free-text nationality such as `German` overwrite a previously stored ISO code, because non-empty CSV values are written verbatim rather than preserved-on-empty.

The comment above says the optional attributes are "filled in by the import service", which is true only when the cell is empty. When the cell holds a bad value it is persisted unvalidated.

Suggested direction: validate the optional fields whenever they are present (semester `>= 1`, nationality as ISO-3166 alpha-2, study degree within the enum) and reject the row otherwise. Since the write target is shared across courses, leniency here is not local to this phase.

servers/core/applicationAdministration/service.go
```diff
@@ -991,6 +995,220 @@ func GetAdditionalScores(ctx context.Context, coursePhaseID uuid.UUID) ([]applic
...
+		nextOrder++
+	}
+
+	// 2. Upsert each student, participation and answers.
+	result := applicationDTO.ImportResult{Rows: make([]applicationDTO.ImportRowResult, 0, len(req.Rows))}
+	participationIDs := make([]uuid.UUID, 0, len(req.Rows))
+	for i, row := range req.Rows {
+		studentInput := row.Student
+		studentInput.HasUniversityAccount = true
+		studentInput.UniversityLogin = strings.ToLower(strings.TrimSpace(studentInput.UniversityLogin))
```
rappm, line 1150: **Lowercasing the login here does not match the case-sensitive lookups used everywhere else.**

`GetStudentByUniversityLogin`, `GetStudentByMatriculationNumberAndUniversityLogin` and `GetStudentRoleStrings` all compare `university_login` case-sensitively, and the partial unique index `student_university_login_unique` is on the raw column. The manual-add path stores whatever case the lecturer typed, since `validateUniversityData`'s regex `^[a-zA-Z]{2}\d{2}[a-zA-Z]{3}$` accepts uppercase.

Failure case: a student already exists as `GE42ABC` from the manual dialog. The import normalizes the same person to `ge42abc`, `ResolveStudentByUniversityCredentials` finds nothing, and `CreateStudent` runs. If the CSV email matches the stored one, `student_email_key` fires and the whole batch aborts with 409 `email already in use`. If the email differs, a second `student` row is created for the same person, and because the Keycloak claim still reads `GE42ABC`, `GetStudentRoleStrings` matches only the old row, so the student never receives the role for the course they were just imported into.

Suggested direction: either normalize case at the boundary consistently (a `lower(university_login)` unique index plus case-insensitive lookups, applied to the manual path too) or stop normalizing here. A normalization that only one write path performs is worse than none.

clients/core/src/managementConsole/applicationAdministration/pages/ApplicationParticipantsPage/components/ImportStudents/utils/buildImportRequest.ts
```diff
@@ -0,0 +1,72 @@
...
+    const student: ImportStudent = {
+      firstName: valueFor('firstName'),
+      lastName: valueFor('lastName'),
+      email: valueFor('email'),
+      universityLogin: valueFor('universityLogin').toLowerCase(),
+      matriculationNumber: valueFor('matriculationNumber'),
+      hasUniversityAccount: true,
+      gender: matchEnumValue(Object.values(Gender), valueFor('gender')),
+      nationality: valueFor('nationality'),
+      studyDegree: matchEnumValue(Object.values(StudyDegree), valueFor('studyDegree')),
```
rappm, line 59: **Unrecognized enum values are dropped and silently replaced by server-side defaults.**

`matchEnumValue` requires an exact case-insensitive match against the wire values and otherwise returns `''`. `StudyDegree` only accepts `bachelor` and `master`, `Gender` only `male|female|diverse|prefer_not_to_say`. Server-side, an empty `StudyDegree` becomes `db.StudyDegreeBachelor` and an empty `Gender` becomes `db.GenderPreferNotToSay`.

Failure case: a roster exported from a university system spells the degree `M.Sc.`, `MSc` or `Master of Science`, or uses German `w`/`m` for gender. Nothing matches, so a whole master's cohort is stored and displayed as bachelor students in the participants table, the study-background diagram and every downstream phase. The preview step shows only aggregate counts, so nothing warns the lecturer.

Suggested direction: have `matchEnumValue` report unmatched non-empty values, surface them in the preview step (count plus the distinct offending values), and let the lecturer either fix the file or confirm the fallback. Silently coercing a value the file did state is different from filling in a value the file omitted.

clients/core/src/managementConsole/applicationAdministration/pages/ApplicationParticipantsPage/components/ImportStudents/utils/parseImportCsv.ts
```diff
@@ -0,0 +1,29 @@
...
+ * numbers and university logins are preserved as strings.
+ */
+export const parseImportCsv = (file: File): Promise<ParsedCsv> =>
+  new Promise<ParsedCsv>((resolve, reject) => {
+    Papa.parse<Record<string, string>>(file, {
+      header: true,
+      skipEmptyLines: true,
+      dynamicTyping: false,
+      transformHeader: (header) => header.trim(),
+      complete: (result) => {
```
rappm, line 21: **`result.errors` is never inspected, so a misparsed row is imported shifted instead of rejected.**

The promise resolves with `result.data` regardless of what papaparse reported. Papaparse assigns cells positionally and pushes overflow into `__parsed_extra`, so `FieldMismatch`, `TooManyFields`, `TooFewFields` and quoting errors leave the row structurally wrong rather than absent.

Failure case: a row contains an unquoted comma in a free-text column, for example `Jane,Doe,ab12cde,jane@tum.de,01234567,female,DE,Informatics, M.Sc.,bachelor,3`. Every column after Study Program shifts by one, so `studyDegree` is read from the Current Semester cell (defaulted to `bachelor` server-side) and `currentSemester` is parsed from a non-numeric cell into `null`. Neither the preview step nor the result summary mentions it.

Suggested direction: reject the file when `result.errors` is non-empty, or at least surface the affected row indices in the upload step before the lecturer can continue. Also worth checking `__parsed_extra` explicitly, since it is the clearest signal that a row does not match the header.

clients/core/src/managementConsole/applicationAdministration/pages/ApplicationParticipantsPage/components/ImportStudents/utils/buildImportRequest.ts
```diff
@@ -0,0 +1,72 @@
...
+export const buildImportRequest = (
+  headers: string[],
+  rows: Record<string, string>[],
+  mapping: Record<string, ColumnTarget>,
+  passStatus: PassStatus,
+): ImportApplicationRequest => {
+  const questionHeaders = headers.filter((header) => mapping[header] === 'question')
+
+  const newQuestions: NewImportQuestion[] = questionHeaders.map((header) => {
+    const maxLength = rows.reduce((max, row) => Math.max(max, (row[header] ?? '').length), 1)
```
rappm: **`allowedLength` is derived from the current file but frozen permanently server-side, which blocks later imports.**

The cap comes from the longest cell in this one file with a floor of 1. `validateApplicationImport` resolves an existing question by title and validates against the persisted `allowedLength`, so the first import decides the limit for every future one.

Failure case: the first CSV has a `Comment` column that is empty in every row, so `rows.reduce(..., 1)` yields 1 and the question is created with `allowed_length = 1`. When the same file is re-imported with the comments filled in, every non-empty comment trips `exceeds the allowed length of 1` and the entire batch is rejected with 400. The lecturer sees only the generic "Import failed" toast (see the `onError` comment) and has no way to discover a hidden cap from a previous import, since nothing in the import UI shows or edits it.

Suggested direction: create imported questions with `MAX_ANSWER_LENGTH` (or another explicit default) rather than a value inferred from sample data. Deriving a permanent constraint from one file's contents produces a limit nobody chose.

servers/core/applicationAdministration/service.go
```diff
@@ -991,6 +995,220 @@ func GetAdditionalScores(ctx context.Context, coursePhaseID uuid.UUID) ([]applic
...
+		cPhaseParticipation, err := coursePhaseParticipation.CreateIfNotExistingPhaseParticipation(ctx, qtx, cParticipation.ID, coursePhaseID)
+		if err != nil {
+			log.Error(err)
+			return applicationDTO.ImportResult{}, errors.New("could not save the course phase participation")
+		}
+		participationIDs = append(participationIDs, cPhaseParticipation.CourseParticipationID)
+
+		for _, ans := range row.Answers {
+			questionID, ok := questionIDByColumn[ans.ColumnKey]
+			if !ok || strings.TrimSpace(ans.Answer) == "" {
```
rappm, line 1226: **A blanked cell cannot clear a previously imported answer.**

Empty answers `continue` here, so the upsert below is never reached for them. Blank means "leave whatever is stored" rather than "clear it".

Failure case: a lecturer spots a wrong `Preferred Team` value, blanks that cell and re-imports. The row is reported as updated and the import succeeds, but the stored answer is untouched, so the application details dialog and the CSV export keep showing the old value indefinitely. There is no other way to clear an imported answer.

Suggested direction: distinguish a column that is absent from the mapping (skip) from a mapped column whose cell is empty (delete the answer row, or upsert an empty value). Ignoring required questions still needs care, so the required-answer check should run against the post-import state rather than the request payload.

clients/core/src/managementConsole/applicationAdministration/pages/ApplicationParticipantsPage/components/ImportStudents/utils/parseImportCsv.ts
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Parses a CSV file in header mode. `dynamicTyping` is disabled so leading zeros in matriculation
+ * numbers and university logins are preserved as strings.
+ */
+export const parseImportCsv = (file: File): Promise<ParsedCsv> =>
+  new Promise<ParsedCsv>((resolve, reject) => {
+    Papa.parse<Record<string, string>>(file, {
+      header: true,
+      skipEmptyLines: true,
```
rappm: **`skipEmptyLines: true` lets a delimiter-only trailing row through, which aborts the whole import.**

In papaparse, `true` only drops a line whose single field is empty; a line of bare delimiters (`,,,,,`) becomes a data row with every value `''`. Excel, Numbers and many university exports end files exactly that way.

Failure case: that row reaches `buildImportRequest` as a student with `universityLogin: ''`, `validateApplicationImport` hits `university login is required for all rows`, and a perfectly valid 500-student file is rejected with 400 while the lecturer sees only the generic "Import failed" toast.

Suggested fix: `skipEmptyLines: 'greedy'`, or filter rows where every mapped value is blank before building the request.

servers/core/applicationAdministration/validation.go
```diff
@@ -275,6 +275,116 @@ func validateApplicationManualAdd(ctx context.Context, coursePhaseID uuid.UUID,
...
+			allowedLengthByColumn[q.ColumnKey] = persisted
+		} else {
+			allowedLengthByColumn[q.ColumnKey] = q.AllowedLength
+		}
+	}
+
+	// Validate the rows.
+	seenLogins := make(map[string]bool, len(req.Rows))
+	for _, row := range req.Rows {
+		login := strings.ToLower(strings.TrimSpace(row.Student.UniversityLogin))
```
rappm, line 365: **In-file duplicate detection covers logins but not emails, so a duplicate address aborts the entire batch.**

`student.email` carries the unique index `student_email_key`, but only `login` is de-duplicated here.

Failure case: a 500-row CSV contains two students who were both given a placeholder address such as `noreply@tum.de`, or a copy-paste duplicate. Validation passes because the logins differ. Inside `PostApplicationImport` the first row inserts, the second row's `CreateStudent` violates the unique index, 23505 is mapped to `ErrEmailAlreadyInUse`, and the transaction rolls back all 500 rows plus the questions created in step 1. The 409 response does not name the duplicated address or the row, and the client replaces even that with the generic toast.

Suggested fix: add an email `seen` set alongside the login one and fail validation with the duplicated address and row index, so the problem is reported before any writes happen rather than as an opaque conflict mid-transaction.

servers/core/applicationAdministration/service.go
```diff
@@ -991,6 +995,220 @@ func GetAdditionalScores(ctx context.Context, coursePhaseID uuid.UUID) ([]applic
...
+
+	questionIDByColumn := make(map[string]uuid.UUID, len(req.NewQuestions))
+	allowedLengthByColumn := make(map[string]int, len(req.NewQuestions))
+	for _, nq := range req.NewQuestions {
+		if existingID, ok := existingIDByTitle[nq.Title]; ok {
+			questionIDByColumn[nq.ColumnKey] = existingID
+			allowedLengthByColumn[nq.ColumnKey] = existingLengthByTitle[nq.Title]
+			continue
+		}
+		questionDBModel := applicationDTO.CreateQuestionText{
```
rappm, line 1118: **This literal omits `AccessibleForOtherPhases` and `AccessKey`, so the insert writes NULLs over the column defaults.**

Both fields stay at `Valid: false`, `GetDBModel()` passes them through, and `CreateApplicationQuestionText` writes explicit `NULL` instead of the migration-0009 defaults `false` and `''`. The non-import path at `service.go:225` persists the client-sent `false`, so imported questions end up in a state no other path can produce.

Failure case: `QuestionText` serializes the pgtype fields directly, so `GET /applications/{id}/form` returns `"accessibleForOtherPhases": null`. On the Questions page `ApplicationQuestionCard` seeds react-hook-form with `defaultValues: { type, ...question }` and validates with `questionConfigSchema`, whose `accessibleForOtherPhases: z.boolean()` rejects `null`. A lecturer who edits an unrelated question and clicks Save is blocked by an "Expected boolean, received null" error on a collapsed Export Settings field of a question they never touched. The `<Switch checked={null}>` in `ExportSettingsFields` also triggers React's controlled/uncontrolled warning.

Suggested fix: set `AccessibleForOtherPhases: pgtype.Bool{Bool: false, Valid: true}` and `AccessKey: pgtype.Text{String: "", Valid: true}` here, matching the normal creation path.

clients/core/src/managementConsole/applicationAdministration/pages/ApplicationParticipantsPage/components/ImportStudents/ImportStudents.tsx
```diff
@@ -0,0 +1,230 @@
...
+        queryKey: ['application_participations', 'students', phaseId],
+      })
+      setResult(importResult)
+      setPage(RESULT_PAGE)
+      toast({
+        title: `Imported ${importResult.created + importResult.updated} students`,
+        variant: 'default',
+      })
+    },
+    onError: () => {
```
rappm: **`onError` takes no argument, so every server diagnostic is discarded.**

The backend returns precise, row-identifying messages: `duplicate university login in file: ab12cde`, `invalid student ab12cde: invalid matriculation number`, `answer for "ab12cde" in column "Team" exceeds the allowed length of 10`, `too many rows in import: 2500`, `university login %q already belongs to a different student`, `email already in use`. All of it is replaced by "Please check the file and try again."

Failure case: a lecturer uploads a 2000-row CSV (the enforced maximum) with one malformed matriculation number. The only feedback is the generic toast, so the file has to be bisected by hand to find the row. Several other findings in this review surface as exactly this toast, which is what makes them hard to diagnose in practice.

Suggested fix: bind the error and render the `utils.ErrorResponse.Error` string in the toast description, the same way the other mutations in this area do. Row-level errors would be even more usable collected per row in the preview step, but showing the server message is the minimum.

servers/core/applicationAdministration/service.go
```diff
@@ -991,6 +995,220 @@ func GetAdditionalScores(ctx context.Context, coursePhaseID uuid.UUID) ([]applic
...
+		studentInput.HasUniversityAccount = true
+		studentInput.UniversityLogin = strings.ToLower(strings.TrimSpace(studentInput.UniversityLogin))
+		studentInput.MatriculationNumber = strings.TrimSpace(studentInput.MatriculationNumber)
+
+		// Detect whether this row updates a student that is already in this course phase, and
+		// preserve any attributes the CSV omits so a partial re-import (or a student known from
+		// another course) does not have their gender/nationality/degree/etc. reset to defaults.
+		outcome := "created"
+		if existing, resolveErr := student.ResolveStudentByUniversityCredentials(ctx, &queries, studentInput.MatriculationNumber, studentInput.UniversityLogin); resolveErr == nil {
+			alreadyInPhase, err := qtx.GetApplicationExistsForStudent(ctx, db.GetApplicationExistsForStudentParams{StudentID: existing.ID, ID: coursePhaseID})
```
rappm: **`GetApplicationExistsForStudent` answers a course-level question, so the created/updated counts are wrong.**

The query is `SELECT EXISTS (SELECT 1 FROM course_participation cp INNER JOIN course_phase ph ON cp.course_id = ph.course_id WHERE cp.student_id = $1 AND ph.id = $2)`. The phase parameter only resolves the course, so this reports "is this student in the course", not "is this student already in this phase". `GetApplicationExists` on `course_phase_participation` is the phase-scoped one.

Failure case: a course has an apply-mode Application phase plus an import-mode one, which is exactly the layout this PR adds to `e2e/seed/e2e_seed.sql` for `fullCourse`. Importing 50 students who already participate in the course creates 50 brand-new `course_phase_participation` rows, but every row is labelled `updated`, so `ImportResultSummary` reports "0 Created / 50 Updated". The lecturer concludes nobody was added and re-runs or escalates. The same wrong `outcome` goes out over the API.

Suggested fix: base the classification on phase participation, ideally on what `CreateIfNotExistingPhaseParticipation` actually did, so the reported outcome cannot drift from the write.

clients/core/src/managementConsole/applicationAdministration/pages/ApplicationParticipantsPage/ApplicationParticipantsPage.tsx
```diff
@@ -19,7 +21,11 @@ export const ApplicationParticipantsPage = (): ReactNode => {
           {participations && customScoresEnabled && (
             <AssessmentScoreUpload applications={participations} />
           )}
-          <ApplicationManualAddingDialog existingApplications={participations ?? []} />
+          {importModeEnabled ? (
```
rappm: **Making the two paths mutually exclusive removes the only way to add a student without a TUM login.**

`ApplicationManualAddingDialog` can add a student with `hasUniversityAccount: false` and no university login, via `UniversitySelection`. The import path cannot: `validateApplicationImport` rejects any row with an empty login, and `student.ValidateImport` calls `validateUniversityData(true, ...)`, whose regex `^[a-zA-Z]{2}\d{2}[a-zA-Z]{3}$` rejects anything that is not a TUM ID.

Failure case: a lecturer running an import-mode phase needs to admit one exchange or external student without a TUM ID. The manual dialog is gone from the page and the import endpoint refuses the row, so the only workaround is switching the phase back to apply mode, which re-opens the public form and changes the phase status badge, just to add one person.

Suggested fix: keep both actions available in import mode, with import as the primary. The two are complementary rather than alternatives.

servers/core/applicationAdministration/validation.go
```diff
@@ -275,6 +275,116 @@ func validateApplicationManualAdd(ctx context.Context, coursePhaseID uuid.UUID,
...
 
+// maxImportRows bounds a single CSV import. Each row runs several queries inside one transaction,
+// so an unbounded file would hold a long-running transaction open.
+const maxImportRows = 2000
+
+func validateApplicationImport(ctx context.Context, coursePhaseID uuid.UUID, req applicationDTO.ImportApplicationRequest) error {
+	ctxWithTimeout, cancel := db.GetTimeoutContext(ctx)
+	defer cancel()
+
+	if len(req.Rows) > maxImportRows {
```
rappm, line 289: **Row count is bounded from above but not from below, and questions are created before any row is read.**

`PostApplicationImport` creates every `NewQuestions` entry in step 1 (`service.go:1047`) and only then loops over the rows, so a request with zero rows still commits the new questions.

Failure case: a lecturer exports a filtered spreadsheet that ends up with headers and no data rows, or deletes the example row from the downloaded template after adding columns. `parseImportCsv` resolves with `headers.length > 0, rows: []`; `goNext` only rejects `headers.length === 0`; `validateMapping` passes because the four required targets are mapped; the preview shows "0 students". Clicking Import commits, and the phase silently gains N text questions with no answers, which the import dialog offers no way to remove. The lecturer sees a success toast.

Suggested fix: reject `len(req.Rows) == 0` here, and have the client block the Import button when there are no rows.

rappm (changes requested):
## Changes requested

I reviewed the import path end to end (client wizard, `PostApplicationImport`, validation, the SQL and role-lookup changes) and left 15 inline comments. The feature works on a clean, well-formed file, but the paths around that happy case need work before this ships. Grouping them by why they matter:

**1. Blocking: security and data loss**

- `permission_strings.sql:24` plus the `middleware.go` change let a Keycloak account receive a course's `-Student` role on a matching university login alone, with no matriculation number. A single mistyped University ID in a CSV grants an unrelated student access to the course. The relaxation applies to every login, not just to import-mode phases, so this is the one item I would treat as a hard blocker.
- `service.go:1194` applies the batch pass status to pre-existing participations too, so re-importing a roster silently resets manually set statuses. `failed` is not offered in the dialog, so it cannot be undone through the import flow.
- `student/validation.go:47` (`ValidateImport`) skips the semester, nationality, study program and study degree checks that `Validate` enforces, while the import writes to the shared `student` row. A `Semester` cell of `0` follows the student into every other course in PROMPT.

**2. Silent data corruption in the CSV pipeline**

Each of these accepts a bad file and produces wrong data rather than an error:

- `parseImportCsv.ts:19` ignores `result.errors`, so an unquoted comma imports a row with every column shifted by one.
- `parseImportCsv.ts:16` uses `skipEmptyLines: true` instead of `'greedy'`, so a trailing `,,,,,` line from Excel rejects the whole file.
- `buildImportRequest.ts:59` maps any unrecognized gender or degree to `''`, and the server then defaults it, so `M.Sc.` silently becomes `bachelor`.
- `service.go:1078` lowercases the university login while every lookup and the unique index are case-sensitive, so a student stored as `GE42ABC` is either duplicated or aborts the batch.

**3. Self-inflicted feature blockers**

- `buildImportRequest.ts:30` freezes each question's `allowedLength` to the longest cell of the first file, so a later import of longer values is rejected by a limit nobody chose and no UI exposes.
- `service.go:1150` skips blank cells, so an imported answer can never be cleared.
- `validation.go:359` de-duplicates logins but not emails, so one duplicated address rolls back the whole batch with an opaque 409.
- `validation.go:286` allows a zero-row import, which still creates the new questions and reports success.
- `service.go:1053` writes NULL for `accessible_for_other_phases` and `access_key`, which breaks saving on the Questions page for questions the lecturer never touched.

**4. Reporting and UX**

- `ImportStudents.tsx:96` discards the server's error message. The backend already returns precise, row-identifying errors, and the UI replaces all of them with "Please check the file and try again." This is what turns most of the items above into unactionable failures on a 2000-row file, so fixing it has leverage well beyond its own severity.
- `service.go:1086` classifies created versus updated with a course-level existence query, so a course with two Application phases reports "0 Created / N Updated" for rows that were genuinely new.
- `ApplicationParticipantsPage.tsx:24` replaces the manual-add dialog rather than complementing it, so an import-mode phase cannot admit a student without a TUM login at all.

Happy to talk through any of these, particularly the role-lookup change, where the right fix depends on whether imports really have to work without a matriculation number.

mathildeshagl (dismissed):
All seven points addressed, and I verified each one against \`fc26611c\` / \`7dd58e49\` / \`3d747c40\`:

1. \`GetStudentRoleStrings\` now reads \`COALESCE(s.matriculation_number,'') = '' OR s.matriculation_number = $1 OR $1 = ''\`, which is the disjunct the empty-string write from \`CreateStudent\` actually needs, and \`TestImport_StudentRoleResolvedWithoutMatriculation\` now asserts with a non-empty \`$1\` — the production login path that was broken.
2. The import route is \`PromptAdmin\` + \`CourseLecturer\`, matching the sibling bulk-admit endpoints.
3. Existing gender / nationality / studyProgram / studyDegree / currentSemester are preserved when the CSV omits them, with defaults applied only to genuinely new students, and \`TestImportApplications_ReImportPreservesOmittedAttributes\` covers it.
4. The length check moved into \`validateApplicationImport\` (400 pre-transaction), and the in-transaction path wraps \`ErrImportAnswerTooLong\` so the handler maps it to 400 too. The test asserts both.
5. \`outcome\` is derived from \`GetApplicationExistsForStudent\`, so it now means the same thing as the client preview.
6. \`maxImportRows = 2000\` bounds the transaction.
7. \`papaparse\` and \`@types/papaparse\` are declared in \`clients/core/package.json\`, and the \`ParseResult\` generic in \`useUploadAndParseCSV\` is properly typed rather than latent \`any\`.

The three items you consciously kept are all reasonable as stated, and the seed PK collision catch on \`d0000010\` was a good save. CI is green.

servers/core/db/query/permission_strings.sql
Mtze: Fixed in ef0499a5. Tightened the predicate to `s.matriculation_number = $1 OR COALESCE(s.matriculation_number,'') = ''` (dropped the `OR $1 = ''` branch). Because `university_login` is unique (partial unique index), it identifies the student on its own; a token that carries no matriculation claim now matches only a student whose stored matriculation is also empty, never one carrying a different number. Documented the reasoning in a query comment and added a regression test (`TestImport_RoleNotResolvedForStoredMatriculation`).

servers/core/applicationAdministration/service.go
Mtze: Fixed in ef0499a5. The batch pass status is now applied only to participations the import newly creates (`createdParticipationIDs`); pre-existing participations keep their current status, so a re-import can no longer flip a hand-set `failed`. Added `TestImportApplications_ReImportKeepsManualPassStatus` and noted the behavior in the docs.

servers/core/student/validation.go
Mtze: Fixed in ef0499a5. `ValidateImport` now rejects an out-of-range semester (`< 1`), a study degree outside the enum, and an invalid gender **when the CSV provides one**, so a bad value can no longer be written to the shared student row. Empty cells stay lenient (defaulted by the service). Nationality is still accepted as-is for now (the client sends free text, not an ISO code); I left ISO enforcement out of this pass rather than half-implement it. Added `TestImportApplications_InvalidOptionalFieldsRejected`.

servers/core/applicationAdministration/service.go
Mtze: Kept as-is, with reasoning. The inconsistency is real, but the correct fix is systemic: a `lower(university_login)` unique index plus case-insensitive lookups applied to the manual path too. Removing the normalization here is the riskier option, since the Keycloak login claim is lowercase and `GetStudentRoleStrings` is case-sensitive, so storing a mixed-case login from a CSV would break the very post-import login this feature exists for. I'd rather track the full case-insensitive normalization as a separate change than introduce a second, subtly different behavior here.

clients/core/src/managementConsole/applicationAdministration/pages/ApplicationParticipantsPage/components/ImportStudents/utils/buildImportRequest.ts
Mtze: Fixed in ef0499a5. Added `collectUnmatchedEnumValues`, and the preview step now shows a warning listing the distinct unrecognized gender / study-degree values and the default they will fall back to, instead of coercing silently.

clients/core/src/managementConsole/applicationAdministration/pages/ApplicationParticipantsPage/components/ImportStudents/utils/parseImportCsv.ts
Mtze: Fixed in ef0499a5. `parseImportCsv` now rejects the file when papaparse reports a `FieldMismatch`/`Quotes` error or when any row carries `__parsed_extra`, naming the offending row, so a shifted row fails loudly instead of importing with the wrong columns.

clients/core/src/managementConsole/applicationAdministration/pages/ApplicationParticipantsPage/components/ImportStudents/utils/buildImportRequest.ts
Mtze: Fixed in ef0499a5. Imported questions are now created with a fixed default length (`MAX_ANSWER_LENGTH`) rather than one inferred from the sample file, so an empty column can no longer freeze `allowed_length` at 1 and block later imports.

servers/core/applicationAdministration/service.go
Mtze: Kept for v1, documented. This matches the locked plan ("empty cells produce no answer row"): the client filters out empty answers and the server skips them. Supporting clear-on-blank needs a client+server change to send and act on mapped-but-empty cells, which I'd rather do deliberately. I documented the limitation in `docs/user/application.md` (blank cell keeps the previous value; edit it on the Questions/participant view instead).

clients/core/src/managementConsole/applicationAdministration/pages/ApplicationParticipantsPage/components/ImportStudents/utils/parseImportCsv.ts
Mtze: Fixed in ef0499a5. Switched to `skipEmptyLines: 'greedy'`, so a delimiter-only trailing row (`,,,,`) is dropped instead of becoming a blank student row that aborts the import.

servers/core/applicationAdministration/validation.go
Mtze: Fixed in ef0499a5. `validateApplicationImport` now also de-duplicates emails in the file (case-insensitively) and fails early with the duplicated address, before any writes, instead of the second row aborting the transaction with an opaque 23505. Added `TestImportApplications_DuplicateEmailRejected`.

servers/core/applicationAdministration/service.go
Mtze: Fixed in ef0499a5. The imported-question literal now sets `AccessibleForOtherPhases: pgtype.Bool{Bool:false, Valid:true}` and `AccessKey: pgtype.Text{String:"", Valid:true}`, matching the non-import path, so the insert writes the column defaults instead of NULL. Added `TestImportApplications_QuestionDefaultsPersisted`.

clients/core/src/managementConsole/applicationAdministration/pages/ApplicationParticipantsPage/components/ImportStudents/ImportStudents.tsx
Mtze: Fixed in ef0499a5. `onError` now binds the error and renders `response.data.error` in the toast description (falling back to the generic message), the same way the sibling mutations in this area do.

servers/core/applicationAdministration/service.go
Mtze: Fixed in ef0499a5. Classification now uses the phase-scoped `GetApplicationExists` (on `course_phase_participation`, checked before the row is created) instead of the course-level `GetApplicationExistsForStudent`, so created/updated counts are correct even when a course has both an apply- and an import-mode phase. This also drives the pass-status scoping in the related thread.

clients/core/src/managementConsole/applicationAdministration/pages/ApplicationParticipantsPage/ApplicationParticipantsPage.tsx
Mtze: Fixed in ef0499a5. The manual **Add Student** dialog now stays available in import mode (with import as the primary action), so a student without a TUM login can still be added. The manual-add route is not mode-gated on the server, so this needed only the page change. Note the mutual exclusivity was only ever about the public form, not the lecturer dialog.

servers/core/applicationAdministration/validation.go
Mtze: Fixed in ef0499a5. `validateApplicationImport` now rejects `len(req.Rows) == 0` before any questions are created, and the client disables the Import button when there are no rows. Added `TestImportApplications_EmptyRowsRejected`.

Mtze:
@rappm thanks for the thorough review - genuinely useful. Addressed in `ef0499a5` (with the merge in `d3f8d4b8`).

**Fixed (13):**
- **Role query scope** - dropped the `OR $1 = ''` branch so a token without a matriculation claim resolves only a student whose stored matriculation is also empty; login is unique so it identifies on its own. Regression test added.
- **Pass status** - applied only to newly-created participations; re-import no longer overwrites hand-set statuses.
- **Created/updated counts** - now phase-scoped (`GetApplicationExists`), correct with mixed apply/import phases in one course.
- **Optional-field validation** - semester/degree/gender validated when the CSV provides a value.
- **Question defaults** - `AccessibleForOtherPhases`/`AccessKey` written as defaults, not NULL.
- **CSV parsing** - reject on `FieldMismatch`/`__parsed_extra`; `skipEmptyLines: 'greedy'`.
- **Question length** - fixed default instead of file-inferred (no more `allowed_length = 1` lock-in).
- **In-file email dedup** - fails early with the address.
- **Zero-row import** - rejected server-side + Import button disabled.
- **Error surfacing** - server message shown in the toast; unrecognized enum values warned in the preview.
- **Manual add in import mode** - dialog kept so no-TUM-login students can still be added.

**Kept intentionally (2), with reasoning on the threads:**
- **Login case normalization** - the proper fix is systemic (`lower()` index + case-insensitive lookups on the manual path too); removing normalization here would risk breaking post-import login (Keycloak claim is lowercase). Tracking separately.
- **Clear-on-blank answers** - matches the locked plan; documented as a limitation for now rather than adding the client+server change ad hoc.

New Go tests cover the role scope, manual-status preservation, question defaults, email dedup, empty rows, and invalid optional fields. `make lint`/tests green locally. Ready for another look.

rappm (approved):

Issue #1822, opened 2026-07-02
E2E: certificate phase module coverage


rappm:
## Context

The certificate service (`servers/certificate`) and client (`clients/certificate_component`) have no e2e coverage and are not part of the e2e stack.

## Scope

Following the pattern from #1811:

- [ ] Add the certificate server (own Postgres DB, seed) and client to `docker-compose.e2e.yml`
- [ ] Extend the seed so the seeded course has a certificate phase with participants
- [ ] Lecturer journey: configure and trigger certificate generation
- [ ] Student journey: download the generated certificate
- [ ] API auth check: certificate endpoints reject unauthenticated and wrong-role requests

Depends on #1810 and #1811.

## Definition of done

- Module boots in `make test-e2e`, journeys and auth checks are green

Issue #1816, opened 2026-07-02
E2E: interview phase module coverage


rappm:
## Context

The interview service (`servers/interview`) and client (`clients/interview_component`) have no e2e coverage and are not part of the e2e stack.

## Scope

Following the pattern from #1811:

- [ ] Add the interview server (own Postgres DB, seed) and client to `docker-compose.e2e.yml`
- [ ] Extend the seed so the seeded course has an interview phase with participants
- [ ] Lecturer journey: create interview slots, see the schedule fill up
- [ ] Student journey: self-assign to a slot
- [ ] API auth check: interview endpoints reject unauthenticated and wrong-role requests

Depends on #1810 and #1811.

## Definition of done

- Module boots in `make test-e2e`, journeys and auth checks are green

PR #1968, opened 2026-07-21
`Testing`: Shard e2e CI per microservice with build-once layer caching

### LINKED ISSUE #1824: E2E: CI scaling with parallel workers and sharding
## Context

The e2e suite runs with `workers: 1` and `fullyParallel: false` (`e2e/playwright.config.ts`). Fine for the current 5 spec files, but a wall-clock problem once core write paths and the per-module suites land.

## Scope

- [ ] Enable parallel workers where safe; keep the per-file serial guarantee (login flows mutate the shared Keycloak session)
- [ ] Shard the suite in CI (e.g. per module) with merged reporting
- [ ] Set a runtime budget and fail or alert when exceeded
- [ ] Verify seeded-data isolation holds under parallelism (tests that mutate data must clean up or use dedicated fixtures)

Should land after the bulk of the module suites (#1816 to #1822) exist, since sharding boundaries depend on them.

## Definition of done

- CI e2e wall-clock time stays within the agreed budget with all suites enabled

magkue:
## Summary

Scales the e2e suite for wall-clock by sharding it across a GitHub Actions
matrix — **one shard per microservice** plus a `core` shard for the core-hosted
tests — with a build-once/fan-out Docker layer cache and merged reporting.
Closes #1824.

## Details

- **Sharding (`.github/workflows/test-e2e.yml`)**: a `build` job builds every
  image once and populates the gha layer cache; a matrix of module shards each
  restores from that cache, boots its **own fresh stack** (own DB seed, so
  modules can't collide on shared data), and runs only its module's specs; a
  `merge` job stitches the per-shard blob reports into one `playwright-report`.
- **Caching overrides** (`docker-compose.e2e.cache.yml` read-only shared by all
  jobs, `docker-compose.e2e.cache-write.yml` added only by the build job so the
  8 shards don't re-export concurrently). CI-only — local `make test-e2e` never
  passes them, so it is unchanged.
- **1 worker per shard** — parallelism comes from the matrix, not from within a
  shard, so no per-spec isolation audit is needed now (the specs are already
  written for it, per `e2e/README.md`). `PW_WORKERS` can raise it later.
- **Runtime budget ~20 min/shard**, enforced by the job `timeout-minutes` and
  Playwright's `globalTimeout` (`PW_GLOBAL_TIMEOUT_MIN`).
- **Config** (`e2e/playwright.config.ts`): env-driven `workers`, `globalTimeout`
  budget, and a per-shard blob reporter (`PW_BLOB` + `PW_SHARD_NAME`).
- **Local helper**: `make test-e2e-shard PATHS="…"` reproduces one CI shard.

The shard partition was verified to cover all 51 spec files exactly once
(7 microservice shards + core).

## Reason / Link to issue

The suite ran with `workers: 1` in a single container — fine at 5 specs, a
wall-clock problem now that all module suites (#1816–#1822) have landed
(51 specs). Closes #1824.

## How to Test

1. **Locally**, reproduce one shard: `make test-e2e-shard PATHS="tests/interview tests/api/interview.api.spec.ts"`.
2. **In CI**, the `build → e2e (matrix) → merge` jobs run; confirm each shard
   restores images from the layer cache (fast rebuild) and the merged
   `playwright-report` artifact contains every shard's results.

> ⚠️ **Needs a CI run to validate the caching path.** The Docker layer-cache
> restore (`docker compose build` under a buildx container builder loading images
> for `compose run`) and the blob-report merge can only be exercised on a real
> runner. Config, YAML, compose merge, and the module→shard partition were all
> validated locally.

## PR Checklist

## Summary by CodeRabbit

* **Testing**
  * End-to-end tests now run in validated, parallel Playwright shards.
  * Each shard uses an isolated environment with shared Docker build caching.
  * Shards enforce time limits and generate separate reports, which are merged into one HTML report.
  * Failure artifacts are preserved for easier troubleshooting.
  * Added commands for running individual E2E shards locally.

* **Documentation**
  * Updated E2E guidance covering sharding, caching, timeouts, reporting, and local execution.


mathildeshagl (changes requested):
The sharding itself is sound and the partition is correct. The problem is that the caching layer the design is built around measurably does nothing, and as a result CI is currently slower than before this PR.

**1. `docker-compose.e2e.cache.yml` / `cache-write.yml` + `.github/workflows/test-e2e.yml:29,85` — the gha layer cache never engages.** From this PR's own run (30549620776): the build job's build step takes 6m45, and every shard's "restore from the gha layer cache" step takes 6m48-6m55 — no faster than a cold build. Shard logs contain zero `CACHED` or `importing cache manifest` lines and show the full yarn/webpack client build re-running; the build job log has no cache-export line at all. The cause is that `type=gha` needs `ACTIONS_RUNTIME_TOKEN` / `ACTIONS_RESULTS_URL`, which GitHub injects only into JS actions, not into `run:` shell steps, so `docker compose build` cannot authenticate — and `ignore-error=true` swallows the export failure silently. Adding `crazy-max/ghaction-github-runtime@v3` before both build steps (or building via `docker/bake-action`) should fix it.

**2. Net wall-clock regression.** Because of (1), `needs: build` prepends roughly 7.5 min to shards that rebuild from scratch anyway: about 17.5 min total observed, versus about 10 min if the `build` job were simply deleted, plus roughly 56 redundant runner-minutes. So this is a clean either/or: fix (1) and the cache pays for itself, or drop the build job and both override files and take the ~10 min.

**3. `test-e2e.yml:120-122`** — `merge` has `if: always()`, so when `build` fails and the shards are skipped it still runs, and `download-artifact` with `pattern:` errors on zero matches. One failure becomes two red jobs. `always() && needs.e2e.result != 'skipped'` avoids it.

**4. `test-e2e.yml:43`** — the job timeout dropped from 40 to 25 min, but 25 min must now cover a ~7 min build plus boot plus tests, while `PW_GLOBAL_TIMEOUT_MIN=20` bounds only the Playwright phase (currently 1.5-3 min). The "Playwright fails first with a clean message" intent does not hold. The comment at :39-42 also states the same thing twice.

**5. `test-e2e.yml:48-67`, `Makefile:180`** — the Playwright path arguments are regexes, not paths. The partition is exact today (I verified all 51 spec files match exactly once, with no overlap), but a future `tests/foo-v2/` would silently join the wrong shard, and an unmatched new directory would run nowhere with no error. A partition-coverage assertion would make this durable.

**6. `Makefile:180`** — `make test-e2e-shard` with no `PATHS` silently runs the whole suite.

Nits: `${{ matrix.paths }}` is inlined into `run:` (static and trusted, but `env:` is the safer habit); `e2e/package.json:12` adds a `merge-reports` script that CI does not use while `README:356` claims it does; the flow-style `build: { ... }` clashes with the surrounding block style.

The good parts are real: a per-shard fresh stack with its own seed is the right isolation model, `fail-fast: false` is right, the new `if: failure()` trace/video upload is a genuine debuggability gain, blob-then-merge is idiomatic Playwright, local `make test-e2e` is deliberately untouched, and README/.gitignore/Makefile were updated together.

mathildeshagl (dismissed):
All six points addressed, and the either/or from (2) is resolved in the direction I would have picked.

1. `crazy-max/ghaction-github-runtime@v3` now runs before the build step, so `type=gha` can actually resolve `ACTIONS_RUNTIME_TOKEN` / `ACTIONS_RESULTS_URL` from a `run:` step. Dropping `ignore-error=true` on the write path is the right call too — a failing export should be loud rather than quietly leaving every later run cold.
2. The build-once job is gone. On the run for `7992dac4` the shards span 11:57:52 to 12:09:33, so about 11.5 minutes end to end against the roughly 17.5 I measured before. The remaining cost is the cold build, which is expected on the first run: this run is the one that populates the scopes, so the cross-run payoff is not yet visible in these numbers. Worth a glance at the next run's build-step timings to confirm the restore actually engages.
3. `always() && needs.e2e.result != 'skipped' && needs.e2e.result != 'cancelled'` keeps `download-artifact` from erroring on an empty pattern.
4. Job timeout is 40 with the two budgets now described distinctly, so Playwright's `globalTimeout` is genuinely the one that fires first, and the duplicated sentence is gone.
5. `shards.json` plus `scripts/shards.mjs check` is a better answer than the assertion I asked for: patterns are anchored on both ends, the check asserts every spec is claimed exactly once *and* that no pattern has gone stale, the matrix is generated from the same file so the two cannot drift, and the `partition` job runs it before any runner is spent. Mirroring Playwright's `forceRegExp` rather than approximating it is the detail that makes this trustworthy.
6. `make test-e2e-shard` with no `SHARD` or `PATHS` now exits 2 and lists the available shards.

The nits are handled as well: `matrix.paths` moved to `env:` with `set -f` so the regex metacharacters survive, CI uses the `merge-reports` script the README claims it does, and the cache overrides are back to block style.

One note before merge: `e2e-tests / core` is red, but it failed in `docker/setup-buildx-action@v3` after 19 seconds with every later step skipped, which is a runner-side flake and not something this PR introduced. It needs a re-run.

rappm (approved):

PR #1969, opened 2026-07-21
`Application`: Show exported question answers in participants table

### LINKED ISSUE #811: Show exported application questions in participants page
Use the shared library participations table with the extraColumns functionality

rappm:
# 📝 Pull Request Template

## ✨ What is the change?

The application participants table now shows the answers to **exported application questions** (questions marked "accessible for other phases" with an access key) as additional columns, one per exported question.

- New backend endpoint `GET /applications/:coursePhaseID/exported-answers` returns, for a course phase, the list of exported-question columns (`questionID`, `key`, `title`, `orderNum`, `type`) plus each participant's answers keyed by `questionID`. Multi-select answers are joined with `, `.
- The endpoint reuses the existing question/answer queries (no migration, no `sqlc` regeneration) and is fully isolated from the participations endpoint.
- The frontend loads the answers via its own query. The columns are toggleable via the existing column-visibility control and are keyed by `questionID` (so duplicate access keys stay separate columns and never clobber base columns). If the answers fail to load, the base table still renders and a toast is shown.

Lecturers want to see the answers to exported application questions directly in the participants overview instead of opening each application individually.

Closes #811

## 🧪 How to Test

1. Open an Application phase that has at least one question with **"accessible for other phases"** enabled and an access key set (text and/or multi-select).
2. Add some applications and answer those questions.
3. Go to the Application **Participants** page.
4. Verify a new column appears for each exported question (header = question title), populated with each participant's answer (multi-select answers comma-separated, missing answers shown as `-`).
5. Toggle the new columns via the column-visibility control.

| before | after |

## ✅ PR Checklist


mathildeshagl (commented):
Correct and well-scoped. My comments are about efficiency and test durability rather than correctness, so nothing here needs to block you.

**1. `servers/core/applicationAdministration/service.go:~910` — over-fetching.** This pulls *every* text and multi-select answer for every participant and discards the non-exported ones in Go. `GetAllApplicationAnswersTextByCourseParticipationIDs` is filtered only by `course_participation_id`, so it also pulls answers from other phases the student applied to. On a large phase that is megabytes of wasted transfer per page load. `servers/core/db/query/course_phase_participation.sql:310-352` already contains the exact exported-answer SQL (`accessible_for_other_phases = true AND access_key IS NOT NULL AND access_key <> ''`); a dedicated sqlc query would be faster and would reuse the existing semantics.

**2. `service.go:~841` (`isExportedQuestion`)** — reimplements that export predicate with `strings.TrimSpace(...) != ""` instead of the SQL `<> ''`. Two sources of truth for "is this question exported" can drift; today they agree only because `validateExportSettings` rejects whitespace in keys.

**3. `service.go:~896`** — calls the heavy `GetAllApplicationParticipations` (joins student + assessment + score) purely to obtain the participation-ID list, then emits an `answers` entry for every participant even when empty. The client's `?? '-'` already handles absent entries, so an ID-only query plus omitting empty entries would shrink both query cost and payload.

**4. `servers/core/applicationAdministration/service_test.go:~546`** — `assert.Equal(4, len(response.Columns))` holds only because the seed dump happens to have no exported questions in phase `4179d58a...`. Any future seed change silently breaks it; asserting on the six fixture IDs specifically would be robust. The test also inserts fixtures via raw SQL while the rest of the suite seeds from `database_dumps/`, and uses `interface{}` rather than `any`.

**5. `ApplicationParticipantsTable.tsx:~61`** — inline `useQuery` without the explicit generic used elsewhere; the module's convention for participation data is a hook (`applicationAdministration/hooks/useGetApplicationParticipations.ts`), so a `useGetExportedApplicationAnswers` hook would match. Also nothing invalidates `['application_exported_answers', phaseId]` after question-config edits, though the default staleTime of 0 mitigates it.

**6. `applicationDTO/exported_application_answers.go:7-11`** — `key` and `type` are returned but unused by the client.

Things I liked: `EXPORTED_ANSWER_COLUMN_PREFIX` keyed by `questionID` cleanly prevents collisions with base and additional-score columns as well as duplicate access keys; the query failure degrades gracefully with a toast; the roles on the new route match the existing participations route; swagger is regenerated; no `any` added; CSV export is correctly unaffected thanks to the explicit `CsvColumn` list; and the test covers duplicates, whitespace keys, and non-exported questions. The gofmt fixes to pre-existing `service.go` code are a welcome cleanup.

mathildeshagl (approved):
The three efficiency points are addressed, and better than I suggested.

1. `GetExportedApplicationQuestionsForCoursePhase` and `GetExportedApplicationAnswersForCoursePhase` scope the fetch to this phase's exported questions in SQL, so answers from other phases are never transferred and discarded in Go, and the multi-select join arrives pre-flattened.
2. `isExportedQuestion` is gone entirely — the predicate now lives only in SQL, matching the `applicationAnswers` projection in `course_phase_participation.sql`. Aligning on `<> ''` rather than `TrimSpace` is the right call, and the test now pins that a whitespace-only key is kept, with the reason in a comment.
3. The heavy `GetAllApplicationParticipations` call is gone and only participations with at least one answer are emitted. I checked `applicationColumns.ts:47` — the `?? '-'` fallback covers the omitted ones.
4. The column assertion now pins the expected question IDs in order rather than only a count.

Points 5 and 6 (the inline `useQuery` versus a `useGetExportedApplicationAnswers` hook, and the unused `key` / `type` in the DTO) are unchanged, but those were the non-blocking ones and neither is worth another round.

CI is green.

PR #1970, opened 2026-07-21
`Core`: Add course search to the course overview

### LINKED ISSUE #932: Search for Courses in the Course Overview

rappm:
## Summary

Adds a search field to the course overview so users can quickly find a course by name or semester.

## Details

- The course cards view (`CourseCards`) now renders a search input that filters the courses by name or semester tag as you type (case-insensitive).
- The cards view is shared by the active, archived, and template course overviews, so all three gain search from this single change.
- When the query matches nothing, a "No courses match your search." message is shown.
- The table view already provides built-in search via `PromptTable`, so it is left unchanged (no duplicate search box).

## Reason / Link to issue

The course overview had no way to search across courses in the cards view. Closes #932.

## How to Test

1. Open the management console course overview (`Courses`) in cards view.
2. Type part of a course name or semester tag into the new search field.
3. Verify the visible cards filter down to matching courses, and that clearing the field shows all courses again.
4. Repeat on the Archived and Template course pages (cards view).

## PR Checklist

## Summary by CodeRabbit

* **New Features**
  * Added client-side course search that filters results by course name or semester.
  * Added an animated empty state when no courses match the search.
  * Improved the course list presentation by animating updates to the filtered results.


JGStyle:
Oh, this isn't on schau mi o yet

JGStyle:
This made me also realize that our current courses page isn't full width, as the others are:

2026-07-29 at 19 29 45

Minor, because only really noticable at high window size, but I noted it here #2011

rappm:
good catch, now it is on schau-mi-o

mathildeshagl (commented):
Nice, well-scoped change — one file gives all three overview pages search, and it correctly avoids duplicating `PromptTable`'s built-in search in table view. One thing I would fix before merge, plus some polish.

**1. `clients/core/src/managementConsole/shared/components/CourseCard/CourseCards.tsx:36` — empty state fires without a query.** The message renders whenever `filteredCourses.length === 0`, not only when a search is active. `TemplateCoursesPage.tsx:21-25` renders `<CourseCards>` unconditionally alongside its own "No template courses yet", so a user with zero templates now sees a search box plus a misleading "No courses match your search." with an empty query. `ActiveCoursesPage` and `ArchivedCoursesPage` are unaffected because they guard with `isEmpty`. Gating the message (and arguably the search bar itself) on `search.trim().length > 0` would do it, or drop the now-redundant paragraph in `TemplateCoursesPage`.

**2. `CourseCards.tsx:24`** — the `Search` icon lacks `pointer-events-none`, so it swallows clicks on the input's left padding instead of focusing it. Both existing search inputs in the repo have it (`courseUserManagement/components/AddUserDialog.tsx:82`, `team_allocation_component/.../TutorSearchStudents.tsx:99`); matching them also fixes the class-order inconsistency.

**3. `CourseCards.tsx:25-31`** — no `aria-label` on the input, so the placeholder is the only accessible name. Cheap to add and consistent with the a11y rules enabled in `clients/biome.json`.

**4. `CourseCards.tsx:18`** — `field?.toLowerCase()`: both `name` and `semesterTag` are non-optional `string` on `Course`, so the optional chaining is dead defensive code.

**5. Design question, not a blocker.** Search state lives inside `CourseCards`, so it resets on every cards/table toggle and is visually indented relative to the page `h1` (which sits outside the `container mx-auto px-4 py-8` wrapper). Hoisting the query up to the page, next to `CourseViewToggle`, would keep it aligned and persistent across view modes. Happy either way — your call.

No tests, which seems fine here: `clients/` has no unit-test infra and no Playwright spec selects courses on `/management/courses` by text, so nothing breaks. An e2e assertion on the filter would be a nice-to-have.

`useMemo` keyed on `[courses, search]`, case-insensitive and trimmed, is the right shape — no debounce needed for a client-side array — and reusing `Input` from `@tumaet/prompt-ui-components` instead of hand-rolling is exactly right.

PR #1971, opened 2026-07-21
`Application`: Suggest a default access key when exporting a question

### LINKED ISSUE #174: Default Export Key
When exporting application questions, suggest a default name for the exported question

rappm:
## Summary

When configuring an application question, enabling "Accessible for Other Phases" now suggests a default access key instead of leaving the field empty.

## Details

- In `ExportSettingsFields`, toggling "Accessible for Other Phases" on now prefills the Access Key with a suggestion derived from the question title (lowercased, runs of non-alphanumeric characters collapsed to `_`, trimmed, capped at 50 chars to match the column limit).
- The suggestion is only applied when the access key is currently empty, so an existing/edited key is never overwritten. The field stays fully editable.
- Turning the toggle off still clears the access key (unchanged behavior).

## Reason / Link to issue

Previously the access key started empty, forcing lecturers to invent a key by hand. Closes #174.

## How to Test

1. Open an Application phase → question configuration and add or edit a question with a title (e.g. "Motivation Letter").
2. Toggle "Accessible for Other Phases" on.
3. Verify the Access Key field is prefilled with a valid suggestion (e.g. `motivation_letter`) and is still editable.
4. Clear the key, toggle off and on again to confirm it re-suggests; set a custom key, toggle off/on, and confirm a non-empty key is not overwritten.

## PR Checklist

### DIFF
--- a/clients/core/src/managementConsole/applicationAdministration/pages/ApplicationQuestionConfigPage/FormPages/FormFields/ExportSettingsFields.tsx
+++ b/clients/core/src/managementConsole/applicationAdministration/pages/ApplicationQuestionConfigPage/FormPages/FormFields/ExportSettingsFields.tsx
@@ -18,6 +18,14 @@ interface ExportSettingsFieldsProps {
   onCsvExportEnabledChange: (checked: boolean) => void
 }
 
+const suggestAccessKey = (title: string): string =>
+  title
+    .trim()
+    .toLowerCase()
+    .replace(/[^a-z0-9]+/g, '_')
+    .replace(/^_+|_+$/g, '')
+    .slice(0, 50)
+
 export const ExportSettingsFields = ({
   form,
   csvExportEnabled,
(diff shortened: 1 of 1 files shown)


magkue (commented):
Clean, well-scoped change, and it works as intended locally (toggling "Accessible for Other Phases" on with a title like `Motivation Letter` prefills `motivation_letter`, and the field stays editable). The `suggestAccessKey` helper only ever emits `[a-z0-9_]`, so the suggestion always satisfies the access-key validation, and the empty-only guard correctly avoids clobbering an existing key.

Two small, non-blocking notes:

1. **Uniqueness.** The suggested key isn't checked for uniqueness against other questions, and there's no validation that flags a duplicate key either — so if two questions end up with the same key the user wouldn't directly notice the collision. In practice this requires giving two questions the same title, which is atypical, so I'd treat this as a nice-to-have rather than a blocker.

2. **Doc nit in "How to Test" step 4.** The step says a custom key isn't overwritten on toggle off/on, but since toggling off clears the key (`form.setValue('accessKey', '')`), toggling back on finds an empty key and re-suggests, replacing the custom value. The "don't overwrite" guard only preserves a pre-existing key when the toggle starts in the off state. Worth aligning either the description or the behavior.

Issue #1724, opened 2026-06-24
Apply ↔ Management: session is lost and re-login is forced


magkue:
Navigating between the application view and the management view forces the user to log in again. The application site should preserve the authenticated session so students move between apply and management without re-authenticating.

**Expected:** A single persistent session across both views; no re-login when switching.

PR #1973, opened 2026-07-21
`Application`: Fix submit-success modal navigation and stop logging out on back

### LINKED ISSUE #1725: Application submit flow: broken modal & navigation, useless "Go to App" button
After submitting an application the flow is broken in several related ways:

- The submit-success modal links to an overview page that **logs the user out** and returns them to the apply view, where they can't even see that they've already applied. It should instead lead to the management view (`/management/courses`).
- **Closing the modal** also returns the user to the application site and logs them out, instead of simply closing.
- The **"Go to App" button** is only clickable while logged into the application form and **cannot be used after submitting**.
- There is **no obvious way to navigate back** to the application site.
- It is not clear that clicking **"Login"** is what enters the app/management area — there should **always** be a clear "Go to App" entry point.

**Expected:** Success modal closes cleanly and routes to management without logout; a persistent, always-visible "Go to App" entry point; a clear path back to the application site.

rappm:
## Summary

Fixes the broken application submit flow: finishing or closing the success modal no longer logs the user out, and there is now a clear, working way into the management view and back to the application site.

## Details

In `ApplicationAuthenticated` the "back"/success actions previously called `logout()`, which hit the Keycloak logout endpoint and returned the user to the apply site logged out. This PR removes that:

- **Success modal button** now routes to the management view (`/management/courses`) and is labelled **"Go to App"**, instead of calling `logout()`.
- **Closing the success modal** (X / overlay / Esc) now simply closes it (`ApplicationSavingDialog` `onOpenChange` always calls `onClose`), instead of logging out and returning to the apply site.
- **Application header "Back to Overview"** now navigates back to the application site landing (`/`) instead of logging out.
- `ApplicationSavingDialog` gained an optional `navigateBackLabel` (defaults to "Back to Overview", so the external/unauthenticated flow is unchanged).

Because the flow no longer destroys the Keycloak session, the header **"Go to App"** entry stays visible and usable after submitting (it is shown whenever a user is logged in), and switching to the management view no longer forces a re-login.

## Reason / Link to issue

Closes #1725.

Partially addresses #1724: the forced re-login when moving between apply and management is removed (the session is no longer destroyed). The remaining redirect round-trip on switching is caused by a separate Keycloak initialization/persistence bug and will be fixed in a dedicated follow-up PR.

## How to Test

1. Open an application phase's authenticated form (`/apply/:phaseId/authenticated`) while logged in.
2. Fill in and submit the application → the "Application Saved" dialog appears.
3. Click **Go to App** → you land on `/management/courses` and are still logged in (no logout, no credential prompt).
4. Submit again and this time close the dialog via the X / clicking outside / Esc → the dialog just closes and you stay on the form, still logged in.
5. Click the header **Back to Overview** → you return to the application site landing without being logged out. The header still shows a **Go to App** entry.

## PR Checklist

### DIFF
--- a/clients/core/src/publicPages/application/components/ApplicationSavingDialog.tsx
+++ b/clients/core/src/publicPages/application/components/ApplicationSavingDialog.tsx
@@ -14,6 +14,7 @@ interface ApplicationSavingDialogProps {
   showDialog: 'saving' | 'success' | 'error' | null
   onClose: () => void
   onNavigateBack: () => void
+  navigateBackLabel?: string
   errorMessage?: string
   confirmationMailSent?: boolean
 }
@@ -22,6 +23,7 @@ export const ApplicationSavingDialog = ({
   showDialog,
   onClose,
   onNavigateBack,
+  navigateBackLabel = 'Back to Overview',
   errorMessage,
   confirmationMailSent = false,
 }: ApplicationSavingDialogProps) => {
@@ -92,10 +94,8 @@ export const ApplicationSavingDialog = ({
   return (
     <Dialog
       open={isOpen}
-      onOpenChange={() => {
-        if (showDialog === 'success') {
-          onNavigateBack()
-        } else {
+      onOpenChange={(open) => {
+        if (!open) {
           onClose()
         }
       }}
@@ -125,7 +125,7 @@ export const ApplicationSavingDialog = ({
             </Button>
           ) : showDialog === 'success' ? (
             <Button onClick={onNavigateBack} className='w-full sm:w-auto'>
-              Back to Overview
+              {navigateBackLabel}
             </Button>
           ) : (
             <Button variant='outline' onClick={onClose} className='w-full sm:w-auto'>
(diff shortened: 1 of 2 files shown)


magkue (approved):
Tested the authenticated submit flow locally — confirmed the core fix works: **Go to App**, closing the success modal, and header **Back to Overview** all navigate correctly and **no longer log the user out**. 👍

**UX suggestion on the success modal (from testing):** because `onOpenChange` now dismisses the success dialog to the underlying page, closing it (X / overlay / Esc) leaves the user back on the application form with an active **Submit** button — which invites a confusing re-submit. Suggest making the success state **terminal** rather than dismissable: drop the close affordance on success and offer two explicit actions — **"Go to App"** (→ management) and **"Back to Overview"** (→ `/`), both without logout — so there's no path back onto the submittable form.

**Shared-component side effect (confirm intended):** the `onOpenChange` change also affects the other consumer, `ApplicationLoginPage` (external/unauthenticated flow, untouched by this PR) — closing its success modal now dismisses (`setShowDialog(null)`) instead of `navigate('/')`. This is the same "back onto the form" situation as above, so worth handling consistently.

**Nit (optional):** `onNavigateBack` is now also wired to a forward action ("Go to App" → management), so the name is slightly misleading — consider `onPrimaryAction`.

PR #1974, opened 2026-07-21
`UI/UX`: Add dark mode support to the public application pages

### LINKED ISSUE #1727: Application page has no dark mode
The application page does not support dark mode, unlike the rest of the app. Add dark-mode styling.

rappm:
## Summary

Adds dark-mode support to the public application pages (application form, landing, legal), which previously always rendered in light mode unlike the rest of the app.

## Details

- `NonAuthenticatedPageWrapper` removed the `dark` class on mount and hardcoded a white background; it also lacked a `DarkModeProvider`. It now wraps its tree in `DarkModeProvider` (which defaults to the system theme) and uses `bg-background text-foreground`.
- `AuthenticatedPageWrapper` now wraps its main content (not just the loading state) in `DarkModeProvider` and uses theme tokens for the background.
- Added an always-visible `ThemeToggle` to the public header so applicants can switch light/dark/system.
- Replaced hardcoded light-only colors (`bg-white`, `text-gray-*`, `text-slate-*`) with theme tokens (`bg-background`, `bg-muted`, `text-foreground`, `text-muted-foreground`, `hover:*`) across the shared header/footer, landing page, about page, contributor list, and the application text-question form so they render correctly in dark mode. `InfoBanner` already had dark variants and was left as-is.

## Reason / Link to issue

Closes #1727.

## How to Test

1. Set your OS to dark mode (or use the new header theme toggle) and open a public application page (`/apply/:phaseId`), the landing page (`/`), and the About page.
2. Verify the background, text, header, and footer render in dark mode with readable contrast (no white boxes / black-on-dark text).
3. Toggle light/dark/system via the header toggle and confirm it switches immediately.
4. Confirm the management console still themes correctly (unchanged).

## PR Checklist

### DIFF
--- a/clients/core/src/publicPages/application/pages/ApplicationForm/components/TextForm/ApplicationQuestionTextForm.tsx
+++ b/clients/core/src/publicPages/application/pages/ApplicationForm/components/TextForm/ApplicationQuestionTextForm.tsx
@@ -94,7 +94,7 @@ export const ApplicationQuestionTextForm = forwardRef(function ApplicationQuesti
                       disabled={isInstructorView}
                     />
                   )}
-                  <div className='absolute right-2 bottom-2 text-sm text-gray-500'>
+                  <div className='absolute right-2 bottom-2 text-sm text-muted-foreground'>
                     {charCount}/{question.allowedLength || 255}
                   </div>
                 </div>
--- a/clients/core/src/publicPages/landingPage/LandingPage.tsx
+++ b/clients/core/src/publicPages/landingPage/LandingPage.tsx
@@ -19,17 +19,17 @@ export function LandingPage() {
   return (
     <NonAuthenticatedPageWrapper>
       <section className='text-center mb-16'>
-        <h2 className='text-2xl font-bold text-gray-900 sm:text-3xl lg:text-4xl mb-4'>
+        <h2 className='text-2xl font-bold text-foreground sm:text-3xl lg:text-4xl mb-4'>
           Course Application Portal
         </h2>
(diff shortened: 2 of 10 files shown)


magkue (approved):
Tested the public pages locally in light/dark/system. Dark mode works well on the pages wrapped in `DarkModeProvider` (landing `/`, apply `/apply/:phaseId`): background, text, header, footer, and the new `ThemeToggle` all render with good contrast. Clean token-swap refactor overall. Approving — the notes below are non-blocking.

**1. Legal/About pages get token swaps but no provider.** `AboutPage`, `ImprintPage`, and `PrivacyPage` don't use `NonAuthenticatedPageWrapper`, so they have no `DarkModeProvider` and no `ThemeToggle`. On a direct load, `/about` ignores the stored theme and renders light (confirmed: `theme=dark` in storage left `<html>` without the `dark` class on that route). It only appears dark if you navigate to it within the SPA after toggling elsewhere, where the root `.dark` class persists. Since the PR description lists "legal" pages as supported and suggests testing the About page in dark, consider wrapping these pages (or the public route subtree) in `DarkModeProvider` too, so the theme is honored on direct load and the toggle is available there.

**2. Hover regression in `AboutPage`.** The "Get in Touch" links became `bg-muted hover:bg-muted` (previously `bg-gray-50 hover:bg-gray-100`), so hover no longer changes the background. Suggest `hover:bg-accent` (or `hover:bg-muted/80`).

**3. Nit.** `Header.tsx` version span uses `text-s`, which isn't a valid Tailwind size (likely `text-sm`/`text-xs`). Pre-existing, but it sits on a line this PR touches.

PR #1975, opened 2026-07-21
`Core`: Only send status mails once per participant

### LINKED ISSUE #1363: Only send Acceptance Mails once for accepted students
## 🚀 Feature Request Description

Confirmation emails upon successful completion of the “Application Phase.” Once students are marked as accepted, the “Acceptance Emails” are sent out via the mailing system. These are automatically sent to everyone with the “Accepted” status.
If I want to invite applicants who are currently in the “Ongoing” status to interviews, the emails should not be sent twice to applicants who have already received an invitation.  

## 🧩 Problem Case

It causes confusion when participants receive the interview invitation twice. 

## 💡 Proposed Solution

Emails should only be sent to participants who have been marked as “accepted” since the last mailing.

rappm:
## Summary

Status/acceptance mails are no longer sent twice to the same participant. Previously every trigger mailed all participants with the target pass status, so re-sending (e.g. after accepting more students) re-notified people who had already received the mail.

## Details

- `ClaimStatusMailRecipients` claims the recipients in a single `UPDATE … WHERE (restricted_data -> 'statusMailSentAt' ->> status) IS NULL … RETURNING …` and returns the claimed participants with their student data. Because claiming and selecting happen in one statement, two concurrent triggers for the same phase and status cannot pick up the same participant, and no transaction or advisory lock is held across the SMTP sends.
- The marker is a per-participation, per-status timestamp in `course_phase_participation.restricted_data` under `statusMailSentAt.<status>` (JSONB, no schema migration, following the existing assessment-reminder precedent). Claiming handles `NULL` `restricted_data` and a non-object `restricted_data` or `statusMailSentAt`, and preserves all other keys.
- A failed send releases its own claim again (`ReleaseStatusMailClaim`), so failed recipients stay eligible for the next trigger while successful ones do not.
- The two recipient reads the status path used before (`GetParticipantMailingInformation`, `GetParticipantMailingInformationByIDsAndStatus`) are gone, replaced by the claim. `GetParticipantMailingInformationByIDs` (manual mail) now also returns `course_participation_id`, so its row type is the one shared with the status path.

**Delivery semantics:** at-most-once. The marker is written before the send, so a crash between claim and send drops that mail rather than duplicating it, and a delivered-but-unrecorded mail is no longer possible (this is why the earlier `markFailures` report field is gone again). The residual gap is a claim that cannot be released after a failed send: that participant is reported under `failedEmails` but will not be picked up by a retry.

**Follow-up:** there is still no supported re-send path after a template typo or a bounce - clearing markers currently requires hand-written SQL. Tracked in #2054.

## Reason / Link to issue

Closes #1363.

## How to Test

1. In an Application phase, accept several students and send acceptance mails.
2. Accept a few more students and send acceptance mails again.
3. Verify only the newly-accepted students receive the mail; previously-mailed students are not mailed again.
4. Rejection mails behave the same, independently of acceptance mails (a passed-then-failed participant still receives the rejection mail).

## PR Checklist


mathildeshagl (changes requested):
The dedup idea and the SQL behind it are right. The transaction wrapper around it is where I would like changes before merge.

**1. `servers/core/mailing/service.go:130-190` — SMTP sends run inside an open transaction holding a per-phase advisory xact lock.** Each trigger pins a pool connection (pgxpool default max is `max(4, NumCPU)`) for the whole batch, at 10s connect + 15s deadline per mail, while concurrent triggers block on the lock and pin connections of their own. A few large sends can exhaust the core DB pool. This also contradicts the PR description, which says holding a DB connection across SMTP sends was deferred.

**2. `service.go:176-186` — the per-participant `MarkFailures` accumulation cannot work in Postgres.** The first failing `MarkStatusMailSent` aborts the transaction (25P02); every later mark then fails, `tx.Commit` fails, and the deferred rollback discards *all* markers even though the mails were delivered. The next trigger re-sends everyone, which is precisely the bug this PR fixes. The same happens if the gin request context is canceled (client disconnect, shutdown) mid-loop. Markers need to commit incrementally, outside a long-lived transaction.

**3. `service.go:188-190` + `mailing/router.go:52`** — on commit failure the report is returned alongside a non-nil error, so the router emits 500 and drops the body. After a partial send the lecturer gets no list of who was actually mailed.

**4. `mailing/mailingDTO/report.go:6`** — `markFailures` is server-only. `MailingReport` lives in the external `@tumaet/prompt-shared-state` package and nothing in `clients/` reads the field (`clients/core/src/network/mutations/sendStatusMail.ts` is the only reference and appears unused), so the "surface instead of swallow" intent does not reach the UI. Needs a shared-state + UI follow-up.

**5. `service.go:165`** — calls `SendCourseMail` directly although the package already has the `sendMailFn` seam (`mailing/manualMail.go:15`) that `manualMail_test.go` swaps. Routing through it would let the send/mark loop — the actual fix — be tested. `service_test.go` currently covers only the SQL primitives.

**6. No supported re-send path.** After a template typo or a bounced mail, markers can only be cleared with hand-written SQL. Worth an explicit product decision rather than leaving it implicit.

Smaller things: `hashtextextended` is an undocumented internal; `to_jsonb(now())` stores transaction-start time in a format that differs from the `UpdateAssessmentReminderLastSentAt` precedent; the lock test depends on a 100ms timing window; the added narrative comments run against the repo's comment-density style; there is unrelated trailing-whitespace churn in `db/query/mailing.sql`.

What is good: the dedup SQL correctly handles NULL `restricted_data` and a non-object `statusMailSentAt` while preserving other keys, per-status independence is explicitly tested, and the sibling `GetParticipantMailingInformationByIDs` row type was kept in sync so the struct conversion in `manualMail.go:78` still compiles.

servers/core/mailing/service.go
```diff
@@ -130,14 +130,17 @@ func SendStatusMailManualTrigger(ctx context.Context, coursePhaseID uuid.UUID, s
 	}
 
 	// 3.) Get the participants to send the status mail to. A recipient list narrows the mail down to
-	// those participants; either way only participants with the given status are mailed.
+	// those participants; either way only participants with the given status that have not been
+	// mailed for it yet are included.
 	participants, err := getStatusMailRecipients(ctx, coursePhaseID, status, recipientCourseParticipationIDs)
```
mathildeshagl: This read is now completely unserialized. With the advisory lock from `47e6ba48` removed, two concurrent triggers for the same phase and status both reach here, both get the same rows (all with `statusMailSentAt.<status> IS NULL`), and both go on to send. A double-clicked button or a retried request is enough.

Claiming instead of reading would close it without reintroducing what I objected to in my first review:

```sql
UPDATE course_phase_participation
SET restricted_data = ... -- the MarkStatusMailSent expression
WHERE course_phase_id = $1
  AND pass_status = $2
  AND (restricted_data -> 'statusMailSentAt' ->> $2::text) IS NULL
RETURNING course_participation_id, ...
```

One statement, no lock, no connection held across SMTP. It flips the failure mode to at-most-once — a crash between claim and send drops a mail instead of duplicating one — which is a genuine trade-off worth making on purpose. Whichever way you go, the description's "Known limitation" paragraph should describe what is actually there now.

servers/core/mailing/service.go
```diff
@@ -150,9 +153,22 @@ func SendStatusMailManualTrigger(ctx context.Context, coursePhaseID uuid.UUID, s
...
+		if markErr := MailingServiceSingleton.queries.MarkStatusMailSent(ctx, db.MarkStatusMailSentParams{
+			CoursePhaseID:         coursePhaseID,
+			CourseParticipationID: participant.CourseParticipationID,
+			Status:                string(status),
+			SentAt:                sentAt,
+		}); markErr != nil {
+			// The mail was delivered but the dedup marker could not be persisted, so a future send
+			// could duplicate it. Surface this instead of swallowing it silently.
+			log.Error("failed to mark status mail as sent for participant: ", markErr)
+			response.MarkFailures = append(response.MarkFailures, participant.Email.String)
```
mathildeshagl: The accumulation works correctly now that each mark is its own statement — this is the part my earlier point (2) was about and it is fixed.

The remaining half of that point is that nothing consumes it. `MailingReport` in `@tumaet/prompt-shared-state` has no `markFailures` field, so a lecturer whose mail was delivered but not recorded still sees nothing. Right now the "surface instead of swallow" intent stops at the JSON boundary.

servers/core/mailing/mailingDTO/report.go
```diff
@@ -3,4 +3,5 @@ package mailingDTO
 type MailingReport struct {
 	SuccessfulEmails []string `json:"successfulEmails"`
 	FailedEmails     []string `json:"failedEmails"`
+	MarkFailures     []string `json:"markFailures"`
```
mathildeshagl: This field is server-only. The client reads `MailingReport` from the external `@tumaet/prompt-shared-state` package, which does not carry it, and nothing under `clients/` references it.

Not blocking on its own, but it needs a shared-state release plus a UI change to mean anything, and it is easy for that follow-up to be forgotten once this merges — the field looks done from inside this repo. Worth an issue linked from here.

mathildeshagl (changes requested):
Four of my six points are resolved, and the transaction rework in `ca041e68` is the right shape:

- SMTP no longer runs inside an open transaction, so no pool connection is pinned across a batch (1).
- Each marker is its own statement, so one failure no longer aborts the rest — and `TestSendStatusMailRetriesFailedRecipients` proves a failed send leaves no marker while a successful one is not re-sent (2).
- The trigger now always returns the report, so a partial send is a 200 with the list instead of a 500 with nothing (3).
- Sends go through the `sendMailFn` seam and `statusMailDedup_test.go` covers the send-and-mark loop end to end (5).
- The smaller notes are all addressed too: `hashtextextended` is gone with the lock, the timestamp is an RFC3339 text value from `nowFn` matching the `UpdateAssessmentReminderLastSentAt` precedent, the timing-dependent lock test is gone, and the whitespace churn in `db/query/mailing.sql` is reverted.

Requesting changes on the concurrency guard, which went the wrong way, and on the two points still outstanding.

**The guard was removed rather than narrowed.** `47e6ba48` added `pg_advisory_xact_lock`; `ca041e68` deletes it and puts nothing in its place. There is now no serialization at all, so two concurrent triggers both read the recipient list, both see `statusMailSentAt.<status> IS NULL` for the same people, and both send. That is the duplicate this PR exists to prevent. I know the description lists it as a known limitation and I am not asking you to hold a connection across SMTP again — the point is that it moved from a narrow window inside a lock to no bound at all, while the title claims the stronger guarantee. Claiming the recipients first (`UPDATE … WHERE marker IS NULL RETURNING …`) and then sending outside any transaction gets you both: no lock, no connection held, and at-most-once instead of at-least-once. That is a real trade-off — a crash between claim and send loses a mail rather than duplicating one — so it is worth deciding deliberately rather than by omission.

**Still open:** `markFailures` does not reach the UI (4), and there is still no supported re-send path after a template typo or a bounce (6). Both are follow-up-sized, but I would rather they be tracked than left implicit — a field that only exists server-side reads as done when it is not.

servers/core/mailing/service.go
rappm: Went with the claim. `ClaimStatusMailRecipients` is one `UPDATE … WHERE (restricted_data -> 'statusMailSentAt' ->> status) IS NULL … RETURNING …` wrapped in a CTE that joins the student data onto the returned rows, so the eligibility check and the marker write are the same statement and the recipient list arrives with it. No lock, no transaction across SMTP, and the recipient-list variant is the same statement with a nullable `uuid[]` filter.

Semantics are at-most-once now, deliberately. A failed send releases its own claim (`ReleaseStatusMailClaim`, a `#-` on the status key only), so failed recipients still retry, which `TestSendStatusMailRetriesFailedRecipients` covers. What is genuinely lost is a crash between claim and send, plus a release that itself fails after a failed send: that participant is reported under `failedEmails` but a retry will not pick it up again.

`TestSendStatusMailConcurrentTriggersMailOnce` pins the case you described without a timing window: the fake `sendMailFn` holds the first trigger open inside its first send and only releases it after a second, overlapping trigger has run to completion, so the second one provably runs after the first claim and must come back empty.

Description rewritten - the "Known limitation" paragraph is now a "Delivery semantics" one that states at-most-once and names the residual gap.

servers/core/mailing/service.go
rappm: Resolved by removing the field rather than by wiring it through. With the marker written before the send, a delivered-but-unrecorded mail cannot happen any more, so there is nothing left for `markFailures` to report and no shared-state release or UI change is needed. `MailingReport` is back to `successfulEmails` / `failedEmails`, and the swagger output is identical to main again.

The one failure that survives - a claim that cannot be released after a failed send - shows up in `failedEmails`, since the send did fail; it is logged as not retryable.

servers/core/mailing/mailingDTO/report.go
rappm: Field is gone, so there is no server-only surface left to follow up on - see the reply on `service.go`. Point 6 (a supported re-send path after a template typo or a bounce) is the one that still needs tracking, and it matters slightly more now that the semantics are at-most-once. Writing that up as a separate issue and linking it here.

rappm: Filed as #2054.

servers/core/mailing/service.go
```diff
@@ -150,39 +162,25 @@ func SendStatusMailManualTrigger(ctx context.Context, coursePhaseID uuid.UUID, s
...
-	for _, participant := range selected {
-		participants = append(participants, db.GetParticipantMailingInformationRow(participant))
+func releaseStatusMailClaim(ctx context.Context, coursePhaseID, courseParticipationID uuid.UUID, status db.PassStatus) {
+	if err := MailingServiceSingleton.queries.ReleaseStatusMailClaim(ctx, db.ReleaseStatusMailClaimParams{
+		CoursePhaseID:         coursePhaseID,
+		CourseParticipationID: courseParticipationID,
+		Status:                string(status),
+	}); err != nil {
+		log.Error("failed to release status mail claim for participant, it will not be retried: ", err)
 	}
```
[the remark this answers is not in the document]
rappm: Not doing this. `releaseStatusMailClaim` only runs after a failed send, and that path already appends the recipient to `failedEmails` before attempting the release, so a lecturer reading the report cannot mistake a stuck claim for a delivered mail - the mail is reported as failed either way.

Adding a second list for release failures is the `markFailures` field that was just removed from this PR: `MailingReport` lives in the external `@tumaet/prompt-shared-state` package, so a new field would need a package release plus a UI change before it meant anything, and until then it reads as done from inside this repo. Returning a non-nil error instead would turn a partial send into a 500 that drops the report, which was also raised against an earlier revision.

What is left is that the stuck participant cannot be re-mailed without hand-written SQL, and that is #2054.

PR #1995, opened 2026-07-23
`Docs`: Add a Quickstart guide for local setup

magkue:
## Summary

Adds a root `QUICKSTART.md` — a concise, few-minutes path to a running local PROMPT 2.0 stack: env setup, client install, starting the servers via Docker, starting the clients via yarn, and opening the app. Complements the fuller `docs/contributor/setup.md`.

## Details

- Environment setup (`.env` / `.env.dev` from the templates).
- Install client dependencies (`cd clients && yarn install`).
- Start the servers in Docker (`docker compose up server-core …`), which also brings up the databases and Keycloak.
- Start the clients with `yarn dev` (or `make clients`).
- Open the app and a few useful commands (`make lint`, `make test`, `make db-down`).
- Points to the `keycloak-local-setup` skill / `docs/contributor/setup.md` for auth specifics.

## Reason

Lowers the barrier for new contributors to get a stack running quickly without reading the full contributor guide first.

### DIFF
--- a/QUICKSTART.md
+++ b/QUICKSTART.md
@@ -1,0 +1,53 @@
+# Quickstart
+
+Get a local PROMPT 2.0 stack running in a few minutes.
+
+## 1. Environment setup
+
+```bash
+cp .env.template .env
+cp .env.dev.template .env.dev
+```
+
+`.env` holds Docker hostnames; `.env.dev` overrides them with localhost values for local
+development. For Keycloak/auth specifics, see the `keycloak-local-setup` skill or
+`docs/contributor/setup.md`.
+
+## 2. Install client dependencies
+
+```bash
+cd clients && yarn install && cd ..
+```
+
+## 3. Start the servers (Docker)
+
+```bash
+docker compose up server-core server-assessment server-team-allocation
+```
+
+This also brings up the required databases and Keycloak. Add or drop `server-*` services as
+needed (see `docker-compose.yml` for the full list).
+
+## 4. Start the clients (yarn)
+
+Run the core shell plus whichever component you're working on, each in its own terminal:
+
+```bash
+cd clients/core && yarn dev                   # app shell, port 3000
+cd clients/assessment_component && yarn dev   # port 3007
+cd clients/team_allocation_component && yarn dev  # port 3008
+```
+
+Or start everything at once: `make clients` (or `cd clients && yarn dev`).
+
+## 5. Open the app
+
+Visit <http://localhost:3000> and log in with a local Keycloak user.
+
+## Useful commands
+
+```bash
+make lint      # lint everything
+make test      # run tests
+make db-down   # stop database and Keycloak
+```


Mtze (changes requested):
There is a quickstart guide in the readme - Especially the inital keycloak setup and env setup is a bit more involved than shown here 

If we add this - we should make it consistent and complete ;)

mathildeshagl (changes requested):
Docs-only, so no code risk, but following the guide end to end does not currently produce a working stack. The blocking gaps:

**1. `QUICKSTART.md:8-11`, `:46` — missing Keycloak client secret step.** Step 1 copies `.env.template`, which ships `KEYCLOAK_CLIENT_SECRET= # FIXME` (`.env.template:120`; `.env.dev.template:35` is also empty). The guide never says to generate and paste it, so `server-core` cannot validate tokens and step 5 ("log in with a local Keycloak user") fails. Both `readme.md` and `docs/contributor/setup.md:104-117` treat this as a required manual step.

**2. `:17-19` — no prerequisites.** `yarn install` needs Corepack enabled for Yarn 4 (`clients/package.json` -> `packageManager: yarn@4.17.1`), and Homebrew strips corepack, so `readme.md` documents `brew install corepack && corepack enable` as mandatory. The guide lists no prerequisites at all (Go 1.26, Node 22.10.x, Docker).

**3. `:50-52` — `make db-down` is the wrong teardown.** It is only `docker compose stop db keycloak` (`Makefile:89`), which leaves the three servers, `db-assessment`, `db-team-allocation`, `keycloak-db`, and SeaweedFS running. This flow wants `docker compose down`.

Worth deciding before merge:

**4. Duplication with the readme.** `readme.md` "Getting Started" is already a quickstart, and it prescribes a different path (native `make server` + `.env.dev`) than this doc (Dockerized servers). Nothing links to `QUICKSTART.md` from `readme.md`, `CONTRIBUTING.md`, or `docs/contributor/setup.md`, and it sits outside the Docusaurus `docs/` site. Two guides that can silently diverge. Would it be better inside `docs/`, replacing the readme section rather than sitting alongside it?

**5. `:12-14`** points human contributors at the `keycloak-local-setup` skill, which is an agent artifact in `.agents/skills/`. `docs/contributor/keycloak-dev.md` is the human-facing equivalent.

**6. `:9`** creates `.env.dev`, which this flow never uses — `docker compose` reads only `.env`. It matters solely for native `make server`.

**7. `:51`** — "run tests" oversells `make test`, which runs Go server tests only (`Makefile:110`).

**8. `:34-36`** — the three `cd` lines only work when run from the repo root in separate terminals. The prose says so, but pasting the block will not work.

To be clear about what is good here: every verifiable technical claim checks out — compose service names, the `depends_on` chain (`server-core` -> db, keycloak, seaweedfs-s3), ports 3000/3007/3008 (`rspack.config.mjs`), and `make clients` being equivalent to `cd clients && yarn dev`. The prose is tight and well formatted.

rappm:
@magkue Any update on this?

PR #1845, opened 2026-07-06
`Assessment`: Add clean PDF/print export for assessments and tutor evaluations

rappm:
# 📝 Pull Request Template

Use this template to provide all necessary information for reviewing and testing your changes.

## ✨ What is the change?

Makes the assessment printable/exportable as a clean PDF instead of the current unusable browser print.

- **Core shell:** hides the sidebar, top header, and footer when printing (`print:hidden`) and makes the content container print-friendly, so any management page prints without the surrounding chrome.
- **Dedicated print reports:** new read-only report components rendered `hidden print:block`. Because they are always mounted, collapsed categories are no longer silently omitted, and background score colors are preserved (`print-color-adjust: exact`). The interactive page is hidden in print via `print:hidden`, so both `Ctrl+P` and the new **PDF / Print** action produce the same nicely ordered document through the browser's *Save as PDF*.
- **Lecturer assessment page:** **PDF / Print** added to the bottom Export menu.
- **Student released-results page:** **PDF / Print** button, respecting the `gradingSheetVisible` / `gradeSuggestionVisible` / `actionItemsVisible` flags so nothing hidden-by-config leaks into the PDF.
- **Per-tutor evaluation page** (opened from the tutor overview table): **PDF / Print** button plus a matching report with per-category averaged scores and descriptions, the individual student scores, and the tutor feedback items.

No new dependency: everything uses the browser's native print. Shared score/label/description helpers (`getScoreLevelDescription`, `getWeightedScoreLevel`, `getLevelConfig`, score mapping) are reused rather than duplicated.

## 📌 Reason for the change / Link to issue

The previous export dragged in the whole shell, printed raw interactive controls, dropped the background colors, and silently omitted any category that was collapsed on screen. This delivers a clean, complete, well-ordered PDF for student assessments and tutor evaluations.

## 🧪 How to Test

1. `make clients` and open a lecturer assessment with **a category collapsed** on screen.
2. Press `Ctrl/Cmd+P` (or **Export → PDF / Print**): the preview shows only the clean report — no sidebar/header/footer, the collapsed category is present, score colors render, sensible page breaks. Choose *Save as PDF*.
3. As a student with released results: use the **PDF / Print** button and confirm the report respects the grading-sheet / grade-suggestion / action-item visibility flags.
4. From the **Tutor Overview** table, click a tutor, then **PDF / Print**: confirm per-category scores, individual student scores, and feedback render.

## 🖼️ Screenshots (if UI changes are included)

| Before         | After         |
[...]

PR #2014, opened 2026-08-06
`Assessment`: Add PDF export for self, peer and tutor evaluations

### LINKED ISSUE #1965: Assessment: PDF Export of Self & Peer Evaluations
### Problem or opportunity

Currently the only way to export an Assessment or Evaluation is to screenprint / print the page. The layout of the print page is not ideal. 

### Desired outcome

It would be cool to have an export button in the Student / Tutor evaluation/assessment details pages to generate a report that can be exported and downloaded. Optionally a bulk export of all participants would be cool for archival reasons. 

### Proposed solution

_No response_

### Alternatives considered

_No response_

### Additional context

_No response_

rappm:
## Summary

Self and peer evaluation results had no export at all, and the tutor evaluation page had an export
button that printed its live interactive DOM. This adds a purpose-built print report to all three,
reusing the one the assessment page already had, plus a bulk export of every participant's report
for archival.

## Details

**A reusable report.** The clean printable layout only existed for assessments, wired directly to
the assessment Zustand store. It is now a props-driven `PrintReport` (title, subtitle, header fact
row, categories, scores, per-category comments, feedback items, and a `children` slot for a trailing
summary). `AssessmentPrintReport` keeps its props and becomes a thin store adapter over it, so both
of its call sites and the existing print e2e test are untouched.

`PrintReport` renders a competency's *average* score level, which is what lets one component serve
both domains: an assessment has one grader per competency, an evaluation has N authors. When a
competency has more than one score it also lists each score with its author. The single-score path
is provably unchanged: `mapScoreLevelToNumber` maps the five levels to 1–5 and
`mapNumberToScoreLevel` buckets at 1.5/2.5/3.5/4.5, so averaging one score round-trips exactly.

**Why the tutor page needed more than a button.** Its print output was bad, not merely unpolished.
`CategoryEvaluation` renders `ScoreLevelSelector` with `hideUnselectedOnDesktop={false}` inside
`md:grid-cols-5`. A real A4 page box is ~680px wide after the `@page { margin: 1.5cm }`, below the
`md` breakpoint, so the grid collapses to one column and every competency prints five stacked
full-width rubric boxes. A 3-category × 4-competency rubric printed roughly 60 boxes. All three
pages now wrap their interactive tree in `print:hidden` and render a `PrintReport` sibling, the same
shape `AssessmentPage` already used.

For the same reason `PrintReport` deliberately uses no responsive utilities: print emulation in
tests keeps the full viewport while a real print does not, so `md:`/`lg:` variants would render
differently in the two contexts.

**Bulk export.** `PDF / Print All` on the self and peer overview pages. Those pages already load
every evaluation in the phase, so this needs exactly one extra request for the phase's feedback
items, against an endpoint that already existed with the same roles as the page. **No backend
change.** Participants with no evaluation of that type are skipped rather than printed empty, and
the reports are mounted only for the duration of a print request — always mounting them would add a
hidden competency card per participant per competency to a page that already renders two charts and
the full participant table.

The print request is a counter rather than a flag, for two independent reasons: React Query reports
`isSuccess` for cached data even while a query is disabled, so a flag would let the first click
print before any report was mounted; and a latched flag would make the second click a silent no-op
because the effect would not re-run.

**Privacy.** Peer and tutor reports name which team member gave which score, mirroring the author
names the on-screen view already shows in its tooltips. All the routes involved are restricted to
`PromptAdmin` and `CourseLecturer`; nothing here is reachable by students, and the student-facing
visibility flags (`evaluationResultsVisible`, `gradingSheetVisible`) gate different surfaces.

Self and peer share one page component, so one change covers both types.

## Reason / Link to issue

Closes #1965.

The issue opens with "the layout of the print page is not ideal" and asks for an export button on
the evaluation detail pages, plus an optional bulk export for archival. #1845 covered assessments
properly and gave the tutor page a button only; per the issue discussion, self and peer had nothing.

Worth stating up front: there is no PDF library in this repo, and this PR does not add one. "PDF
export" here means `window.print()` plus print CSS, with the browser dialog producing the PDF — the
mechanism #1845 established. The server-side `/student-assessment/:id/export` endpoint is JSON-only
and explicitly rejects `?format=pdf`.

## How to Test

Configure an assessment phase with self, peer and tutor evaluations enabled and some submitted
evaluations, then:

1. `/self-evaluations/:courseParticipationID` and `/peer-evaluations/:courseParticipationID` — a
   `PDF / Print` button sits bottom-right. In the print preview the report should be a single clean
   flow: no sidebar, no five-box rubric grids, no Radix tooltips, sane page breaks.
2. The peer report should show per-evaluator score/name rows; the self report a single score per
   competency.
3. `/tutors/:tutorId` — same, and confirm the old five-box output is gone.
4. `/participants/:courseParticipationID` and the student `/results` page — output must be
   **unchanged**. This is the refactor's regression surface; compare side by side against `main`.
5. `/self-evaluations` and `/peer-evaluations` — `PDF / Print All` shows a spinner, then prints one
   report per participant, one participant per page.

Automated: `make test-e2e` (two new tests in `e2e/tests/assessment/print.spec.ts`).

## Screenshots

Not attached. Producing them needs a running stack with a configured multi-evaluator phase; the
print-preview comparison in steps 1–4 above is the thing a reviewer should actually look at.

## PR Checklist

Notes on the unchecked boxes:

- **Tested locally:** `tsc --noEmit` is clean and `biome check` reports zero errors across all 1018
  client files with no new warnings. The Playwright suite was still running when this PR was opened
  and the manual print-preview walkthrough above has not been done yet — both need doing before
  merge.
- **Screenshots:** see the Screenshots section.
- **Documentation:** no user or contributor docs describe the print export, so there is nothing to
  update.

E2E coverage is self-evaluation only. Peer and tutor reports need seeded `team_allocation` team
data, which the assessment e2e fixtures do not provide, so the author-name lookup and the
multi-evaluator rows are covered by the manual steps rather than by tests.

## Summary by CodeRabbit

* **New Features**
  * Added printable assessment reports for participant and tutor results.
  * Added individual and bulk PDF/print options with scores, competencies, teams, evaluator details, and feedback.
  * Added categorized positive and negative feedback sections and score indicators.
  * Added feedback retrieval for course phases.

* **Bug Fixes**
  * Improved report generation by excluding participants without evaluations and waiting for feedback before printing.

* **Tests**
  * Added coverage for individual and bulk assessment printing workflows, including self-evaluations.


mathildeshagl (dismissed):
Good extraction. Pulling the shared `PrintReport` out of `AssessmentPrintReport` and having the assessment report become a caller with `children` for its completion section is the right factoring, and the note about deliberately avoiding responsive utilities — because A4 is below the `md` breakpoint while Playwright's print emulation keeps the viewport — is exactly the kind of thing that would otherwise be rediscovered painfully later.

The one thing I checked carefully was whether the per-rater `authorName` list de-anonymizes peer feedback, since a printed PDF travels further than a page view. It does not: `CategoryEvaluation` already renders `getTeamMemberName(teams, evaluation.authorCourseParticipationID)` on screen, so the report matches what the lecturer already sees, and these are the lecturer-facing routes rather than the student results page. Worth stating explicitly because #2016 introduces a two-rater anonymity floor for the student-facing view, and the two changes look contradictory until you notice they have different audiences.

Some notes, none blocking:

**1. `getTeamMemberName` only searches `team.members`.** An author who has since left the team renders as "Unknown member". That is pre-existing behavior, but a PDF makes it durable in a way a transient screen does not — a printed report with "Unknown member" next to a score is harder to explain to a student than the same string on a page a lecturer can refresh.

**2. The per-competency description now derives from the average rather than an actual rating.** `getScoreLevelDescription(averageScoreLevel, competency)` is identical to the old behavior for assessments, where there is one score per competency. For a multi-rater peer report it can print the rubric text for a level nobody actually awarded. Probably what you want given the chip above it shows the same average, but it is a real change in what the sentence means, so worth a line in the description.

**3. The single-participant report carries `Evaluators: N` in `meta`; the bulk report omits it.** A lecturer comparing the two side by side will notice. Same for the fact that only the single-participant page gates its print button on `categories.length > 0` while the overview gates on `isEnabled && categories.length > 0`.

**4. `clients/package.json`** — the `motion` reorder is unrelated churn.

The bulk print is the part that could most easily have gone wrong and did not. Fetching feedback items only on demand keeps the overview cheap; the `printRequests` counter with its comment explaining why a latched boolean would ignore the second click is the sort of reasoning that saves the next person an hour; filtering out participants with no scores rather than printing empty reports is the right default; and `break-before-page` on all but the first is the correct primitive. The e2e coverage is solid too — stubbing `window.print` into a console message to make an un-scriptable native dialog observable, asserting the reports are absent from the DOM before the click and that exactly one appears after, and adding `deleteOwnEvaluations` because evaluations have no admin delete endpoint so the reset would otherwise fail on the category foreign key.

`getAverageScoreLevel` widening from `Evaluation[]` to `CompetencyScore[]` is the right direction for a helper that two different score shapes now use. Approving.

PR #2016, opened 2026-08-06
`Assessment`: Allow running a phase with evaluations only

### LINKED ISSUE #1410: Allow to deactivate Assessment in Assessment to only do evaluations

rappm:
## Summary

Lets a lecturer switch off the assessment inside an assessment phase, so the phase collects only self, peer and tutor evaluations. Phases that keep the assessment behave exactly as before.

## Details

`course_phase_config` already modelled three optional sub-features with an identical `*_evaluation_enabled` / `_schema` / `_start` / `_deadline` shape. The main assessment was the only one without an enabled flag, so this mostly finishes an existing pattern rather than adding a new one.

**Configuration**

- New `assessment_enabled` column (migration `0026`), defaulting to `TRUE` so existing phases are untouched.
- The value is resolved in the service instead of with `COALESCE` in SQL: for a workflow-mode flag, a client that omits the field has to preserve the stored value rather than silently re-enable the phase.
- Disabling is refused while tutor-facing assessment data exists, checked inside the write transaction. Blank category comments and action items are excluded from that predicate, since the UI creates empty action items on click and category comments have no delete route.

**Settings**

- `SchemaConfigurationCard` already had the toggle machinery and explicitly opted the assessment out of it, so dropping that special case is most of the change. `AssessmentSettingsCard` stops duplicating the shared card's markup and passes its visibility switches as children, which are already gated on the card being active.
- The toggle inherits the card's existing lock, so it greys out once assessment data exists instead of hiding grades that were already given.
- Release results moves into its own card: it applies to evaluation-only phases too, and it used to sit inside the block that now hides.

**Lecturer views**

The participants table stays (it is how evaluation completion is tracked) but drops the score and grade columns, the assessment diagrams and the click-through into the grading form. Statistics, the per-participant assessment page and the schema detail page explain that the assessment is disabled. The `StudentDetail` remote is guarded too, since core renders it in the cross-phase student view.

**Students**

A new `GET /evaluation/my-results` gives an evaluation-only phase something to return on release: the student's own self-evaluation next to the anonymized averages their peers gave them, printable through the existing PDF button. Self and peer results render as separate sections because the two types can use different schemas.

A competency rated by fewer than two distinct peers is dropped entirely, because with a single rater the "average" is that named teammate's exact score.

**Bugs found on the way**

- `CreateOrUpdateCoursePhaseConfig` read the `Begin` error in its "config exists" check because of a shadowed `err`. Harmless for `results_released` (the zero value matched the intended default) but it would have inverted the new flag on create.
- The write guard originally went through `GetCoursePhaseConfig`, which lazily creates a config row. A guard must not write as a side effect of a request it is about to reject, and it turned malformed requests for unknown phases into 500s. It now reads the row directly and treats a missing row as the default.
- `useReleaseAssessmentResults` gates release on every assessment being final, which would have made an evaluation-only phase permanently unreleasable.
- A disabled React Query reports `isPending` forever, so switching the assessment queries off would have left `AssessmentDataShell` stuck on its loading page.
- `deleteAssessmentCompletion` existed as a network mutation that nothing called. Without it, one accidental finalized grade locks a phase out of evaluation-only mode for good, since per-item deletes are blocked by the completion and unmarking is blocked after the deadline. It is now wired to a control on the completion card.

## Reason / Link to issue

Closes #1410

## How to Test

1. `make db`, then `make server` and `make clients`.
2. On an existing assessment phase, confirm settings, participants, statistics and the student results page are unchanged. Assessment stays enabled by default.
3. On a fresh phase, turn **Assessment enabled** off in `Settings`. The schema, timeframe and visibility controls disappear, the release card stays and is clickable, participants loses its assessment columns, diagrams and row clicks, and statistics, `/participants/:id` and the schema detail page show the disabled notice.
4. As a student, submit self and peer evaluations, release results as lecturer, then open the results page: own self-evaluation plus peer averages, with a competency rated by only one peer absent. Check the PDF/print output.
5. Add an empty action item and an empty category comment, then disable: this must succeed. Type a real comment and disable again: 409 with the toggle locked.
6. Finalize an assessment, confirm disabling is blocked, use **Delete Summary** on the assessment page, then disable successfully.
7. `curl -X POST` to `student-assessment` on a disabled phase returns 409.
8. Copy the phase in the Course Configurator and confirm `assessment_enabled` carries over.

Server tests run against real Postgres via testcontainers:

```bash
cd servers/assessment && go test ./...
```

## Screenshots

Not attached yet. The automated checks below all pass, but the UI flows in steps 2 to 6 have not been exercised manually in a browser, so screenshots and a manual pass are still outstanding.

## PR Checklist

### Checks run

- `servers/assessment`: full suite green, including new coverage for the flag round-trip, omitted-field preservation, all four blocking data classes, the blank-artifact exemption, and the peer-average anonymity floor.
- `make lint-servers` (go vet, all services), `gofmt`, biome and `tsc --noEmit`: clean.
- `make test` across all servers is green except a pre-existing failure in `servers/self_team_allocation/team` (3 tests, all 500s), which is untouched by this branch and reproduces on `main`.

## Summary by CodeRabbit

* **New Features**
  * Added an Assessment Enabled setting for course phases, supporting evaluation-only phases.
  * Added student-facing self-evaluation and anonymized peer-evaluation results with print-ready reports.
  * Added controls to release or withdraw evaluation results.
  * Added the ability to delete completed assessment summaries with confirmation.
* **Improvements**
  * Assessment pages, statistics, settings, and participant views now adapt when assessments are disabled.
  * Added clearer evaluation-only messaging and guidance.
  * Prevented disabling assessments when existing assessment data would be affected.
* **Documentation**
  * Documented evaluation-only phases, result release behavior, and assessment settings.


clients/assessment_component/src/assessment/pages/SettingsPage/components/AssessmentSettingsCard/AssessmentSettingsCard.tsx
```diff
@@ -29,180 +11,60 @@ export const AssessmentSettingsCard = () => {
...
+            checked={assessmentVisibility.evaluationResultsVisible}
+            onCheckedChange={assessmentVisibility.setEvaluationResultsVisible}
+            disabled={isSaving}
+            title='Show evaluation results before submission'
+            description='Assessment authors can review self-, peer-, and student-to-tutor evaluation results before they finalize the assessment.'
+          />
         </div>
       </div>
-    </Card>
+    </SchemaConfigurationCard>
```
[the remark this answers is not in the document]
rappm: Not applying this one: `ReleaseResultsSection` was moved out of `AssessmentSettingsCard` and now renders in `SettingsPage.tsx` under its own "Results" card, precisely so evaluation-only phases keep the release controls. Adding it back here as a sibling would render it twice.

servers/assessment/coursePhaseConfig/service.go
```diff
@@ -171,11 +177,30 @@ func CreateOrUpdateCoursePhaseConfig(ctx context.Context, coursePhaseID uuid.UUI
...
+	if !assessmentEnabled {
+		hasData, err := qtx.PhaseHasAssessmentData(ctx, coursePhaseID)
+		if err != nil {
+			log.WithError(err).Error("Failed to check for existing assessment data")
+			return err
+		}
+		if hasData.Bool {
+			return ErrCannotDisableAssessmentWithData
+		}
+	}
```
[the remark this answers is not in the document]
rappm: Partly addressed. The data check now runs *after* the config write inside the same transaction, so its snapshot also sees any assessment write that committed while the disable request was in flight; the phase then stays enabled via rollback. That covers the realistic interleaving, since assessment writes are single autocommit statements while the disable request carries handler latency.

It does not fully close the window: a write that begins before the recheck's snapshot and commits after it is still invisible. Closing that needs the writers to take part in a locking protocol, which in practice means a DB-level trigger on the four data tables. I left that out deliberately: the assessment test databases are built from the dumps in `database_dumps/` rather than from migrations, so a trigger would ship untested, and it would turn a clean 409 into a 500. The residual outcome is also benign and reversible (data hidden, never lost, and re-enabling restores it). Happy to add the trigger if you'd rather have the hard guarantee.

clients/assessment_component/src/assessment/pages/hooks/useGetAllAssessmentCompletions.ts
```diff
@@ -4,13 +4,23 @@ import { useParams } from 'react-router-dom'
...
+  const query = useQuery<AssessmentCompletion[]>({
     queryKey: ['assessmentCompletions', id],
     queryFn: () => getAllAssessmentCompletionsInPhase(id ?? ''),
-    enabled: !!id,
+    enabled,
   })
+
+  // A disabled query stays pending forever, which would hang the loading gates above it
+  if (!enabled) {
+    return { ...query, data: EMPTY_ASSESSMENT_COMPLETIONS, isPending: false, isError: false }
```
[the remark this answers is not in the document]
rappm: Fixed the hook: while disabled it now returns an inert `refetch`, so the shared retry handler on the participants and statistics pages can no longer reach `getAllAssessmentCompletionsInPhase`. Same treatment applied to `useGetAllAssessments`, `useGetAllScoreLevels`, `useGetStudentAssessment` and `useGetMyEvaluationResults`.

No test for it: `assessment_component` has no client-side test setup at all, so this would mean introducing one.

mathildeshagl (changes requested):
The design is right: `course_phase_config` already modelled three optional sub-features with the same shape, and this finishes the pattern rather than inventing one. The server side is careful and well covered. Requesting changes on the client side, which the description itself flags as unexercised, plus two consistency issues in the new code.

**1. The client half of a 69-file change has never been opened in a browser and has no tests.** The description is admirably direct about this ("the UI flows in steps 2 to 6 have not been exercised manually in a browser"), and I would rather not merge past it. This is 34 client files: a new results page with its own print report, a settings card restructure, the release card moving out of a block that now hides, four lecturer views gaining disabled states, and five separate loading-state changes. None of it has a client test, and `e2e/tests/assessment/` gains no evaluation-only spec. The Go changes are covered well enough that I am not worried about them; the client changes are covered by nothing. A single e2e run through step 3 (disable, confirm the schema and timeframe controls go, confirm participants drops its columns, confirm the release card still works) would catch most of what could be wrong here.

**2. `evaluationResults.go` / `assessments/service.go:313` — the anonymity floor applies to only one of the two paths that show a student peer averages.** `GetStudentEvaluationResults` passes `minPeerRaters` (2); `GetStudentAssessmentResults` still passes `1`. So the same student, looking at the same peer scores, sees a single-rater competency on an assessment-enabled phase and not on an evaluation-only one. The reasoning in the comment — with one rater the average is that named teammate's exact score — does not depend on whether the assessment is enabled. Either both paths take the floor, or the comment should explain why the assessment path is exempt (results being lecturer-mediated there would be a reasonable answer, but it needs saying).

**3. `evaluations/router.go` (`getMyEvaluationResults`) — a student GET can create a config row.** It calls `coursePhaseConfig.GetCoursePhaseConfig`, which lazily inserts defaults on `sql.ErrNoRows` (`coursePhaseConfig/service.go:75-98`). This is the write-on-read that the PR correctly identifies and avoids in `RequireAssessmentEnabled`, with the reasoning spelled out in a comment right there — but the new student-facing endpoint reintroduces it. Reading the row directly and treating a missing row as the default, exactly as the guard does, would make the two consistent. (`getMyAssessmentResults` has the same shape on main, so this is a pre-existing pattern rather than something you introduced — but it is new code repeating it.)

**4. `coursePhaseConfig/service.go` — the post-write guard narrows the race but the comment reads as if it closes it.** "Checked after the write so the snapshot also covers writes that `RequireAssessmentEnabled` waved through while this request was in flight" holds only for writes that committed before the `EXISTS` runs. The config UPDATE takes no lock on `assessment`, `assessment_completion`, `category_assessment` or `action_item`, so a concurrent assessment write that commits after the snapshot still lands on a phase that is now disabled. I think the trade-off is fine — the alternative is locking four tables to flip a boolean — but the comment should describe the window it actually leaves rather than implying there is none.

**5. Two different idioms for the disabled-query problem, in one PR.** `AssessmentDataShell` guards inline (`assessmentEnabled && isCategoriesPending`), which is clean and needs no shim. `useGetStudentAssessment` instead returns a hand-built object overriding `data`, `isPending`, `isError` and `refetch`, while leaving `status`, `fetchStatus`, `isSuccess` and `error` as the disabled query's — so the returned object contradicts itself, and the next consumer that reads `isSuccess` or switches on `status` gets a value that disagrees with `isPending`. React Query's `isLoading` (`isPending && isFetching`) is already false for a disabled query and would do this without the shim; failing that, the inline guard used elsewhere in this same PR is the safer pattern.

**6. `db/query/course_phase_config.sql`** — `assessment_enabled` resolves in Go while every sibling flag resolves with `COALESCE(...)` in SQL, inside the same statement. Your reasoning is right and I would not change the behavior, but the `.sql` file is where the next person will look, and right now it just reads as an inconsistency. One line of comment above the upsert would carry it.

Things I checked and am happy with:

- Migration `0026` is the next free number, and `ADD COLUMN IF NOT EXISTS ... NOT NULL DEFAULT TRUE` is safe on existing rows, with a matching down. Existing phases genuinely stay untouched.
- `RequireAssessmentEnabled` is applied to every assessment mutation — `student-assessment` POST and DELETE, completion POST/PUT/mark-complete/unmark/delete, action item POST/PUT/DELETE, category assessment POST — and to no read route. That is the correct cut.
- `ErrCannotDisableAssessmentWithData` maps to 409 alongside the existing schema-change error.
- The blank-artifact exemption really does what its comment says: `!~ '^[[:space:]]*$'` covers tabs and newlines, which `btrim`'s default would not.
- `configExists` is a genuine bug fix, not just a readability change — the later `err == nil` was testing a reassigned `err`, and it would have inverted the new flag on create.
- `AggregateEvaluations` is tested for the type filter, the distinct-author floor, and that duplicate rows from one author do not satisfy it. That last test is the one that matters and it is there.
- `useDeleteAssessmentCompletion` being wired up is a good catch: without it one accidental finalized grade would lock a phase out of evaluation-only mode permanently.

clients/assessment_component/src/assessment/pages/EvaluationOverviewPage/components/EvaluationPrintReport.tsx
```diff
@@ -0,0 +1,60 @@
+import { ScoreChip } from '../../components/AssessmentPrintReport/ScoreChip'
```
mathildeshagl: This path no longer exists. #2014 renamed `pages/components/AssessmentPrintReport/` to `pages/components/PrintReport/`, so this import is what turns `quality-clients` red for every component:

```
EvaluationPrintReport.tsx(1,27): error TS2307: Cannot find module
  '../../components/AssessmentPrintReport/ScoreChip'
```

Worth more than a path fix, though. The same PR added a generic `PrintReport` (`title`, `subtitle`, `meta`, `categories`, `scores`, `categoryComments`, `feedbackItems`) that `EvaluationResultsPage` already uses on `main`, and this component hand-rolls the same header plus per-category plus `ScoreChip` layout beside it. Feeding the shared component would keep the two print outputs from drifting apart the next time either is restyled.

clients/assessment_component/src/assessment/pages/EvaluationOverviewPage/components/EvaluationResultsSection.tsx
```diff
@@ -0,0 +1,187 @@
+import { mapNumberToScoreLevel, type ScoreLevel, useCourseStore } from '@tumaet/prompt-shared-state'
+import { Card, CardContent, ErrorPage } from '@tumaet/prompt-ui-components'
+import { Loader2 } from 'lucide-react'
+import { useEffect, useMemo } from 'react'
+import { useParams } from 'react-router-dom'
+
+import { AssessmentType } from '../../../interfaces/assessmentType'
+import type { CategoryWithCompetencies } from '../../../interfaces/category'
+import { ScoreChip } from '../../components/AssessmentPrintReport/ScoreChip'
```
mathildeshagl: Same stale path as in `EvaluationPrintReport.tsx` — `ScoreChip` now lives under `pages/components/PrintReport/`.

servers/assessment/coursePhaseConfig/service.go
```diff
@@ -108,15 +112,36 @@ func GetCoursePhaseConfig(ctx context.Context, coursePhaseID uuid.UUID) (courseP
 	return coursePhaseConfigDTO.MapDBCoursePhaseConfigToDTOCoursePhaseConfig(config), nil
 }
 
+// GetStoredCoursePhaseConfig is the read-only counterpart of GetCoursePhaseConfig: it never creates
+// a row, so read paths stay reads. An unconfigured phase carries the column defaults, which is what
```
mathildeshagl: Non-blocking, but the comment promises more than the code delivers. An unconfigured phase does not come back with the column defaults — only `AssessmentEnabled` is set. `evaluation_results_visible` (0017), `grade_suggestion_visible` and `action_items_visible` (0018) all default `TRUE` in SQL and are returned as Go `false` here.

It is harmless today: both callers return 204 on `!ResultsReleased` before reading any of those, and a phase with released results always has a row. But the next caller will read this comment and trust the whole struct. Either set the three flags or narrow the comment to the one field it actually covers.

mathildeshagl (changes requested):
Five of the six points from my last review are addressed, and well:

- **Peer-rater floor** (`e81640f6`) — exported `MinPeerRaters` and applied on both student-facing paths, with a comment saying why self stays at 1. That was the right resolution.
- **Write-on-read** (`6973598c`) — `GetStoredCoursePhaseConfig` shared by both student endpoints and the guard, with a test that asserts `COUNT(*) = 0` after the read. Exactly the shape I was after.
- **Race comment** (`3472adb`/follow-up) — now states the window it leaves rather than implying there is none.
- **Two idioms** (`d354b2f9`) — the self-contradicting override objects are gone from all five hooks; they only default `data` now and the pages guard on `assessmentEnabled` inline, matching `AssessmentDataShell`.
- **SQL comment** (`0dfcadcf`).
- **Client coverage** (`451f16be`) — `evaluation-only.spec.ts` is a good spec. It walks disable, persistence, participant columns, the statistics notice, release, and the student view, and checks the API side (`my-results` 204 then 200) alongside the UI. That answers my main objection.

Requesting changes on one thing that is new since my last pass, plus the reuse question it exposes.

**The branch no longer compiles against `main`.** `quality-clients` fails for every component with `TS2307: Cannot find module '../../components/AssessmentPrintReport/ScoreChip'`. #2014 renamed that directory to `pages/components/PrintReport/`, and the two new files still import the old path. GitHub reports the branch as mergeable because the conflict is semantic rather than textual, which makes it easy to miss. The knock-on is that `e2e-tests` is skipped, so **the new evaluation-only spec has never actually run** — the coverage that resolves point 1 is currently unverified.

While fixing the path, please look at whether `EvaluationPrintReport` should exist at all: #2014 also landed a generic `PrintReport` that takes title, subtitle, meta, categories, scores, category comments and feedback items, and `EvaluationResultsPage` on `main` already drives it. Two print layouts for the same phase will drift on the next styling change.

Happy to re-review as soon as CI is green — everything else here I am satisfied with.

clients/assessment_component/src/assessment/pages/EvaluationOverviewPage/components/EvaluationPrintReport.tsx
rappm: Merged `main` and reworked this: `EvaluationPrintReport` is now a thin wrapper that feeds the shared `PrintReport` (one report for the self-evaluation, one for the averaged peer feedback), so the header/category/`ScoreChip` layout has a single owner again. The categories are pre-filtered to the scored competencies so peer competencies rated by a single peer still stay out rather than showing as "Not assessed".

clients/assessment_component/src/assessment/pages/EvaluationOverviewPage/components/EvaluationResultsSection.tsx
rappm: Fixed, now imports from `../../components/PrintReport/ScoreChip`. `yarn tsc -p assessment_component/tsconfig.json --noEmit` is clean.

servers/assessment/coursePhaseConfig/service.go
rappm: Good catch. Took the first option: the fallback now sets `EvaluationResultsVisible`, `GradeSuggestionVisible` and `ActionItemsVisible` alongside `AssessmentEnabled`, so the struct really does carry the boolean column defaults. The comment now also says what it does not cover (the timestamps stay zero, since no row exists to supply them). `TestGetStoredCoursePhaseConfigDoesNotCreateARow` asserts all four flags plus the two `FALSE` ones.

PR #2020, opened 2026-08-07
Audit Log: Add centralized audit log (store, API, UI)

Mtze:
## Summary

Adds a robust, centralized audit log that serves both as a human-readable **activity feed** for instructors/admins and a **compliance/security trail** (grade/role/deletion accountability, denied-access attempts). Actions are captured automatically with no per-route work; the log is searchable and filterable, scoped to a course for lecturers and platform-wide for admins.

> **Depends on** prompt-edu/prompt-sdk#111 (the shared `audit` package). Until that is released, `servers/core/go.mod` uses a local `replace` directive (marked `TODO(audit)`) — CI will not resolve it until the SDK is released and the version bumped. This is expected for the draft.

## Details

**Backend (`servers/core`)**
- Migration `0026_audit_log`: append-only table with keyset-friendly indexes and a `BEFORE UPDATE` trigger rejecting mutations (no domain FKs, so the trail outlives referenced rows). sqlc queries for create, keyset-paginated filtered list, and retention delete.
- `auditLog` module consuming `prompt-sdk/audit`: a DB sink + core actor extractor, the auto-capture middleware wired **first** in `main.go` so it wraps every route, read endpoints (`GET /api/audit-log` admin-wide, `GET /api/courses/:uuid/audit-log` per-course), a shared-secret ingest endpoint (`POST /api/audit`) for phases, a transactional `RecordTx`, and a daily retention pruner (`AUDIT_RETENTION_DAYS`, unset ⇒ never prune).
- Feature toggle: everything is a no-op unless `AUDIT_ENABLED=true`.

**Frontend (`clients/core`)**
- Course + admin audit-log pages built on `PromptTable` with **server-side** filtering (actor role, outcome, source, date range, search) and cursor pagination via `useInfiniteQuery`; routes + sidebar entries with the existing permission gating.

**Docs / config / tests**
- User, admin, and developer guides; `.env.template` and Go-rules pointer; `AUDIT_ENABLED` set in the e2e stack.
- Core integration tests (testcontainers): DB sink + course-id resolution, `RecordTx` atomicity, anti-UPDATE trigger, retention, filtering, keyset pagination. Playwright spec for nav + access control.

## Reason / Link to issue

No audit trail exists today. Design captured in an extensive plan + grilling session.

## How to Test

1. Set `AUDIT_ENABLED=true` in `.env.dev`, then `make db && make server && make clients`.
2. As a lecturer, perform a mutating action in a course, open **Course → Audit Log**, and confirm the entry appears with actor/action/time; try the filters and search.
3. As `PROMPT_Admin`, open **Admin → Audit Log** for the platform-wide view; a non-admin gets *Access Denied*.
4. Backend: `cd servers/core && go test ./auditLog/...`

## PR Checklist

## Summary by CodeRabbit

- **New Features**
  - Added platform-wide and course-level audit logs in the management console.
  - Audit entries include actors, actions, outcomes, entities, timestamps, and source services.
  - Added filtering by actor, action, role, outcome, service, entity, and date range.
  - Added paginated loading for older entries.
  - Access is restricted to administrators and authorized course lecturers.
  - Added configurable enablement, retention, and service authentication.
  - Audit records are append-only and support secure ingestion.

- **Documentation**
  - Added administrator, user, and contributor guidance for audit logging.


servers/core/main.go
```diff
@@ -170,6 +171,10 @@ func main() {
 	initKeycloak(api, *query)
 	permissionValidation.InitValidationService(*query, conn)
 
+	// Audit logging must be initialized before the other modules so its
+	// auto-capture middleware wraps their routes.
+	auditLog.InitAuditLogModule(api, *query, conn)
```
rappm: Gin snapshots the handler chain when a route or subgroup is registered, so `api.Use(audit.Middleware(...))` in here cannot wrap anything that was registered earlier. `initKeycloak(api, *query)` runs five lines above, which means all `/keycloak/:courseID/...` routes keep the pre-audit chain.

So granting or revoking a course role (`PUT`/`DELETE /api/keycloak/:courseID/group/:groupName/members/:userID`) writes no audit entry at all, which is probably the one thing we most want in there. The comment above the call states the opposite of what happens.

Moving this call above `initKeycloak` fixes it.

servers/core/auditLog/router.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+}
+
+// ingestAuth validates the per-service shared secret on the ingest endpoint.
+func ingestAuth(keys map[string]string) gin.HandlerFunc {
+	return func(c *gin.Context) {
+		service := c.GetHeader("X-Audit-Service")
+		token := c.GetHeader("X-Audit-Token")
+		expected, ok := keys[service]
+		if service == "" || !ok || subtle.ConstantTimeCompare([]byte(token), []byte(expected)) != 1 {
```
rappm: `subtle.ConstantTimeCompare([]byte(""), []byte(""))` returns 1, so a service configured with an empty key authenticates every request that sends its name.

`AUDIT_INGEST_KEYS=interview:` or `AUDIT_INGEST_KEYS=interview:${AUDIT_INGEST_KEY}` with `AUDIT_INGEST_KEY` unset (both variables ship blank in `.env.template`) is enough. Anyone can then `POST /api/audit` with `X-Audit-Service: interview` and no token.

Please skip empty keys in `parseIngestKeys` and reject an empty `expected` here.

servers/core/auditLog/router.go
```diff
@@ -0,0 +1,164 @@
...
+// is taken from the (trusted) request body; source_service is stamped from the
+// authenticated service identity, never the body.
+func ingestAuditEvent(sink *DBSink) gin.HandlerFunc {
+	return func(c *gin.Context) {
+		var e audit.Event
+		if err := c.ShouldBindJSON(&e); err != nil {
+			c.JSON(http.StatusBadRequest, utils.ErrorResponse{Error: err.Error()})
+			return
+		}
+		e.SourceService = c.GetString(ingestServiceContextKey)
```
rappm, line 128: Only `SourceService` is stamped from the authenticated identity. Actor, course, action and outcome come straight from the body, so one leaked phase key is enough to write entries that look like a `PromptAdmin` acting in any course. The append-only trigger then makes them impossible to correct.

At minimum the ingesting service should only be allowed to write events for its own phases, and `outcome` / `sourceService` should be validated instead of taken as given.

servers/core/auditLog/sink.go
```diff
@@ -0,0 +1,144 @@
...
+// db.Queries (via pointer) and a transaction-bound *db.Queries.
+type courseResolver interface {
+	GetCourseIDByCoursePhaseID(ctx context.Context, id uuid.UUID) (uuid.UUID, error)
+}
+
+// eventToParams maps a neutral audit.Event onto the sqlc insert params,
+// resolving course_id from the course phase when it is not already set.
+func eventToParams(ctx context.Context, r courseResolver, e audit.Event) db.CreateAuditEntryParams {
+	courseID := pgUUID(e.CourseID)
+	if !courseID.Valid && e.CoursePhaseID != "" {
```
rappm, line 47: `course_id` is only backfilled from the course phase, but the SDK fills `CourseID` from the `courseId` / `courseID` params, and core's course routes are registered with `:uuid`. For `PUT /api/courses/:uuid`, `DELETE /api/courses/:uuid`, `PUT /api/courses/:uuid/archive`, `PUT /api/courses/:uuid/phase_graph` and `POST /api/courses/:uuid/participations` both `CourseID` and `CoursePhaseID` are empty, so the row is inserted with `course_id = NULL`.

`ListAuditLog` filters on `course_id = $1`, so a lecturer opening their course log does not see a course being archived, renamed or deleted. Needs a fallback on the `uuid` param (which means passing the gin context in here).

.env.template
```diff
@@ -21,6 +21,19 @@ CHAIR_NAME_LONG=TUM Research Group for Applied Education Technologies
 # Server configuration
 SERVER_ADDRESS=0.0.0.0:8080
 
+# ============================================================================
+# AUDIT LOG (optional; off by default)
+# ============================================================================
+# Enable the centralized audit log. When unset/false, nothing is captured.
+AUDIT_ENABLED=
```
rappm, line 28: These are documented as core settings in `docs/admin/audit_log.md`, but neither `docker-compose.yml` nor `docker-compose.prod.yml` passes them into `server-core` (both list the environment explicitly). Only `docker-compose.e2e.yml` was updated.

So setting `AUDIT_ENABLED=true` in `.env` does nothing: `InitAuditLogModule` returns early, no routes are registered, and the sidebar entry (which is rendered unconditionally) leads to "Failed to load the audit log.". Please add the three variables to both compose files.

servers/core/auditLog/router.go
```diff
@@ -0,0 +1,164 @@
...
+	for _, pair := range strings.Split(raw, ",") {
+		pair = strings.TrimSpace(pair)
+		if pair == "" {
+			continue
+		}
+		service, key, ok := strings.Cut(pair, ":")
+		if !ok {
+			continue
+		}
+		keys[strings.TrimSpace(service)] = strings.TrimSpace(key)
```
rappm: A `map[string]string` keeps one key per service, so `interview:oldkey,interview:newkey` ends up as `{"interview": "newkey"}` and the old key is gone as soon as core restarts.

That breaks the rotation procedure in `docs/admin/audit_log.md` ("Two values may be listed per service"): until the phase is redeployed its events get a 401, and the SDK sink only logs the failure, so they are dropped silently. Should be `map[string][]string` with a compare against each accepted key.

servers/core/auditLog/service.go
```diff
@@ -0,0 +1,150 @@
...
+		log.Info("audit retention disabled (AUDIT_RETENTION_DAYS not set); entries are never pruned")
+		return
+	}
+	days, err := strconv.Atoi(raw)
+	if err != nil || days <= 0 {
+		log.WithField("AUDIT_RETENTION_DAYS", raw).Warn("invalid AUDIT_RETENTION_DAYS; audit retention disabled")
+		return
+	}
+
+	pruneOnce(ctx, days)
```
rappm: The first prune runs synchronously, and `StartRetentionPruner` is called from `main()` before `router.Run`. On the first deploy where `AUDIT_RETENTION_DAYS` is set on an instance that has been collecting entries, this is one unbatched `DELETE` over the whole backlog, and core does not bind its port until it finishes. The startup probe kills the container and the prune starts over.

Please move it inside the goroutine.

clients/core/src/managementConsole/auditLog/components/AuditLogFilterBar.tsx
```diff
@@ -0,0 +1,99 @@
...
+  const update = (partial: Partial<AuditLogFilters>) => onFiltersChange({ ...filters, ...partial })
+
+  const dateRange: DateRange | undefined = filters.from
+    ? { from: new Date(filters.from), to: filters.to ? new Date(filters.to) : undefined }
+    : undefined
+
+  const onDateChange = (range: DateRange | undefined) =>
+    update({
+      from: range?.from ? range.from.toISOString() : undefined,
+      to: range?.to ? range.to.toISOString() : undefined,
```
rappm: `range.to` is local midnight at the start of the selected day, and the backend applies `created_at <= to_time`, so the selected end day is excluded entirely. Picking "Aug 1 - Aug 7" shows nothing from Aug 7.

The `from` bound has the same problem in the other direction: local midnight converted to UTC pulls in entries from the previous evening. Please send `startOfDay(range.from)` / `endOfDay(range.to)`.

servers/core/db/query/audit_log.sql
```diff
@@ -0,0 +1,37 @@
...
+  AND (sqlc.narg('actor_role')::text IS NULL OR actor_role = sqlc.narg('actor_role'))
+  AND (sqlc.narg('outcome')::text IS NULL OR outcome = sqlc.narg('outcome'))
+  AND (sqlc.narg('action_key')::text IS NULL OR action_key = sqlc.narg('action_key'))
+  AND (sqlc.narg('entity_type')::text IS NULL OR entity_type = sqlc.narg('entity_type'))
+  AND (sqlc.narg('source_service')::text IS NULL OR source_service = sqlc.narg('source_service'))
+  AND (sqlc.narg('course_phase_id')::uuid IS NULL OR course_phase_id = sqlc.narg('course_phase_id'))
+  AND (sqlc.narg('from_time')::timestamptz IS NULL OR created_at >= sqlc.narg('from_time'))
+  AND (sqlc.narg('to_time')::timestamptz IS NULL OR created_at <= sqlc.narg('to_time'))
+  AND (sqlc.narg('search')::text IS NULL OR (
+        actor_name ILIKE '%' || sqlc.narg('search') || '%' OR
```
rappm: The search string goes into the pattern unescaped, so `%` and `_` from user input act as wildcards. Searching for `_` or `%` matches every row, and `student_id` also matches `studentXid`.

Please escape `%`, `_` and the backslash in `parseListFilters` and add an explicit `ESCAPE` clause.

servers/core/auditLog/router.go
```diff
@@ -0,0 +1,164 @@
...
+		if t, err := time.Parse(time.RFC3339, v); err == nil {
+			f.From = &t
+		}
+	}
+	if v := c.Query("to"); v != "" {
+		if t, err := time.Parse(time.RFC3339, v); err == nil {
+			f.To = &t
+		}
+	}
+	if v := c.Query("cursorTs"); v != "" {
```
rappm: All three time parses drop the value when it fails to parse, with no error branch.

For `from` / `to` that means a malformed date returns the full unfiltered log with a 200. For `cursorTs` it is worse: the server ignores the cursor, returns the first page again with the same `nextCursor`, and `useInfiniteQuery` keeps appending the same 50 rows on every `Load more` (also duplicate keys in the table). A 400 on an unparseable value would be better.

clients/core/src/managementConsole/auditLog/AuditLogView.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+      <AuditLogFilterBar filters={filters} onFiltersChange={onFiltersChange} />
+
+      {isError ? (
+        <ErrorPage message='Failed to load the audit log.' />
+      ) : isLoading ? (
+        <LoadingPage />
+      ) : (
+        <>
+          <PromptTable data={entries} columns={columns} />
```
rappm: `PromptTable` paginates client-side with a default page size of 100, while the server returns 50 per page. After the second `Load more` the table still shows only the first 100 rows and moves the rest onto its own second page, so the button fires a request and nothing visibly changes.

Sorting and the table's own filter also only apply to what has been loaded so far. Either pass a page size that matches and hide the internal pager, or load everything for the current filter and drop `Load more`.

servers/core/auditLog/sink.go
```diff
@@ -0,0 +1,144 @@
...
+		return permissionValidation.PromptLecturer
+	}
+	for _, courseRole := range []string{permissionValidation.CourseLecturer, permissionValidation.CourseEditor, permissionValidation.CourseStudent} {
+		for userRole := range roles {
+			if strings.HasSuffix(userRole, courseRole) {
+				return courseRole
+			}
+		}
+	}
+	for r := range roles {
```
rappm, line 158: Ranging over a map and returning the first element gives a random role, since Go randomizes map iteration order. The same user can end up with a different `actor_role` on every request, so the Role column and the role filter disagree with each other.

This is reachable for anyone whose roles carry none of the known suffixes (e.g. a custom course group), including their denied 403 attempts. Sorting the roles first, or returning `""`, would at least be stable. The `roles` slice built above has the same issue.

clients/core/src/managementConsole/auditLog/components/AuditLogFilterBar.tsx
```diff
@@ -0,0 +1,99 @@
...
+    })
+
+  return (
+    <div className='flex flex-wrap items-center gap-3'>
+      <Input
+        placeholder='Search actor, action, entity…'
+        value={filters.search ?? ''}
+        onChange={(e) => update({ search: e.target.value || undefined })}
```
rappm: Every keystroke updates `filters`, which is part of the query key, so typing an eight-letter name fires eight requests and flips the view to `LoadingPage` each time.

Each of those runs five `ILIKE '%...%'` predicates over `audit_log`, which the leading wildcard makes a full scan, on a table that is never pruned by default. Please debounce the value before it reaches the query key.

servers/core/db/migration/0026_audit_log.up.sql
```diff
@@ -0,0 +1,46 @@
+-- Migration: Create the central audit log
+
+CREATE TABLE audit_log (
+  id              uuid PRIMARY KEY DEFAULT gen_random_uuid(),
+  created_at      timestamptz NOT NULL DEFAULT now(),
+  actor_id        uuid,                              -- always a human in practice; nullable only for the maintenance edge
+  actor_name      text NOT NULL DEFAULT '',
+  actor_email     text NOT NULL DEFAULT '',
```
rappm, line 8: `actor_name`, `actor_email` and `entity_name` are personal data kept indefinitely (retention is off by default), but `ExecuteCoreDeletion` was not extended to cover this table, so a deletion request leaves them in place and searchable by any admin.

The append-only trigger also blocks the anonymize-by-UPDATE approach used for notes and files, so this cannot be added later without a schema change. Worth deciding now whether the trigger should allow an anonymizing update, or whether the deletion pipeline should scrub these rows.

servers/core/db/query/audit_log.sql
```diff
@@ -0,0 +1,37 @@
...
+        actor_email ILIKE '%' || sqlc.narg('search') || '%' OR
+        action ILIKE '%' || sqlc.narg('search') || '%' OR
+        entity_name ILIKE '%' || sqlc.narg('search') || '%' OR
+        entity_id ILIKE '%' || sqlc.narg('search') || '%'))
+  AND (sqlc.narg('cursor_ts')::timestamptz IS NULL OR
```
rappm, line 33: Written as an `OR` chain, Postgres cannot turn this into an index start bound on `(created_at DESC, id DESC)`, so every page rescans and discards the rows already returned and `Load more` gets slower the further back you go.

The row-wise form `(created_at, id) < (cursor_ts, cursor_id)` uses the index the migration adds for exactly this.

rappm (changes requested):
Nice structure overall, but a few things need to be fixed before this can go in.

Blocking:
- `InitAuditLogModule` runs after `initKeycloak`, so the Keycloak role routes are never captured.
- Core's course routes use `:uuid`, so `course_id` stays NULL and course lecturers see almost nothing in their course log.
- The ingest endpoint accepts an empty configured key and trusts the actor from the request body.
- The new `AUDIT_*` variables are not passed to `server-core` in `docker-compose.yml` / `docker-compose.prod.yml`, so the feature cannot be turned on.

The remaining points are inline: key rotation, the blocking prune at startup, the date range end bound, `ILIKE` escaping, silently dropped filter params, table pagination, search debouncing, role determinism, the cursor predicate, and audit PII vs. the privacy deletion pipeline.

servers/core/main.go
Mtze: Fixed - split into `InitAuditLogCapture` (registered before `initKeycloak`) and `InitAuditLogRoutes` (after `permissionValidation`), so the auto-capture middleware now wraps the Keycloak course-role routes as well.

servers/core/auditLog/router.go
Mtze: Fixed - `parseIngestKeys` now drops empty keys and `matchesAnyKey` cannot match an empty token, so a blank configured key no longer authenticates every request.

servers/core/auditLog/router.go
Mtze: By design: only trusted internal services hold a per-service key, and `source_service` is stamped from the matched key (never the body). The residual risk of a leaked key forging entries for that one service is accepted for an append-only log, and a per-phase write restriction belongs to the phase rollout. Added `outcome` enum validation on ingest as cheap hardening.

servers/core/auditLog/sink.go
Mtze: Fixed - the sink backfills `course_id` from the `:uuid` entity id on `/api/courses/:uuid` routes (scoped to that path, since `:uuid` maps to other entity types elsewhere), so course archive/rename/delete/participation events are now course-scoped.

.env.template
Mtze: Fixed - added `AUDIT_ENABLED`/`AUDIT_RETENTION_DAYS`/`AUDIT_INGEST_KEYS` to `server-core` in both `docker-compose.yml` and `docker-compose.prod.yml`.

servers/core/auditLog/router.go
Mtze: Fixed - keys are now `map[string][]string` and a request authenticates against any accepted key, so old and new keys coexist during rotation.

servers/core/auditLog/service.go
Mtze: Fixed - the initial prune now runs inside the goroutine, off the startup path, so a large backlog delete no longer blocks the port bind.

clients/core/src/managementConsole/auditLog/components/AuditLogFilterBar.tsx
Mtze: Fixed - the client now sends `startOfDay(from)`..`endOfDay(to)`.

servers/core/db/query/audit_log.sql
Mtze: Fixed - user input is escaped and the query uses an `ESCAPE` clause, so `%`/`_` are matched literally.

servers/core/auditLog/router.go
Mtze: Fixed - unparseable `from`/`to`/`cursorTs`/`cursorId`/`limit` now return 400 instead of silently serving the full log or looping `Load more` on duplicate rows.

clients/core/src/managementConsole/auditLog/AuditLogView.tsx
Mtze: Fixed - `PromptTable` `pageSize` is set to the loaded row count so all loaded rows stay on one page and `Load more` visibly appends.

servers/core/auditLog/sink.go
Mtze: Fixed - the roles slice is sorted and the fallback role is the lexicographically smallest, in both core and the SDK, so the value is stable across requests.

clients/core/src/managementConsole/auditLog/components/AuditLogFilterBar.tsx
Mtze: Fixed - the search value is debounced (400ms) before it reaches the query key.

servers/core/db/migration/0026_audit_log.up.sql
Mtze: Intentional: audit entries are retained until the retention window expires and the table is append-only (a deliberate product decision). I reframed the admin docs to present this as an operator-configured retention exception (lawful basis + retention schedule + erasure-request handling) rather than a blanket exemption.

servers/core/db/query/audit_log.sql
Mtze: Fixed - switched to the row-wise predicate `(created_at, id) < (cursor_ts, cursor_id)` so it uses the composite index.

servers/core/auditLog/router.go
```diff
@@ -0,0 +1,164 @@
...
+		authMiddleware(),
+		permissionValidation.CheckAccessControlByRole(permissionValidation.PromptAdmin),
+		listGlobalAuditLog)
+
+	// Per-course log (course lecturers of that course, or admins).
+	api.GET("/courses/:uuid/audit-log",
+		authMiddleware(),
+		permissionValidation.CheckAccessControlByID(permissionValidation.CheckCoursePermission, "uuid",
+			permissionValidation.CourseLecturer, permissionValidation.PromptAdmin),
+		listCourseAuditLog)
```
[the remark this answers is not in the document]
Mtze: Not applied - the audit log is course-wide across all phases, so there is no single `coursePhaseID` to key it on. Core already exposes course-level routes at `/api/courses/:uuid` (course, copy, participations) using the same `CheckCoursePermission` pattern; the `course_phase` path rule targets phase services that rely on the SDK auth middleware.

servers/core/auditLog/service_test.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+// insertRaw inserts a row directly, controlling created_at (which the sink never
+// sets) — used to seed ordering/retention scenarios.
+func (s *AuditLogTestSuite) insertRaw(createdAt time.Time, courseID, actorRole, outcome, action string) {
+	_, err := s.conn.Exec(s.ctx,
+		`INSERT INTO audit_log (created_at, course_id, actor_role, outcome, action, source_service)
+		 VALUES ($1, $2, $3, $4, $5, 'core')`,
+		createdAt, pgUUID(courseID), actorRole, outcome, action)
+	require.NoError(s.T(), err)
```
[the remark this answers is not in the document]
Mtze: Not applied - the test helpers set `created_at` for the retention/keyset/ordering scenarios, which the production append-only `CreateAuditEntry` deliberately does not expose. The raw SQL is test-only setup.

Mtze:
Thanks for the thorough review @rappm - addressed in the latest push. Summary:

**Fixed**
- Register the auto-capture middleware before `initKeycloak` (split into `InitAuditLogCapture` + `InitAuditLogRoutes`), so Keycloak role routes are captured.
- Backfill `course_id` from the `:uuid` entity id on `/api/courses/:uuid` routes.
- Reject empty ingest keys; support multiple keys per service for zero-downtime rotation.
- Return 400 on unparseable date/cursor/limit params; escape ILIKE wildcards; row-wise keyset predicate.
- Move the initial retention prune off the startup path.
- Client: whole-day date bounds, debounced search, table pageSize matched to loaded rows.
- Deterministic actor role/roles (core + SDK).
- Pass `AUDIT_*` to `server-core` in both compose files.
- Docs: reframed GDPR retention section; de-sensitized the metadata example; documented audit query params in Swagger.

**Intentionally kept (rationale on the threads)**
- Ingest trusts the actor from the body (per-service keys, `source_service` stamped from the matched key; append-only residual risk accepted). Added `outcome` validation.
- Audit entries are retained until the retention window expires rather than scrubbed on privacy deletion (deliberate; docs reframed).
- Course read route stays at `/api/courses/:uuid/audit-log` (course-wide across all phases; matches core's existing course-level route pattern).

New unit tests cover the ingest-key hardening, the course_id backfill, and the search escaping. Ready for another look.

rappm (dismissed):

PR #2026, opened 2026-08-11
`Assessment`: Restrict peer and tutor evaluation targets to the author's team

rappm:
## Summary

`POST /evaluation` and `POST /evaluation/feedback-items` bind the *author* of a record to the caller's JWT, but for `type: peer` and `type: tutor` the *target* `courseParticipationID` is taken from the request body and never checked. Any enrolled student can therefore submit competency scores and written feedback about any other participant in the phase, including students on other teams. Tutor-typed records are the sharper end of this: they surface in the lecturer-facing tutor evaluation view, so a student can attribute arbitrary tutor feedback to a tutor they were never assigned to.

## Details

`CheckEvaluationIsEditable` already enforced the target invariant for `type: self` (added in #1790). This extends the same idea to the two remaining student-authored types instead of leaving them as inert data:

- `type: peer` must target a **different** member of the author's own team.
- `type: tutor` must target one of that team's **tutors**.
- An author who is not a member of any team in the phase cannot write either type.
- `type: self` and the invalid-type handling are unchanged.

Team membership comes from the existing `coursePhaseConfig.GetTeamsForCoursePhase` resolution, behind a `getTeamsForCoursePhaseFn` seam matching the pattern already used in `reminderRecipients.go`.

Routers now distinguish these rejections from internal errors: `IsTargetAuthorizationError` maps them to `403` in `createOrUpdateEvaluation`, `createOrUpdateEvaluationCompletion` and `markEvaluationAsCompleted`, instead of falling through to `500`.

**Deliberate trade-offs, worth a look during review**

- **The write path now calls the core service.** `CheckEvaluationIsEditable` previously only touched the local DB. For peer and tutor writes it now performs one `FetchAndMergeCoursePhaseWithResolution` call per write, and the client saves one request per competency, so a peer evaluation form costs N such calls. `type: self` is unaffected (no fetch). If this shows up in practice the fix is a short-lived per-request team cache, which I left out rather than guess at an invalidation policy.
- **It fails closed.** If core is unreachable the write is rejected rather than allowed through. That is the right default for an authorization check, but it does couple peer and tutor writes to core availability.
- **Delete is covered too.** `DeleteEvaluation` runs the same check with the stored row's values, so a student removed from their team can no longer delete peer evaluations they previously wrote. That is consistent with them not being able to edit those rows either, but it is a behavior change rather than a pure tightening.
- Because this sits in `CheckEvaluationIsEditable`, feedback items pick the rule up automatically once #2025 routes their create path through the same gate. The two PRs are independent and can merge in either order.

**Tests**

- `testutils.SetupMockCoreService` gains the `/api/course_phases/:id/course_phase_data` route so the resolution path can be exercised, with a fixed team fixture (`MockTeamMembers`, `MockTeamTutors`, `MockOutsiderParticipationID`) that the assessment test participants belong to.
- Added coverage for a peer target outside the team, a peer record targeting the author, a tutor record targeting a teammate rather than a tutor, and an author with no team.
- The existing `TestCreatePeerEvaluation` happy path still passes: both participants are members of the mock team.

## Reason / Link to issue

Found during a security review of the assessment service, alongside #2025. #1790 fixed the equivalent hole for `type: self` but left the peer and tutor paths gated only on the author.

Note that an earlier framing of this finding proposed gating peer/tutor writes on an assessor role. That is wrong for this codebase: tutor evaluations are authored *by students about their tutors* (see the student-facing `EvaluationOverviewPage` and `GET /evaluation/tutor/:courseParticipationID`, whose path param is the tutor). Team membership is the correct constraint.

## How to Test

1. `cd servers/assessment && go test ./evaluations/...`
2. As an enrolled student, `POST /assessment/api/course_phase/<phase>/evaluation` with your own `authorCourseParticipationID`, `"type": "peer"`, and the `courseParticipationID` of a student on another team. Expect `403` (previously `201`).
3. Repeat with `"type": "tutor"` and a peer's ID rather than your team's tutor. Expect `403`.
4. Confirm the normal flow still works: submit and complete a peer evaluation for a teammate and a tutor evaluation for your team's tutor.

## PR Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Added authorization checks when creating, editing, completing, or deleting evaluations and feedback.
  - Peer evaluations can now target only members of the author’s team.
  - Tutor evaluations require the selected participant to be the team’s tutor.
  - Self-evaluations and submissions by participants without a team are rejected.
  - Invalid or unauthorized targets now return a clear HTTP 403 response instead of allowing the operation to proceed.


mathildeshagl (changes requested):
The rule is the right one and the framing note about assessor roles versus team membership is correct. I checked it against the client: `EvaluationOverviewPage` already resolves peer and tutor targets from the same `useGetAllTeams` data, matches the author on `team.members`, and renders neither section when no team matches. So the server constraint mirrors what the UI already offers, and a phase with no team resolution was never showing these forms. Requesting changes on the auth-header plumbing and the test coverage around it.

**1. `service.go` (`authHeaderFromContext`) — the authorization check degrades silently when the context is not a `*gin.Context`.** The type assertion falls back to `""`, which reaches `FetchAndMergeCoursePhaseWithResolution` as an unauthenticated call. All four current call sites pass `c` directly, so this works today, but the failure mode is the problem: the moment anyone passes `c.Request.Context()` — the idiomatic choice, and what `db.GetTimeoutContext` produces — every peer and tutor write starts failing with a generic 500 and no indication that an auth header went missing. `GetTeamsForCoursePhase` already takes the header as an explicit parameter; threading it through `CheckEvaluationIsEditable` the same way would make the dependency compiler-checked instead of a runtime type assertion. If you would rather keep the implicit lookup, it should at least distinguish "no header available" from "core said no" so the log says which one happened.

**2. `service_test.go` — the new tests cannot catch (1).** `suite.suiteCtx` is `context.Background()`, so `authHeaderFromContext` returns `""` in every one of them, and the mock core does not check the `Authorization` header. All four tests pass identically whether the header plumbing works or is completely broken. Exercising at least one case through a `*gin.Context` carrying a header, with the mock asserting it arrived, would close that.

**3. No positive tutor-evaluation test exists anywhere in the service.** All four new tests assert rejection, and `assessmentType.Tutor` appears in no other `_test.go` file except `reminderRecipients_service_test.go`. Tutor writes are now gated on `containsPerson(team.Tutors, courseParticipationID)`, so if `parseTeam` ever stopped populating `Tutors` — a key rename in the resolution payload would do it, silently, since `parseTeam` defaults the field to an empty slice rather than failing — every tutor evaluation would 403 in production with the suite still green. A test that a tutor record targeting `MockTeamTutors[0]` is accepted is the single most valuable one to add here.

**4. `evaluations/router.go:253` — `deleteEvaluation` still maps everything to 500.** `IsTargetAuthorizationError` was only wired into `createOrUpdateEvaluation`, so the scenario the description calls out ("a student removed from their team can no longer delete peer evaluations they previously wrote") surfaces as an internal error rather than the 403 the rest of the PR is careful to produce. `ErrSelfEvaluationTargetMismatch` has the same problem on that path, though that one predates this change.

Worth coordinating with #2025 rather than fixing twice: that PR replaces this exact if-chain with an `evaluationErrorStatus(err)` switch that both `createOrUpdateEvaluation` and `deleteEvaluation` call. Landing #2025 first and adding the target-authorization case to that switch resolves (4) for free and avoids a conflict in the same function.

**5. On the performance trade-off.** Request scope needs no invalidation policy, which I think makes the cache cheaper than the description assumes: `c.Set` / `c.Get` keyed by course phase ID, populated on the first team fetch of the request, is discarded when the request ends. That turns an N-competency peer form back into one core call per save without anyone having to reason about staleness. Not blocking, but I would rather have it than a documented N-calls cost.

Smaller notes:

- `checkEvaluationTarget`'s `default: return nil` is safe only because the switch in `CheckEvaluationIsEditable` below rejects unknown types with `ErrInvalidEvaluationType`. That coupling is worth a comment, since the two switches are now 60 lines apart and read as if the first one were the complete story.
- `teamOfMember` returns the first team containing the author. Fine given a participant belongs to one team, but it silently picks a winner if that ever stops holding.

What is good: the `getTeamsForCoursePhaseFn` seam matches the `reminderRecipients.go` precedent exactly, failing closed is the right default for an authorization check and is stated as a deliberate choice, `type: self` and the invalid-type handling are genuinely untouched, and putting this in `CheckEvaluationIsEditable` rather than the router means #2025's feedback item path inherits it for free. The mock core fixture is well chosen — the existing `TestCreatePeerEvaluation` happy path keeps passing because participants 1 to 3 really are the mock team's members, rather than because the check was weakened.

rappm:
Thanks, this was a useful review. All four blocking points are addressed. On point 5 I think the premise does not hold, details below.

**1. `authHeaderFromContext` (fixed in 2a0e106).** Removed the type assertion entirely and threaded the header through `CheckEvaluationIsEditable` as an explicit parameter, as you suggested, so it matches `GetTeamsForCoursePhase` and is compiler-checked. The routers read `c.GetHeader("Authorization")` at the edge, which is the idiom already used in `coursePhaseConfig/router.go`. Since the merge brought in #2025, this now covers six call sites rather than four: the two feedback item write paths needed the same treatment, so peer and tutor feedback items resolve teams with the caller's credentials too.

**2. Tests could not catch (1) (fixed in 2a0e106).** The mock core now records the `Authorization` header of every course phase data resolution, and `TestTargetCheckForwardsAuthHeaderToCore` asserts it arrived. I verified it actually bites: passing `""` at the call site instead of the header fails the test with `expected: "Bearer ..." / actual: ""`. Previously that mutation was invisible.

**3. No positive tutor test (fixed in 2a0e106).** Added `TestCreateTutorEvaluation`, asserting a record targeting `MockTeamTutors[0]` is accepted. Two things were needed to make it possible: `tutor_evaluation_enabled` defaulted to false in `evaluations.sql` so tutor writes were closed, and the tutor needed a `course_participation` row for the FK. Worth noting for anyone writing similar tests: `GetEvaluationsForParticipantInPhase` filters `type != 'tutor'`, so the assertion goes through `GetEvaluationsForTutorInPhase`.

**4. `deleteEvaluation` mapping everything to 500 (fixed in 5ec1340).** Took your advice and let #2025 land first, then merged `main` and added the case to the `evaluationErrorStatus` switch, which resolved the conflict in that function rather than fighting it. `feedbackItemErrorStatus` had the same gap, so it got the same case. `TestDeleteEvaluationRejectsTargetOutsideTeam` covers the exact scenario from the description; without the switch case it fails with 500 against the expected 403.

**5. On the request-scoped cache.** I do not think this would buy anything, and I would rather not add the plumbing for a no-op. A request-scoped `c.Set`/`c.Get` cache only pays off if a single request resolves teams more than once, and none of them do: every write path (`createOrUpdateEvaluation`, `deleteEvaluation`, both completion endpoints, both feedback item endpoints) calls `CheckEvaluationIsEditable` exactly once, so there is exactly one core call per request. The cache would be populated and discarded without ever being read.

The N in "N-competency peer form" is N separate HTTP requests, not N calls within one. `EvaluationForm.tsx` subscribes to `form.watch` per competency and fires one POST per score change, so ten competencies are ten requests, each with one team resolution. Request scope cannot collapse those. Getting the reduction you are after needs either a cross-request cache with the TTL and invalidation policy you were right to want to avoid, or a batch write endpoint so the client sends the form in one request. Happy to open a follow-up issue for the batch endpoint if you think the saving is worth it, but it seems out of scope here.

**Smaller notes.** Both documented in code: `checkEvaluationTarget`'s `default: return nil` now carries a comment pointing at the switch in `CheckEvaluationIsEditable` that rejects unknown types, and `teamOfMember` states that a participant belongs to exactly one team per phase, so the first match is the only match.

One caveat on verification: the `servers/assessment` suite is green per package on my machine, including everything I touched, but full-suite runs fail on a random unrelated package each time with trivial tests hanging 75 to 99 seconds (the same tests pass in about 2 seconds alone). That is container startup starvation locally, so CI is the authority on the full run.

mathildeshagl (approved):
All four blocking points are addressed, and I checked each against the branch head rather than the reply.

**1. Auth header plumbing.** `authHeaderFromContext` is gone; `CheckEvaluationIsEditable` takes `authHeader string` explicitly and the routers read `c.GetHeader("Authorization")` at the edge. Threading it through the two feedback-item paths that arrived with #2025 was the right call — that is six call sites now compiler-checked instead of four relying on a runtime type assertion.

**2. Tests can now catch (1).** `testutils.MockCore` records the header of every course phase data resolution and `TestTargetCheckForwardsAuthHeaderToCore` asserts it arrived. The assertion bites: it compares against the suite's own token, so an empty header fails it.

**3. Positive tutor case.** `TestCreateTutorEvaluation` accepts a record targeting `MockTeamTutors[0]`, verified through `GetEvaluationsForTutorInPhase`. Enabling `tutor_evaluation_enabled` and adding the tutor's `course_participation` row in the dump were both needed, and both are there.

**4. Delete path.** Letting #2025 land first and adding the case to the shared `evaluationErrorStatus` switch was the better resolution — `deleteEvaluation` now routes through it, `feedbackItemErrorStatus` got the same case, and `TestDeleteEvaluationRejectsTargetOutsideTeam` covers the scenario from the description.

**5. On the request-scoped cache — you are right and I was wrong.** I checked every write path: `createOrUpdateEvaluation`, `deleteEvaluation`, both completion endpoints and both feedback item endpoints each call `CheckEvaluationIsEditable` exactly once, so a request-scoped cache would be populated and never read. The N really is N separate requests from `EvaluationForm`'s per-competency POSTs, which request scope cannot collapse. A batch write endpoint is the only thing that would, and that is out of scope here — worth its own issue if the saving turns out to matter.

Both smaller notes are now documented in code. Approving.

PR #2039, opened 2026-08-12
`Assessment`: Add CampusOnline-compatible grade export

### LINKED ISSUE #2018: Assessment: Campus Online compatible Grade export
### Problem or opportunity

Since most users of prompt also have to use campus online to fill in the final grades, it would be cool to be able to upload a grade sheet export from campus online that is filled with the grades of the phase and then automatically downloaded again with the grades filled in. This can then be uploaded to campus online again. 

### Desired outcome

Easier way to fill in grades into campus online 

### Proposed solution

_No response_

### Alternatives considered

_No response_

### Additional context

_No response_

rappm:
## Summary

Lecturers had no way to get PROMPT's assessment grades into CampusOnline and had to retype every grade by hand. This adds a **CampusOnline Grade Export** card as the last card of the assessment settings: upload the exam list downloaded from CampusOnline, review what was matched, download the filled file, and upload that back into CampusOnline.

## Details

The whole flow is client-side, in `clients/assessment_component`. There is no new endpoint, no Go change and no migration: both inputs (the phase participants and the assessment completions) are already fetched by the assessment client, and every existing CSV feature in this repo parses and serializes in the browser (`encoding/csv` has zero hits under `servers/`).

**New card** — `src/assessment/pages/SettingsPage/components/GradeExportCard/`

- `GradeExportCard.tsx` spells out the five-step flow and wires the upload field, the match summary and the download button. Appended after `AssessmentReminderCard` in `SettingsPage.tsx`; `/settings` is already restricted to `PromptAdmin` and `CourseLecturer`, so no auth work was needed.
- `hooks/useCampusGradeExport.ts` holds only the parsed file in local state and derives the filled result from it, so a grade edited elsewhere is picked up without re-uploading. The uploaded `File` is deliberately not retained, which makes producing the download a pure function of state.
- `components/CampusCsvUploadField.tsx` is a drag-and-drop picker modelled on the matching phase's `UploadButton`, but props-only and with the drop zone itself as the button so it stays keyboard-operable.
- `components/CampusGradeFillSummary.tsx` reports filled rows plus four warning buckets.

**Round-trip fidelity** — `utils/campusCsvFile.ts`, `utils/campusCsvEncoding.ts`

Only `GRADE` and `DATE_OF_ASSESSMENT` are written; the other 24 columns come back exactly as CampusOnline wrote them, along with the column order, delimiter, line endings, trailing newline and blank rows.

- The file is parsed **positionally** (`header: false`) rather than into objects. Object mode collapses duplicate header names into a single key and reshapes rows whose field count differs from the header, both of which lose data on the way back out. Addressing the grade column by index avoids both.
- Encoding is tracked per file and restored on the way out. CampusOnline usually exports Windows-1252 and `TextEncoder` only writes UTF-8, so writing back needs an explicit Windows-1252 encoder or every umlaut in a student name is corrupted. UTF-8 with a BOM keeps its BOM; UTF-16 is rejected with an actionable message rather than round-tripped badly.

**Matching** — `utils/fillCampusGrades.ts`

`REGISTRATION_NUMBER` is matched against the student's matriculation number, tolerating leading zeros on either side, and falls back to family plus first name. A key that resolves to more than one student is reported as ambiguous and left untouched, because grading the wrong student is the worst outcome this feature could produce.

Grades come from `assessment_completion.grade_suggestion` for assessments marked as complete. Students without a finalized assessment keep their empty grade cell and are listed as skipped, never auto-failed, and the download is never blocked. Grades outside the grade vocabulary are treated as "no grade", which is what the legacy `6.0` backfill in the assessment database means.

The summary also reports students who are graded in PROMPT but have **no row in the file at all** — their grade would otherwise silently never reach CampusOnline, which usually means the file was downloaded for the wrong exam date or course group.

**Two formats are assumptions**, isolated as single constants in `utils/campusGradeFormat.ts`: grades use a decimal point (`1.7`) and dates `DD.MM.YYYY`, mirroring the file's own values when the column already carries dates. Nothing in this repo documents what CampusOnline's import parser accepts, so these want one check against a real import; changing either is a one-line change by design.

**Dependency** — `papaparse` is added to `assessment_component` (already the parser used by `matching_component`). CampusOnline quotes its fields, so a delimiter may legally appear inside a value and splitting on the delimiter is not safe. `@types/papaparse` comes with it because the repo forbids `any`; that made `matching_component`'s bare `Papa.ParseResult` fail typecheck, so it now names its row type — the one change outside the assessment component.

## Reason / Link to issue

Closes #2018

## How to Test

1. `make db && make server && make server-assessment && make clients`
2. Open an assessment phase, grade two students and mark their assessments as final; leave a third ungraded.
3. Build a CSV with the header from #2018 (`;`-delimited, all fields quoted, ideally Windows-1252 with an umlaut in one name) containing those three students plus one unknown registration number.
4. Go to **Settings**, scroll to the last card, and upload the file. The summary should report 2 filled, 1 not graded yet and 1 unmatched.
5. Click **Download Filled CSV** and diff it against the input:
   ```bash
   diff <(iconv -f cp1252 -t utf-8 in.csv) <(iconv -f cp1252 -t utf-8 out.csv)
   ```
   Only `GRADE` and `DATE_OF_ASSESSMENT` cells may differ. The umlaut, the quoting, the delimiter and the line endings must be unchanged.

Automated:

```bash
make lint
make test-e2e   # or: docker compose -f docker-compose.e2e.yml --env-file e2e/.env.e2e run --rm \
                #        e2e-runner npx playwright test tests/assessment/grade-export.spec.ts
```

`e2e/tests/assessment/grade-export.spec.ts` covers the happy path, a Windows-1252 file keeping its encoding, and a non-CampusOnline file being rejected. The first asserts round-trip fidelity rather than just the grade (byte-identical header, all 26 columns in order, a semicolon quoted inside a course title surviving, the unmatched row keeping its empty grade); the second asserts the umlaut comes back as the single Windows-1252 byte rather than the UTF-8 pair. A dedicated seeded phase (`d0000011`) keeps release state and schema locking from crossing the print and visibility specs, following the existing one-phase-per-spec convention.

## PR Checklist

## Note for reviewers

`make test` is currently red on this branch: `servers/self_team_allocation/team` fails `TestAssignTeamRoute`, `TestAssignTeamRouteTeamFull` and `TestLeaveTeamRoute` with 500s. This PR contains no Go or server files, so those failures are pre-existing and unrelated. `servers/assessment` passes in full.

## Summary by CodeRabbit

* **New Features**
  * Added CampusOnline grade export to assessment settings.
  * Upload CSV files by selecting or dragging and dropping.
  * Automatically match students, fill grades, and download completed CSVs.
  * View processing summaries, warnings, skipped entries, and missing students.
  * Added controls for releasing and hiding assessment and evaluation results.
  * Assessment settings now adapt when grading is disabled.
  * Supports common CSV encodings while preserving file structure.

* **Documentation**
  * Added guidance for grade exports and evaluation-only assessment phases.

* **Tests**
  * Added end-to-end coverage for export workflows and assessment modes.


clients/assessment_component/src/assessment/pages/SettingsPage/components/GradeExportCard/utils/campusCsvFile.ts
```diff
@@ -0,0 +1,82 @@
...
+ * Rebuilds the file from `rows`, restoring the original delimiter, line ending
+ * and trailing newline. Blank rows stay blank instead of turning into a single
+ * quoted empty field.
+ */
+export const serializeCampusCsv = (csv: ParsedCampusCsv): string => {
+  const lines = csv.rows.map((row) =>
+    isBlankRow(row) ? '' : row.map(escapeCsvField).join(csv.delimiter),
+  )
+
+  return lines.join(csv.newline) + (csv.hasTrailingNewline ? csv.newline : '')
```
mathildeshagl, line 82: The trailing newline is counted twice here.

With `skipEmptyLines: false` above, a file ending in `\r\n` already parses to a final `['']` row. `lines` therefore ends with an empty string and the `join` puts a newline in front of it — and then `hasTrailingNewline` appends another. Reproduced against the real papaparse in this repo using the `buildCampusOnlineCsv` fixture:

```
parsed row count: 3        (header, one data row, trailing [''])
input  tail:  ..."";""\r\n
output tail:  ..."";""\r\n\r\n
extra bytes:  2
```

Real exports end with a newline, so this is the normal path rather than an edge case, and it contradicts the fidelity guarantee in the file's own doc comment.

Either drop the `hasTrailingNewline` re-append, since the trailing empty row already encodes it, or strip that row at parse time and keep the flag as the single source of truth. The second reads better to me — the flag then means what its name says, and `fillCampusGrades` stops having to skip a blank row it never wanted.

clients/assessment_component/src/assessment/pages/SettingsPage/components/GradeExportCard/utils/campusCsvFile.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  const newline = result.meta.linebreak || DEFAULT_NEWLINE
+
+  return {
+    fileName,
+    rows,
+    columnIndices,
+    delimiter: result.meta.delimiter || ';',
+    newline,
+    hasTrailingNewline: file.text.endsWith(newline),
```
mathildeshagl: This flag and the trailing `['']` row that `skipEmptyLines: false` produces encode the same fact, which is what makes the double-append below possible. Whichever fix you pick, only one of the two should survive.

e2e/tests/assessment/helpers.ts
```diff
@@ -349,3 +349,114 @@ export async function resetAssessmentPhase(
...
+    } else if (char === '\n') {
+      endRow()
+    } else {
+      field += char
+    }
+  }
+
+  if (field !== '' || row.length > 0) endRow()
+
+  return rows.filter((entry) => entry.length > 1 || entry[0] !== '')
```
mathildeshagl: This filter is why the extra trailing line in the download is invisible to the suite: blank rows are dropped before any assertion sees them, so `toHaveLength(3)` in the spec passes whether or not the round trip is faithful.

Reasonable for the row-level assertions, but it means nothing in the test actually checks fidelity at the byte level except the header line. Worth keeping the filter for the row checks and adding a separate assertion on the raw string.

e2e/tests/assessment/grade-export.spec.ts
```diff
@@ -0,0 +1,203 @@
...
+      'No matching student in this phase',
+      UNKNOWN_REGISTRATION_NUMBER,
+    )
+    await phase.expectGradedStudentMissingWarning(SEEDED_PHASE_STUDENTS.student2.lastName)
+
+    const downloaded = (await phase.downloadFilledCampusCsv()).toString('utf-8')
+    const rows = parseCampusOnlineCsv(downloaded)
+
+    // The header must come back exactly as CampusOnline wrote it.
+    expect(downloaded.split('\r\n')[0]).toBe(campusOnlineHeaderLine())
```
mathildeshagl, line 124: This is the only byte-level assertion in a feature whose central claim is byte-level fidelity, and it covers one line.

Something like this would carry the claim properly and would have caught the trailing-newline bug:

```ts
const expected = csv
  .split('\r\n')
  .map((line, index) => (index === stanRowIndex ? withGradeAndDate(line) : line))
  .join('\r\n')
expect(downloaded).toBe(expected)
```

Then the per-column checks below become the readable explanation of what changed rather than the entire guarantee.

docs/user/assessment.md
```diff
@@ -171,6 +171,46 @@ If a **different schema** is configured for self or peer evaluation, those score
...
+2. In PROMPT, open **Settings** in the assessment phase and upload that file into the CampusOnline Grade Export card.
+3. Check the summary. PROMPT reports which rows it filled, which it skipped, and why.
+4. Click **Download Filled CSV**.
+5. Upload the downloaded file back into CampusOnline.
+
+Steps 1 and 5 happen in CampusOnline; PROMPT is only involved in the middle.
+
+### How Students Are Matched
+
+PROMPT matches each row of your file to a student of the phase by **registration number** (compared against the student's matriculation number, ignoring leading zeros). If a row has no usable registration number, PROMPT falls back to matching by **family name and first name**.
```
mathildeshagl: The code is more generous than this sentence. `fillCampusGrades` falls back to name matching whenever the registration lookup returned no candidates — which includes a row that has a registration number that simply does not match anyone, not only a row that lacks one.

That is the better behaviour (a mistyped number still gets matched, and a name collision is reported as ambiguous rather than guessed), so I would change the sentence rather than the code.

mathildeshagl (changes requested):
This is careful work and the hard parts are right. Parsing positionally to dodge papaparse's object-mode collapsing of duplicate headers, tracking the encoding per file and writing it back, reporting ambiguity instead of guessing, and the "graded in PROMPT but absent from the file" bucket — that last one is the failure that would otherwise silently lose a grade, and I would not have thought to ask for it. I checked the Windows-1252 table against the standard: all 27 C1 positions are correct and the holes at 0x81, 0x8D, 0x8F, 0x90, 0x9D are right. The `?` fallback in the encoder is unreachable in practice, since everything written back either came out of a Windows-1252 file or is ASCII.

Requesting changes on one concrete defect in the round trip the PR is built around, and on the test that should have caught it.

**Every file that ends with a newline gains a blank line.** `parseCampusCsv` uses `skipEmptyLines: false`, so a trailing newline already materialises as a final `['']` row, and `serializeCampusCsv` then emits that row as an empty line *and* appends `csv.newline` again for `hasTrailingNewline`. I ran the real papaparse in this repo against the actual `buildCampusOnlineCsv` fixture:

```
parsed row count: 3          (header, one data row, trailing [''])
input  tail:  ..."";""\r\n
output tail:  ..."";""\r\n\r\n
extra bytes:  2
```

Real CampusOnline exports end with a newline, so this is the normal path, not an edge case. Whether the import parser tolerates it I cannot say — but the PR's claim is byte fidelity, and this breaks it.

**The e2e cannot catch it, by construction.** `parseCampusOnlineCsv` filters blank rows out before any assertion runs, so `toHaveLength(3)` passes either way, and the only byte-level check is the header line. Given that round-trip fidelity is the whole thesis, I would rather see one assertion that the download equals the input with only the `GRADE` and `DATE_OF_ASSESSMENT` cells substituted. That single check would have caught this and would keep catching the next one.

Non-blocking, all in the docs and the summary:

- `docs/user/assessment.md` says the name fallback applies when a row "has no usable registration number", but the code falls back whenever the number matched nothing — which also covers a present-but-wrong number. The code's behaviour is the more useful one; the docs should describe it.
- The same section says the file comes back exactly as delivered, but every field is deliberately re-quoted. That is documented in `campusCsvFile.ts` and not in the user docs, and a lecturer diffing the two files will notice.
- The description says the download is never blocked, but the button is disabled when `report.filled.length === 0`. Only bites when nothing matched at all, so it is defensible — just not what the text says.
- A student appearing twice in the uploaded file gets both rows filled with no warning, while every other anomaly gets one. Cheap to add to the summary.
- `@types/papaparse` lands in `assessment_component` but the typecheck it fixed was in `matching_component`, which is where papaparse is actually a dependency. Putting the types next to the package they type would keep the two components independent.

mathildeshagl (approved):
Approving. All three blocking points from my last review are addressed, and addressed the way I'd hoped rather than the cheap way.

I re-verified the fix rather than reading the diff: ran the branch's `parseCampusCsv`/`serializeCampusCsv` pair against the real papaparse in this repo, using the actual 26-column CampusOnline header. The round trip is byte-faithful for CRLF-with-trailing-newline, no trailing newline, an interior blank line, a doubled trailing newline, LF-only line endings, a semicolon inside a quoted field, header-only, and a comma-delimited file.

- **Trailing newline.** `bbae9e0` strips the terminal `['']` row at parse time, so `hasTrailingNewline` is now the single source of truth and `fillCampusGrades` no longer has to skip a blank row it never wanted. That was the option I preferred and it reads better than the alternative.
- **The test that should have caught it.** `7feb2d3` goes further than I asked. The full-file byte assertion is there, the blank-row filter is gone from the assertion parser so a dropped or added line now shows up, and `encodeWindows1252` is a real CP1252 mapping instead of a Latin-1 truncation, exercised by a fixture with `€` and a smart quote. Those two characters are exactly the ones that distinguish the encodings, and asserting on bytes `0x80` and `0x93` is the right way to prove it.
- **Docs.** `6935d34` describes the actual matching behaviour.

Non-blocking, and only the ones still open from last time. All one-liners, fine as follow-ups:

- `docs/user/assessment.md` still says every other column comes back "exactly as CampusOnline delivered", without mentioning that every field is deliberately re-quoted. That's documented in `campusCsvFile.ts` but not where a lecturer diffing the two files would look.
- The PR description still says the download is never blocked; the button is still disabled on `report.filled.length === 0`.
- A student appearing twice in the uploaded file still gets both rows filled with no warning, while every other anomaly gets one.
- `@types/papaparse` lives in `assessment_component` while `matching_component` has `papaparse` without types. Weaker than I made it sound last time, since deps hoist to `clients/node_modules` and resolve either way. Separately: `assessment_component` pins `^5.5.4` while core and matching are on `^5.6.0`.

Nice work on the round trip.

PR #2040, opened 2026-08-12
`Testing`: Speed up e2e stack rebuilds with Docker dependency caching

### LINKED ISSUE #2023: E2E: suite runtime is dominated by Docker rebuilds and serial boot
The e2e suite is slow, but most of the wall clock is **Docker build and stack boot, not Playwright**. On a warm-cache local run the e2e images alone took roughly 4 minutes to rebuild before the runner even started. This issue collects the places worth looking at, roughly in order of impact.

## 1. Build phase (dominant cost)

- **`make test-e2e` always rebuilds.** `Makefile:141-147` unconditionally runs `docker compose build` before `run` — there is no cache-only / skip-build path, so a one-line spec edit pays for a full build.
- **No `.dockerignore` in the component build contexts.** `clients/.dockerignore` exists, but Docker reads the ignore file from the *context root*, and the component images build from `./clients/core`, `./clients/interview_component`, etc. — none of which have one. That ships ~2.9 GB of host `node_modules` plus `build/` output to the daemon per run (assessment 828 MB, matching 544 MB, interview 543 MB, self team allocation 472 MB, certificate 294 MB, core 247 MB), and `COPY . ./` then bakes it into the image on top of the base image's correctly-installed Linux modules. Side effect worth checking: with `nodeLinker: node-modules`, darwin-arm64 native binaries end up shadowing the Linux ones.
- **`clients/Dockerfile` copies source before installing.** `COPY . ./` precedes `yarn install --immutable`, so any change in any client invalidates the install layer and triggers a full monorepo reinstall. It also installs **all 10 workspaces** when the e2e stack only needs 6. Worth restructuring to the usual manifests-first ordering.
- **Go images have no build cache mounts.** `servers/Dockerfile` runs `COPY . .` → `go build` with no `--mount=type=cache` for `/root/.cache/go-build` or `/go/pkg/mod`, so every source change recompiles the full dependency graph from scratch, across five services.
- **Six independent cold rspack production builds**, each in its own container with no shared or persistent compiler cache.

## 2. Stack boot

- **Long serial dependency chain:** `clients-base` → six client builds → `client-core` (gated on 9 `service_started` deps) → `e2e-runner` (gated on 10 health conditions). Health checks use `interval: 5s` / `start_period: 10s`, so readiness is quantized in 5s steps, and `global-setup`'s own pollers sleep 2s between attempts across six sequential probes.
- **Six sequential Keycloak logins.** `e2e/src/global-setup.ts:97-128` loops the roles one at a time, each with a fresh browser context through the real login form (`waitForURL` allows 30s each). They are fully independent — running them concurrently, or moving them to a Playwright `setup` project, would collapse six round-trips into one.

## 3. Test execution

- **`workers: isCI ? 1 : undefined`** (`e2e/playwright.config.ts:13`) — CI runs ~35 spec files single-threaded, and `retries: 2` makes a flaky spec cost 3x its runtime. Worth re-checking whether the single-worker constraint is still real: the comment at `playwright.config.ts:8-9` cites a shared Keycloak session, which the per-role `storageState` approach already addresses.
- **`fullyParallel: false`** — parallelism is per-file only, so the tail is bounded by the slowest single file (the assessment / interview / certificate journey specs).
- **Uniform 15s expect timeouts** across nearly every Page Object, plus `expect.timeout: 10_000` and a 60s test timeout. Free on the happy path, but every failure burns the full budget — times three on CI.

## Suggested order

1. Add `.dockerignore` to `clients/core` and each `clients/*_component` — best impact-to-effort ratio, and it removes the native-binary shadowing at the same time.
2. Restructure `clients/Dockerfile` to install before copying source, scoped to the workspaces e2e actually needs.
3. Add Go build cache mounts to `servers/Dockerfile`.
4. Parallelize the role logins in `global-setup`.
5. Revisit the CI worker count.

Items 1-3 should bring the warm-cache rebuild close to zero; 4-5 shorten the runtime tail.

rappm:
## Summary

Cuts the Docker rebuild time that dominates the e2e suite by fixing the build contexts, adding
BuildKit caches for the Go and Yarn dependency layers, and reordering the client Dockerfile so a
source edit no longer re-runs `yarn install`. Also adds `make test-e2e SKIP_BUILD=1` for iterating on
specs without a build pass.

## Details

**Client build contexts (`clients/.dockerignore`, `clients/*/.dockerignore`)**

- Adds a `.dockerignore` to `clients/core` and every `clients/*_component`, so host-generated
  artifacts (`node_modules`, `.cache`, `dist`, tool caches) never enter the per-component build
  context. The rspack persistent cache under `node_modules/.cache/rspack` alone is 250 MB to 850 MB
  per component on a developer machine. Contents are intentionally identical across the components
  and the files say so, so they stay in sync.
- Adds `.yarn/install-state.gz` to `clients/.dockerignore`, since the install now runs before the
  source is copied and the host copy would otherwise land on top of the container's.

**Go service builds (`servers/Dockerfile`, `servers/certificate/Dockerfile`)**

- Pins `GOMODCACHE` and `GOCACHE` to the toolchain defaults, then mounts both as BuildKit caches for
  `go mod download` and `go build`. The mount IDs match across both Dockerfiles, so every Go service
  shares one module cache instead of each downloading its own copy.
- The module cache mount is repeated on the `go build` step on purpose: cache mounts are not part of
  the image, so without it the build would re-download. The build cache is what makes a source-only
  change recompile the changed package instead of the whole dependency graph.

**Client dependency layer (`clients/Dockerfile`)**

- Copies `package.json`, `yarn.lock`, `.yarnrc.yml`, and each workspace manifest before running
  `yarn install --immutable`, so the install layer is only invalidated by a manifest or lockfile
  change. A source edit now only invalidates the final `COPY . ./`.
- Pins `YARN_GLOBAL_FOLDER` so the BuildKit cache mount provably targets the directory Yarn 4
  actually writes its downloads to.
- The workspace manifest copies are one line per entry in the `workspaces.packages` array. That array
  is an explicit list rather than a glob, so a workspace added there but not here makes
  `yarn install` fail instead of drifting silently.
- Bumps the base image to `node:24.19.0` to match the repo-wide Node version.

**Skip-build escape hatch (`Makefile`, `e2e/README.md`)**

- `make test-e2e SKIP_BUILD=1` reuses the images already on the machine and skips the build pass.
  Documented as only safe when nothing under `clients/`, `servers/`, or `e2e/` changed since the last
  build.

**Playwright config (`e2e/playwright.config.ts`)**

- Comment-only. Corrects the stale claim that login flows share a Keycloak session (global-setup
  writes one `storageState` per role) and records why `workers` is still 1 on CI: a few spec files
  share a seeded phase, which needs carving out into dedicated phases before the worker count can go
  up.

## Reason / Link to issue

Closes #2023. Every `make test-e2e` run rebuilt all service images from scratch: the Go services
re-downloaded their modules and recompiled the full dependency graph, the client images re-ran
`yarn install` on any source edit, and the per-component build contexts shipped hundreds of megabytes
of host caches to the daemon.

## How to Test

1. `make test-e2e` once to populate the BuildKit caches, then touch a Go source file under
   `servers/core/` and run it again. The rebuild should skip `go mod download` and recompile only the
   changed package.
2. Touch a client source file (e.g. under `clients/core/src/`) and rerun. The `yarn install` layer
   should be a cache hit.
3. `make test-e2e SKIP_BUILD=1` with no source changes. The suite should run against the existing
   images without a build pass.
4. Confirm the suite still passes end to end and that `make test-e2e-down` cleans up.

## PR Checklist

## Summary by CodeRabbit

* **Chores**
  * Improved application and service container builds with better caching, reducing repeat build times.
  * Reduced unnecessary files included in component container builds for more efficient packaging.
  * Added an option to reuse existing end-to-end test images, speeding up test iteration.

* **Documentation**
  * Updated end-to-end testing guidance with image reuse instructions and cautions about stale changes.
  * Clarified test setup requirements and execution behavior for consistent local and CI runs.


mathildeshagl (dismissed):
Approving — the Docker work is correct. The eight workspace manifest copies match `workspaces.packages` exactly, `COPY . ./` after the install is safe because `node_modules` stays ignored, and pinning `YARN_GLOBAL_FOLDER` / `GOMODCACHE` / `GOCACHE` before mounting them is the right defensive move rather than trusting the toolchain defaults to stay put. Repeating the module cache mount on the `go build` step is correct and the comment saying why is the sort of thing that stops someone "cleaning it up" later.

One thing I would like resolved before this closes the issue, and three small ones.

**The caching does nothing on CI, and #2023 is about CI.** `.github/workflows/test-e2e.yml` runs `docker compose build` directly on a fresh `ubuntu-latest` runner — no `setup-buildx-action`, no `cache-from`/`cache-to`, no persisted BuildKit state. `RUN --mount=type=cache` contents are never exported by the gha layer cache, so every CI run still cold-downloads the Go modules and re-runs `yarn install`. The `.dockerignore` win is close to zero there too: a fresh checkout has no `node_modules/.cache/rspack` to exclude in the first place. So the benefit here is entirely local iteration, which is real and worth having, but the issue is titled "e2e suite runtime is dominated by docker rebuilds and serial boot" and this addresses neither half on CI — the `playwright.config.ts` change is comment-only and `workers` stays at 1. I would either downgrade `Closes #2023` to `Refs`, or add buildx with `cache-from/to: type=gha` to the e2e workflow so the CI half actually lands.

Smaller:

- `E2E_BUILD = $(if $(SKIP_BUILD),true,…)` treats any non-empty value as true, so `make test-e2e SKIP_BUILD=0` silently skips the build — the opposite of what someone typing `0` means. `$(filter 1 true yes,$(SKIP_BUILD))` would read the way the flag is documented.
- `test-e2e-ui` still hard-codes `run --rm --build`, so the target you iterate in most cannot reuse images. Odd asymmetry given the motivation.
- The Go mounts use the default `sharing=shared` while the Yarn one right beside them uses `sharing=locked`, and compose builds services in parallel. Go's own locking makes this safe, but since the two mounts made opposite choices a sentence saying it was deliberate would save the next reader the trip.
- Eight byte-identical `.dockerignore` files kept in sync by a comment: the ninth component is the one that will drift.
- The description says it bumps the base image to `node:24.19.0`; `main` already has that and the diff does not touch the `FROM`.

PR #2057, opened 2026-08-14
`Core`: Adjust Student History View Access Rights to include CourseLecturer, with corresponding filtering

JGStyle:
## ✨ What is the change?

The PR adjusts the Student History View access route to be accessible to Course_Lecturers and uses the existing Course filtering

Also add tests to verify the new behaviour.

Follow up fix for the Student History View

## 🧪 How to Test

Verify that an instructor with Course Lecturer role for a specific course or more can only see
those courses that he has the role for inside of the student history view.

An admin or PromptLecturer can still see both.

Instructor with only access to course the student applied for -can only see the single course

Admin - can see everything

| Admin | Limited Instructor |
| --- | --- |
| 2026-08-14 at 22 32 22@2x | 2026-08-14 at 22 32 00@2x |

## ✅ PR Checklist

- [!] Code is clean, readable, and documented
- [!] Documentation updated (if relevant)

## Summary by CodeRabbit

* **New Features**
  * Course lecturers and editors can now view enrollments for courses they can access.
  * Enrollment results are limited according to the viewer’s course permissions.
* **Bug Fixes**
  * Instructor notes are now visible only to administrators and lecturers.
  * Unauthorized or invalid enrollment access requests return appropriate errors.
* **Refactor**
  * Standardized course and role naming to improve consistency across permissions and course management.

### DIFF
--- a/servers/core/student/router.go
+++ b/servers/core/student/router.go
@@ -204,7 +204,11 @@ func getStudentEnrollments(c *gin.Context) {
 		handleError(c, http.StatusForbidden, errors.New("missing user roles"))
 		return
 	}
-	userRoles := rolesVal.(map[string]bool)
+	userRoles, ok := rolesVal.(map[string]bool)
+	if !ok {
+		handleError(c, http.StatusInternalServerError, errors.New("invalid roles format in context"))
+		return
+	}
 
 	studentEnrollments, err := GetStudentEnrollmentsByID(c, id)
 	if err != nil {
--- a/servers/core/student/router.go
+++ b/servers/core/student/router.go
@@ -19,7 +19,7 @@ func setupStudentRouter(router *gin.RouterGroup, authMiddleware func() gin.Handl
 	student.GET("/", permissionRoleMiddleware(permissionValidation.PromptAdmin, permissionValidation.PromptLecturer), getAllStudents)
 	student.POST("/", permissionRoleMiddleware(permissionValidation.PromptAdmin, permissionValidation.PromptLecturer), createStudent)
 	student.PUT("/:uuid", permissionRoleMiddleware(permissionValidation.PromptAdmin, permissionValidation.PromptLecturer), updateStudent)
-	student.GET("/:uuid/enrollments", permissionRoleMiddleware(permissionValidation.PromptAdmin, permissionValidation.PromptLecturer), getStudentEnrollments)
+	student.GET("/:uuid/enrollments", permissionRoleMiddleware(permissionValidation.PromptAdmin, permissionValidation.PromptLecturer, permissionValidation.CourseLecturer, permissionValidation.CourseEditor), getStudentEnrollments)
 }
 
 // getAllStudents godoc
@@ -198,11 +198,21 @@ func getStudentEnrollments(c *gin.Context) {
 		handleError(c, http.StatusBadRequest, parseErr)
 		return
 	}
-  studentEnrollments, err := GetStudentEnrollmentsByID(c, id)
+
+	rolesVal, exists := c.Get("userRoles")
+	if !exists {
+		handleError(c, http.StatusForbidden, errors.New("missing user roles"))
+		return
+	}
+	userRoles := rolesVal.(map[string]bool)
+
+	studentEnrollments, err := GetStudentEnrollmentsByID(c, id)
 	if err != nil {
 		handleError(c, http.StatusInternalServerError, err)
 		return
 	}
+
+	studentEnrollments = FilterEnrollmentsToAccessibleCourses(studentEnrollments, userRoles)
 	c.IndentedJSON(http.StatusOK, studentEnrollments)
 }
 
--- a/servers/core/student/service.go
+++ b/servers/core/student/service.go
@@ -219,9 +219,9 @@ func FilterEnrollmentsToAccessibleCourses(enrollments studentDTO.StudentEnrollme
 
 	accessibleCourses := make([]studentDTO.CourseEnrollmentDTO, 0, len(enrollments.Courses))
 	for _, course := range enrollments.Courses {
-		courseIdentifier := course.SemesterTag + "-" + course.Name
-		if userRoles[courseIdentifier+"-"+permissionValidation.CourseLecturer] ||
-			userRoles[courseIdentifier+"-"+permissionValidation.CourseEditor] {
+		courseIdentifier := permissionValidation.CourseIdentifier(course.SemesterTag, course.Name)
+		if userRoles[permissionValidation.CourseRoleName(courseIdentifier, permissionValidation.CourseLecturer)] ||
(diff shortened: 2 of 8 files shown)


servers/core/student/router.go
```diff
@@ -19,7 +19,7 @@ func setupStudentRouter(router *gin.RouterGroup, authMiddleware func() gin.Handl
 	student.GET("/", permissionRoleMiddleware(permissionValidation.PromptAdmin, permissionValidation.PromptLecturer), getAllStudents)
 	student.POST("/", permissionRoleMiddleware(permissionValidation.PromptAdmin, permissionValidation.PromptLecturer), createStudent)
 	student.PUT("/:uuid", permissionRoleMiddleware(permissionValidation.PromptAdmin, permissionValidation.PromptLecturer), updateStudent)
-	student.GET("/:uuid/enrollments", permissionRoleMiddleware(permissionValidation.PromptAdmin, permissionValidation.PromptLecturer), getStudentEnrollments)
+	student.GET("/:uuid/enrollments", permissionRoleMiddleware(permissionValidation.PromptAdmin, permissionValidation.PromptLecturer, permissionValidation.CourseLecturer, permissionValidation.CourseEditor), getStudentEnrollments)
```
rappm, line 22: Opening this route to `CourseLecturer` / `CourseEditor` makes the enrollments card work, but the card directly above it in the same dialog still fails.

`ApplicationDetailsPage.tsx:195-198` renders `InstructorNotes` and `CourseEnrollments` side by side, and `GET /instructor-notes/s/:student-uuid` is still registered for `PromptAdmin, PromptLecturer` only (`instructorNote/router.go:18`).

Failure case: a course lecturer opens the application details dialog for a participant. The enrollment list now populates, directly underneath the literal text `error fetching instructor notes` (`InstructorNotes.tsx:12-14`). Both cards failed before this PR, so this is not a regression, but a half-working dialog reads as a new bug in the half that was just fixed. The standalone `/management/students/:studentId` page is `PROMPT_ADMIN`-gated in `App.tsx:106`, so this dialog is the only affected surface.

Suggested fix: give the notes route the same per-course filtering, or hide the card for callers who lack the role. If neither belongs in this PR, an issue linked from here would do.

servers/core/student/router.go
```diff
@@ -198,11 +198,21 @@ func getStudentEnrollments(c *gin.Context) {
...
 		return
 	}
-  studentEnrollments, err := GetStudentEnrollmentsByID(c, id)
+
+	rolesVal, exists := c.Get("userRoles")
+	if !exists {
+		handleError(c, http.StatusForbidden, errors.New("missing user roles"))
+		return
+	}
+	userRoles := rolesVal.(map[string]bool)
```
rappm: Unchecked type assertion. `permissionValidation/middleware.go:59` and `permissionValidation/checkPermission.go:22` both read this context key with the comma-ok form and return a 500 when it does not hold a `map[string]bool`; here a wrong type panics the handler.

Only `keycloakTokenVerifier` (and the SDK mock) write the key today, so nothing reaches it, but the `exists` check three lines above already handles the missing case and then gives the same safety away for the wrong-type case.

`userRoles, ok := rolesVal.(map[string]bool)` with a 500 on `!ok` costs one line and keeps this consistent with the other two readers.

servers/core/student/service.go
```diff
@@ -210,3 +211,20 @@ func GetStudentEnrollmentsByID(ctx context.Context, id uuid.UUID) (studentDTO.St
...
 }
+
+func FilterEnrollmentsToAccessibleCourses(enrollments studentDTO.StudentEnrollmentsDTO, userRoles map[string]bool) studentDTO.StudentEnrollmentsDTO {
+	if userRoles[permissionValidation.PromptAdmin] || userRoles[permissionValidation.PromptLecturer] {
+		return enrollments
+	}
+
+	accessibleCourses := make([]studentDTO.CourseEnrollmentDTO, 0, len(enrollments.Courses))
+	for _, course := range enrollments.Courses {
+		courseIdentifier := course.SemesterTag + "-" + course.Name
```
rappm: This is the fourth place that assembles a Keycloak role name by hand. `GetPermissionStringByCourseID` in `permission_strings.sql` and `permissionValidation.checkUserRole` (`fmt.Sprintf("%s-%s", courseIdentifier, role)`) already own the `<semesterTag>-<name>-<role>` scheme.

The failure mode of a copy is silent: if the convention changes again, the way it did when the `-cg-` custom roles were added, this filter matches nothing, and a legitimate course lecturer gets an empty `courses` array with a 200 and no log line. Building the identifier through the existing helper turns that into a compile or test failure instead.

Two things I checked that are fine as written: the string matches how `keycloakRealmManager/addingCourse.go:32` provisions the roles, and it is more robust than the `split_part` in the SQL for course names that contain a `-`. `UpdateCourse` (`db/query/course.sql:107`) touches neither `name` nor `semester_tag`, so a rename cannot desync it either.

servers/core/student/service_test.go
```diff
@@ -107,3 +108,61 @@ func (suite *ServiceTestSuite) TestCreateStudent() {
 func TestServiceTestSuite(t *testing.T) {
 	suite.Run(t, new(ServiceTestSuite))
 }
+
+func TestFilterEnrollmentsToAccessibleCourses(t *testing.T) {
```
rappm, line 112: Good table, and the `CourseStudent` case is the right one to pin. It only covers the filter in isolation, though.

The part of this change that can actually break is the pairing of the relaxed `permissionRoleMiddleware` on the route with the filter in the handler, and nothing exercises that: `router_test.go` has no enrollments case, and its mock auth middleware only ever injects `PROMPT_Admin`.

A router test that injects `ss25-ios-Lecturer` alone and asserts a 200 carrying only that course would pin the behavior the description promises, and would catch a later middleware change that widens the route without widening the filter.

rappm (changes requested):

servers/core/student/service.go
JGStyle: Fair critique. I introduced common helpers for this and made the places use that function in 509a4ac1a3c672c1542857e992271764bf9ffdfe

servers/core/student/router.go
JGStyle: The problem is that instructor notes are not really associated with any course in the current model l- so i couldn't say show these notes only to instructors that have this course.
For now, the most logical step is to hide, which i implemnted in the follow up commit.
This probably makes most sense once we have a Group Abstraction in prompt
I created an issue to track this #2059 (also created one for Student History View, where said we should also update the visiblity logic then #2060)
in 509a4ac1a3c672c1542857e992271764bf9ffdfe

servers/core/student/router.go
JGStyle: Fixed. Switched to userRoles, ok := rolesVal.(map[string]bool) with a 500 on !ok
in 509a4ac1a3c672c1542857e992271764bf9ffdfe

rappm (approved):

Issue #1988, opened 2026-07-22
Adopt the shared getStudentName util


rappm:
## Summary

Replace inline student display-name formatting with the shared util.

Blocked on prompt-edu/prompt-lib#234 (shared `getStudentName`).

## What is duplicated (verified, 5 components + core)

Inline `` `${firstName} ${lastName}` `` and a dedicated helper:

- assessment: `clients/assessment_component/src/assessment/pages/utils/getTeamMemberName.ts:7` (+ inline)
- interview: `InterviewCard.tsx`, `StudentCard.tsx`, `InterviewTimesDialog.tsx`
- self_team_allocation: `TeamCard.tsx`, `TutorImportDialog.tsx`
- team_allocation: `TeamAllocationPage.tsx`, `TutorSelection.tsx`, `StudentDetailDialog.tsx`

## Work

- Replace the inline concatenations and `getTeamMemberName.ts` with the shared `getStudentName`.

Issue #1987, opened 2026-07-22
Adopt shared role-preset constants in routes and sidebars


rappm:
## Summary

Replace the repeated permission arrays in every phase's routes and sidebar with the shared role
presets.

Blocked on prompt-edu/prompt-shared-state#60 (shared role-preset constants).

## What is duplicated (verified, 7 components x 2 files)

`[Role.PROMPT_ADMIN, Role.COURSE_LECTURER]` (and variants) in:

- `clients/*/src/*/routes/index.tsx`
- `clients/*/src/*/sidebar/index.tsx`

## Work

- Import the shared presets (e.g. `LECTURER_ROLES`) and delete the inline arrays.

Issue #1985, opened 2026-07-22
Remove inlined request try/catch wrappers in favor of shared error handling


rappm:
## Summary

Remove the inlined `try { ... } catch (err) { console.error(err); throw err }` wrapper from every
query/mutation once the shared axios instance handles logging centrally.

Blocked on prompt-edu/prompt-shared-state#59 (shared request error handling).

## What is duplicated (verified, ~130 files)

The identical wrapper across the network layers of assessment (~58), team_allocation (~21),
self_team_allocation (~12), certificate (~5), example (~1):

- `clients/example_component/.../network/queries/getExampleInfo.ts:3`
- `clients/assessment_component/.../network/queries/getCoursePhaseParticipations.ts:6`

## Work

- After the shared response interceptor lands, strip the try/catch from the ~130 call sites so each
  function is just `return (await instance.get/post(url)).data`.

rappm:
Still blocked. The upstream blocker prompt-edu/prompt-shared-state#59 is closed, but the shipped 1.2.0 factory does not actually carry the response error interceptor: `createAuthenticatedAxiosInstance` installs only the request interceptor, and `logAndRethrow` is attached separately to the two module-level singletons. `logNetworkError` is also not re-exported from the package, so consumers cannot attach it themselves.

Stripping the ~130 try/catch wrappers against 1.2.0 would silently drop all network error logging in the phase components. Filed prompt-edu/prompt-shared-state#78 to move the interceptor into the factory; this issue should wait for that release.

Context: #2068 adopted the factory across 6 phase components and deliberately left the wrappers in place for this reason.

PR #2068, opened 2026-08-20
`General`: Adopt the shared authenticated axios instance in phase components

### LINKED ISSUE #1979: Adopt the shared authenticated axios instance in phase components
## Summary

Delete the per-phase authenticated axios configs and import the shared factory. The local `Patch`
interface copies live in the same files and are dropped in the same change.

Blocked on prompt-edu/prompt-shared-state#57 (shared factory).

## What is duplicated (verified)

- Authenticated axios instance - 7 files across 6 components (byte-identical interceptor, differ only
  by `env.*_HOST` key):
  - `clients/example_component/.../network/exampleServerConfig.ts:16`
  - `clients/assessment_component/.../network/assessmentServerConfig.ts:15`
  - `clients/certificate_component/.../network/certificateServerConfig.ts:9`
  - `clients/interview_component/.../network/interviewServerConfig.ts:15`
  - `clients/self_team_allocation_component/.../selfTeamAllocationServerConfig.ts:15` (+ `coreServerConfig.ts:9`)
  - `clients/team_allocation_component/.../teamAllocationServerConfig.ts:15`
- Local `Patch` interface - 5 copies (assessment, example, interview, self_team_allocation,
  team_allocation `*ServerConfig.ts`).

## Work

- Replace each local config with `createAuthenticatedAxiosInstance(env.<PHASE>_HOST)`.
- Import `Patch` from `@tumaet/prompt-shared-state`, delete the 5 local copies.

rappm:
## Summary

Removes the duplicated authenticated axios setup from the phase components and replaces it with the
`createAuthenticatedAxiosInstance` factory that `@tumaet/prompt-shared-state` 1.2.0 exports.

## Details

- Bumped `@tumaet/prompt-shared-state` and `@tumaet/prompt-ui-components` to `1.2.0` in
  `clients/package.json` and regenerated `clients/yarn.lock`.
- Replaced the local `axios.create` plus request interceptor in six phase configs with
  `createAuthenticatedAxiosInstance(env.<PHASE>_HOST)`: assessment, certificate, example, interview,
  self team allocation, team allocation.
- Deleted the five dead local copies of the `Patch` interface (assessment, example, interview, self
  team allocation, team allocation). Nothing imported them, so no import rewiring was needed; the
  interface is now exported by `@tumaet/prompt-shared-state` for future consumers.
- Deleted `clients/self_team_allocation_component/.../network/coreServerConfig.ts` and pointed its
  single consumer at the shared `axiosInstance` singleton instead of the factory. That singleton is
  already bound to `env.CORE_HOST`, and the team allocation component calls the very same core
  endpoint through it, so reusing it removes a whole file rather than replacing one duplicate with
  another. It also brings the sanitized response error logging that the singleton carries.

Interceptor equivalence was verified against the shipped 1.2.0 build before deleting the local code:

- The local guard `localStorage.getItem('jwt_token') && ... !== ''` and the factory's `if (token)`
  behave identically, since the empty string is falsy. The factory adds a
  `typeof localStorage !== 'undefined'` guard, which is strictly safer.
- `config.headers.Authorization` and `config.headers['Authorization']` are the same assignment.
- The local configs resolved `parseURL(host)` once at module load; the factory resolves it per
  request inside the interceptor. Same value either way, because `env` is a snapshot normalized at
  module load.

No behavioral regressions found. The only behavior change is the additive response error logging on
the one self team allocation core query, which comes from switching to the shared singleton.

Out of scope: the inlined `try { } catch (err) { console.error(err); throw err }` wrappers in the
network layers stay untouched. `createAuthenticatedAxiosInstance` installs only the request
interceptor and `logNetworkError` is not publicly exported, so removing them would silently drop all
network error logging in the phase components. That cleanup is #1985 and remains blocked.

The dependency bump in the first commit is intentionally identical to the one on the branches for
#1987 and #1988, so those PRs are unblocked by the same version and the three branches merge cleanly
instead of conflicting on `clients/package.json` and `clients/yarn.lock`.

## Reason / Link to issue

Closes #1979

## How to Test

1. `cd clients && yarn install --immutable`
2. `make lint`
3. `cd clients/<component> && npx tsc --noEmit -p tsconfig.json && yarn build` for each of the six
   touched components.
4. Run the stack and open an assessment, interview, certificate, team allocation, and self team
   allocation phase. Requests must still carry the `Authorization: Bearer` header and hit the
   correct per-phase host.

## PR Checklist


mathildeshagl (dismissed):
Approving. Good refactor, and the interceptor-equivalence section in the description is the reason this was quick to review.

I checked the claims against the published `@tumaet/prompt-shared-state@1.2.0` tarball rather than taking them on trust, and they all hold:

- The factory does `parseURL(baseHost || '')` internally, so passing `env.X_HOST` straight in needs no local `|| ''`.
- `if (token)` is equivalent to the old `getItem() && getItem() !== ''`, and the `typeof localStorage !== 'undefined'` guard is strictly safer.
- `env` really is a module-load snapshot (`normalizeEnv(window.env)` evaluated at import), so resolving `parseURL` per request instead of once at module load gives the same value. That was the claim most worth checking and it is correct.

The refactor is also complete rather than partial: all seven duplicated setups under `clients/` are accounted for, six replaced and one deleted. And `Patch` is declared in five files and imported by none, so deleting it needed no import rewiring, as you say.

One equivalence the description doesn't cover, worth knowing rather than changing: `baseURL` moves from an instance default to a request interceptor, so a per-request `baseURL` override would now be overwritten instead of taking precedence. No call site does that today (the only `baseURL` occurrences in `clients/` are the ones this PR deletes), so it is latent rather than a bug. Absolute URLs passed as `url` are unaffected either way, since axios ignores `baseURL` for those.

Cosmetic, take it or leave it: `const x = createAuthenticatedAxiosInstance(...)` followed by `export { x }` could be a single `export const`. It does preserve the shape the files already had, which is a fair reason to keep it.

Agreed on leaving the `try/catch` wrappers alone until #1985 unblocks.

mathildeshagl (approved):

PR #2069, opened 2026-08-20
`General`: Adopt shared role-preset constants in routes and sidebars

### LINKED ISSUE #1987: Adopt shared role-preset constants in routes and sidebars

rappm:
## Summary

Replaces the permission arrays that were copy-pasted across every phase's `routes/index.tsx` and
`sidebar/index.tsx` with the `LECTURER_ROLES` and `EDITOR_ROLES` presets that
`@tumaet/prompt-shared-state` 1.2.0 now exports. No effective permission set changes anywhere in
this PR.

## Details

Three commits:

1. **Dependency bump.** `@tumaet/prompt-shared-state` and `@tumaet/prompt-ui-components` go from
   1.1.1 / 1.1.6 to 1.2.0, with `clients/yarn.lock` regenerated. 1.2.0 of the shared state package
   is what ships the presets (prompt-edu/prompt-shared-state#60). Both packages are bumped together
   so the dependency commit is identical to the one in the sibling branches for #1979 and #1988.
2. **Phase components.** 14 files across the 7 in-repo phase components.
3. **Core.** The same substitution in `clients/core/src/App.tsx`, the application phase routes, and
   the management console sidebar definitions. Route guards and sidebars in core carry the exact
   same duplicated arrays, so leaving them out would have left the dedup half done.

The presets, verified against the installed 1.2.0 package rather than assumed:

| Preset | Contents |
| --- | --- |
| `LECTURER_ROLES` | `[Role.PROMPT_ADMIN, Role.COURSE_LECTURER]` |
| `EDITOR_ROLES` | `[Role.PROMPT_ADMIN, Role.COURSE_LECTURER, Role.COURSE_EDITOR]` |

Only literal, exact matches for those two arrays were substituted. Nothing was widened or narrowed.

### Call sites deliberately left inline

- `[Role.PROMPT_ADMIN, Role.COURSE_LECTURER, Role.COURSE_EDITOR, Role.COURSE_STUDENT]`, the
  student-facing landing route and top-level sidebar entry in assessment, certificate, interview,
  self team allocation, and team allocation. There is no student-inclusive preset in 1.2.0, and
  inventing one locally would defeat the point of the shared constants. A `STUDENT_ROLES` preset in
  the shared package would be a reasonable follow-up.
- `[Role.PROMPT_ADMIN]` and `[Role.PROMPT_ADMIN, Role.PROMPT_LECTURER]` in `App.tsx`, the
  platform-admin routes. Neither matches a preset.
- `requiredPermissions: []` in `example_component/routes/index.tsx`, which intentionally means
  "no permission required".

The `Role` import stays in every file that still references the enum directly and was dropped from
the files that no longer do.

One incidental cleanup: `ApplicationRoutes.tsx` carried a `// empty means no permissions required`
comment on a non-empty array. It was already wrong before this PR and is now removed rather than
carried along.

## Pre-existing inconsistencies found, not changed here

Deliberately left as-is so this PR stays a pure refactor. Happy to open follow-up issues.

1. **Application phase, route vs sidebar.** The landing route (`path: ''`) allows
   `EDITOR_ROLES`, but the sidebar entry that leads to it requires `LECTURER_ROLES`. A Course Editor
   can reach the page by URL but gets no sidebar entry for it.
2. **`team_allocation_component/sidebar/index.tsx`.** The top-level sidebar item has no
   `requiredPermissions` at all, so it is unconditionally visible, while its route requires
   `[PROMPT_ADMIN, COURSE_LECTURER, COURSE_EDITOR, COURSE_STUDENT]`. Its sibling
   `self_team_allocation_component` does declare the four roles on the top-level item. Likely an
   oversight in team allocation.

## Reason / Link to issue

Closes #1987. Unblocked by prompt-edu/prompt-shared-state#60.

## How to Test

The important property is that no effective role set changed. That was checked mechanically rather
than by eye: for every touched file, expanding `LECTURER_ROLES` and `EDITOR_ROLES` back to their
literal arrays and reformatting reproduces the `origin/main` file byte for byte, with the only
diffs being the import statement itself and the one deleted stale comment.

To re-check by hand:

1. `cd clients && yarn install --immutable`
2. `npx biome check` and `npx tsc --noEmit -p tsconfig.json` in each touched component, plus
   `make lint` at the repo root.
3. Log in as a Course Editor and confirm the assessment and certificate Participants pages and the
   course configurator are still reachable, and that settings / statistics pages are still
   lecturer-only.

## PR Checklist


mathildeshagl (dismissed):
Approving. The property that matters here is that no effective permission set changed, and I checked that mechanically rather than by eye, in roughly the way you describe: expanded `LECTURER_ROLES` and `EDITOR_ROLES` back to their literal arrays in all 23 touched files and compared against the merge base with whitespace normalized.

17 files matched exactly. The other six differ only in the multi-line reformatting of the import statement itself, plus the one deleted stale `// empty means no permissions required` comment in `ApplicationRoutes.tsx`. Nothing was widened or narrowed. The preset contents are confirmed straight out of the 1.2.0 tarball and match your table.

Including core was the right call. Leaving `App.tsx` and the management console out would have left the arrays that guard the configurator and course settings duplicated, which is where a drift would matter most.

Both of your "pre-existing, not changed here" findings are real. I confirmed the second one and it is the more interesting of the two, so I have opened an issue for it: `ExternalSidebar.tsx` falls through to `hasComponentPermission = true` when `requiredPermissions` is absent, so the team allocation top-level sidebar item is unconditionally visible to any course member, including a student who is not in that phase. UI-only, since the route guard still holds, but worth fixing.

One note for the shared package rather than this PR: the presets are typed `Role[]` rather than `readonly Role[]`, and they are now a single module-level array aliased across roughly forty call sites. Every consumer I looked at is read-only (`some`, `includes`, `length`), so this is safe today, but `as const` upstream would make it safe by construction. Small side benefit of the change as written: passing a stable array reference to `PermissionRestriction` is better for memoization than a fresh literal on every render.

Agreed on leaving the student-inclusive array inline. Inventing a local `STUDENT_ROLES` would defeat the point; a preset in the shared package is the right follow-up.

PR #2070, opened 2026-08-20
`General`: Adopt the shared `getStudentName` util across the clients

### LINKED ISSUE #1988: Adopt the shared getStudentName util

rappm:
## Summary

Replaces every inline `` `${firstName} ${lastName}` `` student display-name concatenation across the
clients with the shared `getStudentName` util now exported by `@tumaet/prompt-ui-components`.

## Details

- Bumps `@tumaet/prompt-shared-state` and `@tumaet/prompt-ui-components` from 1.1.1 / 1.1.6 to
  1.2.0 (the release that ships `getStudentName`). Both packages are bumped together so the
  dependency commit is identical to the one in the sibling cleanup branches and the lockfile merges
  cleanly.
- `core`: sidebar avatar, breadcrumbs, application details navigation buttons, application
  assessment comment author, student table full-name accessor, instructor note create form.
- `assessment_component`: assessment header and profile, assessment form peer badges, assessment
  completion and action item panel author, evaluation overview targets, tutor evaluation page and
  results page, participant results and overview pages, print report title, feedback item row.
- `interview_component`: interview card self-assignment, student card, interview times dialog,
  schedule management assignments and student picker.
- `self_team_allocation_component`: team card members and tutors, tutor import dialog.
- `team_allocation_component`: team allocation page members and tutors, tutor selection, student
  detail dialog.
- `core/.../interfaces/StaffMember.ts`: `staffMemberFullName` reimplemented the same
  filter/join logic by hand, so it now delegates to the shared util and keeps only its `username`
  fallback.

### Behavior differences worth knowing

The shared util is not a literal drop-in. It is
`[firstName, lastName].map((p) => p?.trim()).filter(Boolean).join(' ')`, so compared to the old
template literal:

| Input | Before | After |
| --- | --- | --- |
| `{ firstName: 'Ada', lastName: 'Lovelace' }` | `Ada Lovelace` | `Ada Lovelace` |
| missing `firstName` | `undefined Lovelace` | `Lovelace` |
| missing `lastName` | `Ada undefined` | `Ada` |
| both missing | `undefined undefined` | `` (empty string) |
| `{ firstName: ' Ada ' }` | `' Ada  '` | `Ada` |

For the normal case (both names present, untrimmed data not expected from Keycloak or the core API)
the output is byte-identical, which is why the sort key in `studentTableColumns` (`fullname`
accessor), the CSV-free print report titles and the persisted author strings in
`AssessmentCard` / `AssessmentCompletion` / `ActionItemPanel` / `InterviewCard` stay compatible with
existing data. The only reachable change is that partially populated records now render a clean
name instead of leaking `undefined` into the UI.

Two related tweaks fall out of that:

- `NavAvatar` had `` `${user?.firstName} ${user?.lastName}` || ' Unknown User' ``. A template
  literal is always truthy, so the fallback was dead code. It is now reachable, and the stray
  leading space was dropped.
- `AssessmentCompletion` and `ActionItemPanel` keep their explicit `user ? ... : 'Unknown User'`
  guard so the persisted author value is unchanged for signed-in users.

### On `getTeamMemberName`

`assessment_component/.../utils/getTeamMemberName.ts` is kept rather than deleted. Its value is not
the formatting (that part now calls `getStudentName`) but the lookup of a member across all teams by
course participation ID, which four call sites share. Inlining that loop four times to remove the
file would add duplication, not remove it. It had no tests, so nothing to port.

## Reason / Link to issue

Closes #1988. Unblocked by prompt-edu/prompt-lib#234, which shipped the shared util in 1.2.0.

## How to Test

1. `cd clients && yarn install --immutable && yarn build`
2. Start the stack and open a course: participant lists, assessment and evaluation pages, interview
   schedule, and both team allocation phases should all render names exactly as before.
3. The management student table's "Full Name" column still sorts and filters as before.

## PR Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved name display consistency across assessments, evaluations, participant management, interviews, and team allocation views.
  * Standardized formatting for students, tutors, instructors, interview participants, and feedback authors.
  * Preserved appropriate fallbacks such as usernames, email addresses, or “Unknown User” when names are unavailable.
  * Improved avatar initials handling by ignoring empty name parts.
* **Style**
  * Cleaned up minor formatting inconsistencies in student detail displays.


clients/core/src/managementConsole/layout/Sidebar/CourseSwitchSidebar/components/NavAvatar.tsx
```diff
@@ -8,9 +8,9 @@ interface NavAvatarProps {
 export function NavAvatar({ avatarOnly = false }: NavAvatarProps) {
   const { user } = useAuthStore()
 
-  const userName = `${user?.firstName} ${user?.lastName}` || ' Unknown User'
+  const userName = getStudentName(user ?? {}) || 'Unknown User'
   const userEmail = user?.email || 'Unknown Email'
-  const initials = `${user?.firstName.charAt(0)}${user?.lastName.charAt(0)}` || '??'
+  const initials = `${user?.firstName?.charAt(0) ?? ''}${user?.lastName?.charAt(0) ?? ''}` || '??'
```
mathildeshagl: This line has the identical bug you just fixed two lines above, and the fix on line 11 makes the mismatch visible.

```ts
const initials = `${user?.firstName.charAt(0)}${user?.lastName.charAt(0)}` || '??'
```

Two problems:

1. `|| '??'` is still dead code, for exactly the reason you give for `|| ' Unknown User'`: a template literal is always a non-empty string here, so the fallback is unreachable.
2. `user?.firstName.charAt(0)` guards `user` but not `firstName`. The optional chain short-circuits the whole expression only when `user` itself is nullish; a `user` present with an undefined `firstName` throws on `.charAt`. That is precisely the partially-populated record this PR says now renders cleanly.

Net effect after this PR: for a user missing a first name, the name renders as `Lovelace` while the avatar right next to it either throws or shows `undefinedL`. The name got better and the initials stayed broken.

Something like:

```ts
const initials =
  [user?.firstName, user?.lastName]
    .map((part) => part?.trim()?.charAt(0))
    .filter(Boolean)
    .join('') || '??'
```

which also makes `'??'` reachable, matches the shared util's shape, and degrades to a single initial instead of nothing when only one name is present.

mathildeshagl (dismissed):
Approving. The behaviour-difference table is the part of this description that earns its keep, and it is accurate: I checked `getStudentName` in the published `@tumaet/prompt-ui-components@1.2.0` and it is exactly the `map(trim).filter(Boolean).join(' ')` you document, so every row of that table is right.

The coverage is also complete rather than mostly complete. After this change the only remaining `firstName`/`lastName` references in `clients/` are genuinely not display names: the `${firstName}-${lastName}` fallback keys in `EvaluationOverviewPage`, single-name table cells, and the `{{firstName}}` mail-template placeholders. Nothing was missed.

Two things I checked because they looked risky and turned out fine:

- The multi-line JSX form `{a} {b}` collapses to `a + ' ' + b` after JSX trims the newline-indented whitespace, so those substitutions really are output-identical, not just close.
- `staffMemberFullName` gets slightly better rather than merely equivalent: a whitespace-only name used to produce a truthy `' '` and win over the `username` fallback; it now falls through correctly.

Agreed on keeping `getTeamMemberName`. The lookup across teams by course participation ID is the actual content of that file, and inlining it four times to delete it would be a net loss.

One finding, left as an inline comment on `NavAvatar.tsx`: the same dead-fallback bug you correctly fix on the `userName` line is still present two lines above on `initials`.

Lower priority, both on lines this PR touches, both pre-existing:

- `AssessmentCard.tsx` line 62: `author` is a persisted comment field, and `comment.author === author` is what gates the delete button. Comparing identity by display name means two staff members with the same name can delete each other's comments. Not made worse here, but that is the line being edited.
- `AssessmentCard.tsx` still has a stray `console.log(filteredComments)` in `handleDeleteComment`, a few lines below. Free to remove while you are in the file.

clients/core/src/managementConsole/layout/Sidebar/CourseSwitchSidebar/components/NavAvatar.tsx
rappm: Adopted your shape for the initials, including the trim, so whitespace-only names no longer produce a blank initial and a single name degrades to one letter.

Note that `55dd36d6` (pushed shortly before your review) had already changed the line to `user?.firstName?.charAt(0) ?? ""`, so the throw was gone and `|| "??"` was reachable again. The trim handling was the part still missing.

PR #2071, opened 2026-08-20
`Module Federation`: Extract shared remote scaffolding

### LINKED ISSUE #1981: Extract shared Module Federation remote scaffolding
## Summary

Extract the shared Module Federation remote scaffolding so each phase stops copying it. This is
monorepo-internal (no external library); the scaffold covers the build config factory and the runtime
mount boilerplate, which together are the single "MF remote scaffold" deliverable.

## What is duplicated (verified, 7 components)

- `rspack.config.mjs` - 7 near-identical files (~110 lines each), differing only by `COMPONENT_NAME`,
  `COMPONENT_DEV_PORT`, and assessment's extra `@hookform/resolvers` alias.
- `index.js` + `bootstrap.tsx` + `App.tsx` (the "Please load this component from inside the Prompt
  Core!" notice) + `provide/index.ts` - identical except the root element id and app title.
  - `clients/assessment_component/src/bootstrap.tsx:5` vs `clients/matching_component/src/bootstrap.tsx`

## Work

- Add a `createRspackConfig({ name, port })` factory (with an options hook for the assessment alias)
  so each config collapses to a few lines.
- Add a `mountRemote(rootId, App)` helper and a generic standalone-notice component.

rappm:
## Summary

Every phase remote carried its own copy of the Module Federation scaffolding: seven
`rspack.config.mjs` files of ~110 lines that differed only in a name, a port, and one alias, plus a
`bootstrap.tsx` and an `App.tsx` per component that did the same two things. This extracts that
scaffolding into `clients/shared/` so a remote declares only what is actually specific to it.

## Details

**Where the shared code lives: `clients/shared/`, deliberately not a yarn workspace.**

```text
clients/shared/rspack/federatedDependencies.mjs   the singleton share scope
clients/shared/rspack/createRspackConfig.mjs      rspack config factory for a remote
clients/shared/runtime/mountRemote.tsx            React root mount for the standalone page
clients/shared/runtime/StandaloneNotice.tsx       the notice that page renders
```

A workspace package was the obvious alternative and was rejected: the runtime files are TypeScript
sources, so consuming them through `node_modules` would put them behind every component's
`exclude: /node_modules/` swc-loader rule and they would ship untranspiled. A plain directory
consumed through relative imports compiles inside each component's own rspack and TypeScript setup,
and needs no dependency, lockfile, workspace, or Dockerfile change. It is already present in every
build context: `clients/Dockerfile` copies all of `clients/` into the `prompt-clients-base` image at
`/app`, and each component image is `FROM` that base, so `/app/shared` exists before `yarn build`
runs in `/app/<component>`.

The build-time half sits under `rspack/` rather than the more obvious `build/` because
`clients/.gitignore` and `clients/.dockerignore` both ignore `build` at any depth. The first version
of this branch used `build/`, and the effect was instructive: every local checkout worked while the
factory was never committed and never reached the base image, so CI failed on the client builds and
the e2e stack failed with `MODULE_NOT_FOUND` on `/app/core/rspack.config.mjs`. Worth knowing before
adding anything else called `build` under `clients/`.

**The build factory.** Each remote config is now:

```js
export default createRspackConfig({
  name: 'example_component',
  port: 3001,
  configUrl: import.meta.url,
})
```

`configUrl` is what keeps the factory faithful. The old configs used `__dirname` for
`devServer.static` and `output.path` while leaving `context` unset, so `entry`, the HTML template,
and the `public/` copy stayed relative to the working directory. Moving the file would have silently
repointed the first group at the shared directory and, worse, changed `output.uniqueName`, which
rspack derives from the `package.json` nearest `context`. The factory therefore resolves the
component directory from the caller's `import.meta.url` and never sets `context`.

**The assessment alias** goes through `resolveAlias: (componentDir) => ({ … })`. It is a callback
rather than a plain object because the two `@hookform/resolvers` paths are resolved against the
component directory, which only the factory knows.

**Share scope.** `federatedDependencies.mjs` is imported by `clients/core` *and* by the factory, so
host and remotes now read the same five singletons (`react`, `react-dom`, `react-router-dom`,
`@tanstack/react-query`, `@tumaet/prompt-shared-state`) from one place and cannot drift apart. This
is the only part of core's config that changed. Core keeps its own config otherwise: it is the host,
and its `remotes`, `publicPath`, css rules, aliases, `optimization`, and compression plugin have
nothing in common with a remote.

**The runtime scaffolding** replaces `bootstrap.tsx` + `App.tsx` per component with a two-line
`bootstrap.tsx`. `App.tsx` is deleted; it was never exposed to core (a phase is mounted through
`./routes`, never through a root component) and was only imported by its own bootstrap.

**Two latent bugs surfaced and are fixed.** `mountRemote` throws on a missing root element instead
of casting `null` through `unknown`, which exposed that the assessment bootstrap looked for
`assessment-root` while its `template.html` declared `template-root`, so its standalone page rendered
nothing at all. Matching, team allocation, and self team allocation had likewise inherited a root id
from whichever component they were copied from (`interview-root`, `template-root`); each now uses its
own.

**Tooling that referenced the removed constants**, all updated: `scripts/new-course-phase.sh` and
`template-repository/init.sh` rewrote `COMPONENT_DEV_PORT`, which no longer exists; the
`module-federation-remote` skill, the `module-federation/remotes` rule, `new_course_phase.md`, and
the two component readmes documented the old per-component constants. `quality-clients.yml` lints
one module directory per matrix entry, so `clients/shared` would have gone unlinted; the lint step
now includes it. `generator-smoke.yml` also triggers on `clients/shared/**`, since the config it
generates depends on it.

`template-repository/` stages an external repository that by definition cannot import from this
monorepo. That migration has not been executed, so the runbook now carries the step to vendor
`clients/shared` into the template and the two import depths to shorten. `MIGRATION.md` is the only
place this is recorded, because the staged `client/` directory itself is created from
`clients/example_component` at migration time.

One unrelated change rode along: the repo's `trailing-whitespace` pre-commit hook normalized the
markdown hard breaks in `clients/matching_component/readme` when that file was touched.

### Deliberately not included

- `src/index.js` and `src/provide/index.ts` are named in the issue as duplicated, but they are not.
  Each is one to three lines of genuinely per-component content: assessment imports `./loadStyles`
  where the others import `./styles.css`, certificate has a `src/provide.tsx` file instead of a
  directory, example's `provide/index.ts` is empty, and the rest re-export a component-specific
  `StudentDetail`. There is nothing to factor out of them.
- The two external remotes (`intro_course_developer_component`, `github_challenge_component`) live
  in their own repositories and are untouched.

## Reason / Link to issue

Closes #1981

## How to Test

The risk here is that a config or share-scope change breaks the app at runtime rather than at build
time, so this was verified by comparing artifacts against `main` rather than by typecheck alone.

1. **Resolved configs are byte-identical.** A script imports each `rspack.config.mjs` with the same
   working directory rspack uses, calls it for `NODE_ENV=development` and `=production`, and
   serializes the result (regexes and plugin option objects included). All seven remotes produce
   output identical to `main` in both modes. Core's differs only in the `?${Date.now()}`
   cache-buster; with that normalized it is identical too.
2. **Built artifacts match.** All eight components build clean (`yarn --cwd <c> build`, caches
   cleared on both sides). Six of the seven `remoteEntry.js` files are byte-identical to `main`;
   assessment's differs in exactly one chunk contenthash reference. Per component the only changed
   assets are the standalone-page chunk, the entry chunk, and the two HTML files, which is precisely
   the surface this PR edits.
3. **Core is a no-op.** Its `main` chunk is byte-identical. Its runtime chunk differs only in the
   MF module ids, which hash the timestamped remote URLs and therefore already differ between two
   builds of an unmodified tree; normalizing the timestamp and the ids makes base, new, and a second
   new build all identical.
4. `yarn tsc --noEmit` per component, `make lint`, and CI's per-module `yarn biome check "<module>"
   shared` all pass with the same warning count as `main` and no errors.
5. **The generator still works.** `scripts/new-course-phase.sh ci_smoke 3990 8990 5490` produces a
   component whose factory call, root element id, and notice title are all renamed consistently,
   passes the smoke workflow's `! grep -rin 'example'` assertion, and typechecks, lints, and builds
   a `remoteEntry.js` with the full share scope. The `scaffold` job covers this in CI.
6. **Every e2e shard is green**, including the phase-module shards that load real remotes into the
   host, plus a full local `make test-e2e` run.

Manual check for a reviewer: `cd clients && yarn dev`, open <http://localhost:3000>, and confirm a
course with several phase types still lazy-loads each remote. Then open a remote's own port directly
(e.g. <http://localhost:3007>) and confirm the standalone notice renders, which is the part that was
broken for assessment before this PR.

## PR Checklist


mathildeshagl (approved):
Approving. This is the best-verified PR of the four, and the description is doing real work rather than narrating the diff.

I reproduced your central check independently instead of trusting it: worktrees for the merge base and this branch, each `rspack.config.mjs` imported with the working directory rspack uses, called for both `NODE_ENV` values, and the result serialized including regex sources and plugin option objects.

- All seven remotes: resolved configs byte-identical in both development and production.
- Core: the only difference is the `?${Date.now()}` cache-buster on the remote URLs. Nothing else, in either mode.

Other claims I checked rather than assumed:

- `exposes` was already the same three entries in all seven configs on main, so the factory hardcoding them loses nothing.
- The assessment bug is real: main had `assessment-root` in the bootstrap against `template-root` in `template.html`, so that standalone page rendered nothing at all. The other three inherited root ids are now self-consistent.
- `clients/shared` is tracked and not ignored. And `build/` genuinely would have been swallowed at any depth by both `clients/.gitignore` and `clients/.dockerignore`, so the `rspack/` naming is load-bearing rather than cosmetic. Thanks for writing that story down in the description; it is the kind of thing that costs the next person an afternoon.
- `rootDir: ".."` in the component tsconfigs already covers `../../shared/...`, so the relative imports need no tsconfig change.
- Docker: the base image's `COPY . ./` from the `clients/` context includes `shared/`, and each component image is `FROM` that base and only overlays its own directory, so `/app/shared` is present at `yarn build` time.

The `configUrl` reasoning is the part I'd have most likely got wrong myself. Deriving the component directory from the caller while leaving `context` unset, so `entry`, the HTML template, the `public/` copy and `output.uniqueName` all keep resolving exactly as before, is subtle and the byte-identical configs confirm you got it right.

I did go looking for one failure mode and it is not there, so recording it as checked: since `shared` now reaches a component build only through the base image, a stale base would silently ship old scaffolding. It cannot happen in CI, because `clients-base` runs unconditionally on every run and every component job consumes `IMAGE_TAG` from that same run. And `dev.yml` uses `paths-ignore` rather than `paths`, so a `clients/shared`-only change still triggers the whole matrix.

Small things, none blocking:

- The `resolveAlias` callback exists only so its one consumer can resolve against `componentDir`. On Node 24 that component could use `import.meta.dirname` and pass a plain object, which removes the callback indirection from the shared factory entirely. The same trick would let `configUrl` become `dirname: import.meta.dirname` and drop the `fileURLToPath` hop.
- `yarn biome check "<module>" shared` lints `shared` once per matrix entry, so eight times. Redundant, but `shared` cannot be a matrix entry of its own since it has no tsconfig or build script, so this looks like the right trade rather than an oversight.
- Worth a line in `clients/shared/readme.md`: building a single component image locally against a stale `prompt-clients-base:latest` now silently uses old shared code. That was impossible before, when all of a component's build inputs came from its own context.

One cross-PR note rather than a change request: #1939 adds an eighth remote with its own copy of exactly this scaffolding, and it does not know about `clients/shared`. Whichever of the two lands second needs a follow-up commit. Details on that PR.

PR #2085, opened 2026-08-23
`Core`: Fix the overlapping search icon and spacing in the course card view

### LINKED ISSUE #2011: Courses page is not full width in the card view
### What happened?

Or at least the cards grid itself.

It caps at 3 cards per row

Image

Normally, we say our pages should be full width, so we could go up

Although
1) this is only noticable on bigger screens (on a 16", this isn't really noticable, on a 27" f.e., it is)
2) It doesn't look too bad in my opinion anyways

so unsure if worth fixing / keep as feature

(For the table: it works as expected.)

### Steps to reproduce

1. Open the courses page at /management/courses
2. Ensure you have a window that's wide enough.
3. Notice the grid stays centered at a certain point.

### Expected behavior

As per usual visual rules prompt pages are supposed to be full width and not crop in the center so to be able to make use of the full screen

### Environment

tested locally 

ProductName:            macOS
ProductVersion:         26.5
BuildVersion:           25F71

With Chrome 150.0.7871.49 (Official Build) (arm64)

and on prompt-dev

### Logs, screenshots, or recordings

Image

### Additional context

_No response_

### LINKED ISSUE #2084: `Core`: Search icon overlaps the placeholder in the course card view
### What happened?

On the course overview pages, the course card view renders the search field with the magnifying glass icon on top of the placeholder text. In Safari the icon and the text "Search courses..." sit on the same spot, so both are unreadable.

The card view also adds a large empty band between the page header ("Courses" plus the card/table toggle) and the search field, which the table view does not have.

### Steps to reproduce

1. Log in and open `/management/general` (Courses).
2. Make sure the card view is selected in the view toggle.
3. Observe the search field: the magnifying glass overlaps the placeholder text, and there is a wide gap above the field.

### Expected behavior

The placeholder text starts to the right of the magnifying glass, and the search field follows the page header with the same spacing the rest of the page uses.

### Environment

PROMPT dev deployment, Safari on macOS. Also affects any browser that keeps the native `searchfield` appearance.

### Additional context

The field is an `<Input type='search'>` with a `pl-9` utility for the icon. Safari renders `input[type=search]` with the native `-webkit-appearance: searchfield`, which ignores custom horizontal padding, so `pl-9` has no effect and the text falls back under the icon.

The spacing comes from the `container mx-auto px-4 py-8` wrapper in `CourseCards`, which the table view does not use. That same wrapper is the reason the card view is not full width (#2011).

rappm:
## Summary

Fixes the course search field in the card view: the magnifying glass no longer sits on top of the placeholder, and the field now follows the page header with the same spacing the rest of the page uses.

## Details

- `Input` is rendered with `type='search'`, which Safari draws with the native `-webkit-appearance: searchfield`. That appearance ignores custom horizontal padding, so the `pl-9` that reserves room for the icon had no effect and the placeholder rendered underneath it. Adding `appearance-none` makes the padding apply again. The `type='search'` (and with it the `searchbox` role the e2e suite queries) is kept.
- `CourseCards` wrapped itself in `container mx-auto px-4 py-8`. Its `py-8` stacked on top of the page's own `gap-6`, leaving a wide empty band under the "Courses" heading, and `container mx-auto px-4` kept the card grid narrower than the table view of the same page. Replaced with `flex flex-col gap-6`, matching the pages that render it and the table view next to it.

## Reason / Link to issue

Closes #2084
Closes #2011

## How to Test

1. Open `/management/general` (Courses) and select the card view.
2. The placeholder text starts to the right of the magnifying glass, in Safari as well as Chrome/Firefox.
3. The search field sits directly under the header with the normal page spacing, and the card grid uses the full page width like the table view does.
4. Repeat on the Template Courses and Archived Courses pages, which render the same component.

## PR Checklist

### DIFF
--- a/clients/core/src/managementConsole/shared/components/CourseCard/CourseCards.tsx
+++ b/clients/core/src/managementConsole/shared/components/CourseCard/CourseCards.tsx
@@ -21,8 +21,8 @@ export const CourseCards = ({ courses }: CourseCardsProps) => {
   }, [courses, query])
 
   return (
-    <div className='container mx-auto px-4 py-8'>
-      <div className='relative mb-8 max-w-md'>
+    <div className='flex flex-col gap-6'>
+      <div className='relative max-w-md'>
         <Search className='pointer-events-none absolute left-3 top-1/2 h-4 w-4 -translate-y-1/2 text-muted-foreground' />
         <Input
           type='search'
--- a/clients/core/src/managementConsole/shared/components/CourseCard/CourseCards.tsx
+++ b/clients/core/src/managementConsole/shared/components/CourseCard/CourseCards.tsx
@@ -30,7 +30,7 @@ export const CourseCards = ({ courses }: CourseCardsProps) => {
           placeholder='Search courses...'
           value={search}
           onChange={(event) => setSearch(event.target.value)}
-          className='pl-9'
+          className='appearance-none pl-9'
         />
       </div>
 


rappm:
Superseded by #2087, which carries the corrected diagnosis: the icon overlap comes from remote stylesheets overriding the host's utilities, not from the browser's native search-field appearance.

PR #2105, opened 2026-08-23
`Module Federation`: Build the Tailwind stylesheet once in core

### LINKED ISSUE #2086: `Core`: Remote stylesheets override host utility classes because each bundle ships its own Tailwind build
### What happened?

Core and every Module Federation remote build their own complete Tailwind stylesheet and inject it with `style-loader`. A remote's `<style>` element is appended to `<head>` when the remote loads, which is after core's. From that point on, every utility the remote's stylesheet also contains overrides core's copy of the same utility, because the rules have equal specificity and the remote's is last.

That silently breaks any place where core overrides a base class of a shared `@tumaet/prompt-ui-components` component through `className`, whenever the override and the base class are different utilities for the same property. The concrete case found in #2084: `<Input className='pl-9'>` renders with `padding-left: 12px`, because the remotes ship `.px-3` (part of the `Input` base classes) but never `.pl-9`.

Evidence from the deployed bundles (v2.16.2):

- `main.*.js` contains `.px-3{padding-inline:...}` at byte 42279 and `.pl-9{padding-left:...}` at byte 44065, so the host build itself is ordered correctly.
- `assessment/main.*.js` contains `.px-3{padding-inline:...}` and no `.pl-9` anywhere.

### Steps to reproduce

1. Log in and open a course phase served by a remote, so its stylesheet is injected.
2. Navigate to a core page that overrides a shared component's base utility, e.g. `/management/general` in card view (`Input` with `pl-9`) or the "Add user" dialog (`Input` with `pl-10`).
3. Inspect the input: `padding-left` is `12px`, and the `.px-3` rule winning the cascade comes from the remote's injected `<style>`.

### Expected behavior

A utility class applied by the host on a shared component wins over that component's base classes, regardless of which remotes happen to be loaded.

### Additional context

Known affected call sites today:

- `clients/core/src/managementConsole/shared/components/CourseCard/CourseCards.tsx` (fixed pointwise in #2084)
- `clients/core/src/managementConsole/courseUserManagement/components/AddUserDialog.tsx`
- `clients/team_allocation_component/src/team_allocation/pages/TeamAllocation/components/TutorSearchStudents.tsx`

Marking individual utilities important is a workaround, not a fix. Directions worth evaluating:

- Emit the shared design-system utilities once from the host and stop remotes from shipping a duplicate base layer.
- Wrap the generated utilities in an explicit `@layer` so duplicated definitions do not depend on injection order.
- Control the insertion point of `style-loader` so remote styles are inserted before the host's.

rappm:
## Summary

`clients/core` now builds the only Tailwind stylesheet in the document and scans every
micro-frontend for it, and the eight phase components stop compiling Tailwind altogether. This
removes the cascade collision from #2086, where a remote's injected stylesheet overrode core's
utilities, and it also fixes the mirror-image bug where a remote's utilities were never emitted at
all.

## Details

Two problems, one root cause.

**Remotes overriding core.** `assessment_component` and `presentation_component` imported their
`styles.css` from their federated entry points, so `style-loader` appended a second complete
Tailwind build to `<head>` after core's. Both sheets carry the same globally named utilities, so the
later one won every equal-specificity duel: on `<Input className='pl-9'>` the remote's `.px-3` beat
core's `.pl-9`, and the course search icon overlapped the placeholder.

**Remotes with no styles at all.** The other six remotes only imported `styles.css` from
`src/index.js`, the standalone dev entry that core never loads, so inside the shell they silently
depended on core's utility vocabulary. Core did not scan them, so every utility they used that core
did not use was missing: 49 of them in `team_allocation_component` alone, including `cursor-grab`,
`border-l-4`, `h-[320px]`, `h-[420px]` and `max-w-[525px]`, plus 15 in `interview_component` and 10
in `self_team_allocation_component`.

Ordering the two sheets cannot fix the first problem. A `@layer` order or a `style-loader` insertion
anchor can only pick one winner for every element, so making the host win means a remote can no
longer override a shared component's base class either, and reversing it restores #2086. Being right
in both directions needs a single sheet.

Changes:

- `clients/core/tailwind.config.js` scans `../*_component/{src,routes,sidebar}` on top of its own
  sources and the shared UI library. Its theme was already a strict superset of the library preset
  the remotes used, so nothing had to be merged.
- Every `*_component` loses `src/styles.css`, `tailwind.config.js` and its CSS import.
  `assessment_component/src/loadStyles.js` is gone; its `@media print` rules move to a plain
  `src/print.css` that carries no Tailwind directives and stays imported from `routes/`. The
  standalone dev pages render nothing but a "load this from core" notice, so no workflow is lost,
  and with the entry points deleted the invariant is structural rather than a rule to remember.
- `pl-9!` in `CourseCards.tsx` goes back to `pl-9`, which is what the fix has to prove.
- `scripts/check-remote-styles.sh` fails CI if a component reintroduces a `tailwind.config.*` or any
  `@tailwind`, `@config`, `@source`, `@apply` or `tailwindcss` import in its CSS. It runs in its own
  job so it does not repeat nine times over the quality matrix. Plain CSS stays allowed.
- The e2e assessment smoke spec gains a cascade assertion, and the rule is written down in the
  Module Federation and shadcn rules, `clients/shared/readme.md`, and the client guide.

One consequence worth knowing: because core scans its siblings, its stylesheet is only complete for
the commit it was built from. CI already guarantees that (every client image is built from the same
`clients-base` tag), but a deployment that mixes per-service image tags across clients could leave
utilities out. The new-phase guide records the matching caveat for out-of-repo phases, which core
cannot scan at all.

## Reason / Link to issue

Closes #2086.

## How to Test

1. `cd clients && yarn --cwd core build`. `grep -c 'cursor-grab{' core/build/main.*.js` is 1, so a
   utility only `team_allocation_component` uses is now in core's sheet.
2. `yarn --cwd assessment_component build`, then
   `! grep -rl 'px-3{' assessment_component/build` passes: no remote emits utilities any more.
3. `./scripts/check-remote-styles.sh` passes, and fails once you add
   `@import 'tailwindcss';` to any component stylesheet.
4. `make db && make server && make clients`, log in, open an assessment phase so the remote loads,
   then use the sidebar home button to go back to the courses list without reloading. The search
   field keeps its 36px left padding and the icon no longer sits on the placeholder.
5. On the same run, open a team allocation phase: the drag handles get `cursor-grab` and the panels
   get their `h-[320px]` / `h-[420px]` heights, none of which rendered before.
6. `make test-e2e-shard SHARD=assessment` covers steps 4 in CI.

## Screenshots

The visual change is the one #2087 patched with an important modifier: in the card view of
`/management/general`, the course search placeholder starts clear of the magnifier icon. It now
holds without the modifier, after any remote has been loaded.

## PR Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented component styles from overriding shared application styles, preserving consistent layouts and utility spacing across screens.
  * Improved navigation and print styling behavior in assessment and other course workflows.

* **Quality Improvements**
  * Added automated checks to detect conflicting component-level Tailwind styles.
  * Added regression coverage confirming shell styling remains intact when switching between course areas.

* **Documentation**
  * Clarified styling guidelines and requirements for component development and external course phases.


rappm:
@magkue ich glaube das ist das was du mal meintest, vielleicht könntest du da noch einen Blick drauf werfen

mathildeshagl (approved):

Issue #2021, opened 2026-08-08
Per-PR preview environments on Kubernetes with a fresh Keycloak and seeded test data


rappm:
### Problem or opportunity

Every pull request currently deploys to one shared dev VM. `deploy-dev` in `.github/workflows/dev.yml` runs on every `pull_request` and every push to `main` with no gating condition, so two open PRs overwrite each other's deployment. #1314 ("Prompts shown PR number does not match the actual pr that was deployed") was a symptom of exactly this contention.

The practical consequences:

- A reviewer cannot click through a PR's changes in a running instance without first checking whether someone else's PR is currently occupying the dev VM.
- Testing anything that depends on state (a migration, a new course phase, a Keycloak role change) means mutating a shared database that the next PR's deploy inherits.
- There is no way to test the deployment path itself per change.

The Helm chart added in #1712 makes isolated per-PR environments feasible for the first time: the chart is namespaced, every stateful dependency is pluggable, and each phase can be toggled independently.

### Desired outcome

Opening a pull request (or applying a label to it, to control cost) produces a fully isolated environment reachable at a per-PR hostname, for example `pr-1234.prompt-ci.<domain>`:

- Its own PostgreSQL, object storage, and Keycloak realm. Nothing shared with another PR or with dev/prod.
- A known set of test users covering every role, so a reviewer can log in as a lecturer or a student immediately.
- Deterministic seeded course data, so screenshots and reproduction steps are stable across environments.
- Images built from that PR's commit.

Closing the pull request removes the environment completely, with no leaked PersistentVolumes.

### Proposed solution

Most of the required pieces already exist in the repository.

| Asset | Why it matters |
| --- | --- |
| `e2e/keycloak/realm.json` | Realm `prompt` with 8 users covering every role (admin, lecturer, course-lecturer, course-editor, student, student2, student-passkey) |
| `e2e/seed/e2e_seed.sql` | Deterministic courses, students, and course-phase types |
| Pinned client UUIDs | `prompt-server` is `a584ca61-fa83-4e95-98b6-c5f3157ae4b4` in the realm file, so `KEYCLOAK_ID_OF_CLIENT` is known **before** install |
| `dev.yml` outputs | `build-clients` and `build-servers` already expose a per-component image tag for the PR's commit |

[...]

PR #2108, opened 2026-08-26
`Seeding`: Add a full-course seed for every service database

### LINKED ISSUE #1564: Create Prompt Seeding Script
Create seeding data for the available dbs in this repository. 
This should include a full course example including data in all course phases, settings etc.

rappm:
## Summary

`make seed` now loads one fully populated demo course into all eight service
databases, so a fresh checkout stops being eight empty databases you have to click
your way out of. The e2e stack uses the same seed instead of its own private dump.

## Details

**A shared, data-only seed (`seed/`).** One SQL file per service database, applied by
`scripts/seed.sh`. Data only, because the servers own the schema: core shells out to
`migrate` at startup and the phase services call `sdkUtils.RunMigrations`. The runner
therefore waits per database until `schema_migrations` reaches the highest numbered
migration in that service's `db/migration` directory and is not dirty. The expected
version is read from the source tree at run time, so there is no version file to keep
in sync.

**Course phase types are no longer seeded.** `initializeTypes.go` creates them with
random UUIDs and *skips their provided/required DTO descriptors* whenever the type
row already exists. That is why the old e2e dump had to hand-mirror
`initializeTypes.go`, and why `Interview`, `Matching` and `Team Allocation` ended up
with no DTO metadata at all. Seeded phases now resolve their type by name, so the
descriptors are the server's own, `base_url` comes out right per environment instead
of being frozen to the Docker form, and the seed can finally wire real
`participation_data_dependency_graph` edges. The runner waits for all eight type
names before inserting - `Application` is created by
`InitApplicationAdministrationModule`, separately from the other seven.

**The demo course.** `iPraktikumDemo` has a phase of every type, eight applicants
narrowing to six participants, a configured application form with answers and scores,
interview slots and scores, skills/teams/survey responses/a published allocation, a
rubric with graded participants and a self evaluation round, a presentation schedule,
a released certificate template, and a mail campaign. `Self Team Allocation` and
`Example` sit off the graph: `course_phase_graph` is UNIQUE on both endpoints so a
course's phases are a strict chain, and no real course runs both allocation
mechanisms. They still carry full data.

**Interview reviews follow their service.** Merging `main` brought in the interview
phase becoming independent (#1880), which moved score, interviewer and answers off
core's participation `restricted_data` into the interview service's own
`interview_review` table, resolved over REST. The six demo reviews moved with them
into `seed/interview.sql`, and the phase now carries the two `interviewQuestions`
their answers reference.

It is a **separate course** rather than an extension of `iPraktikumFull` because the
Playwright journeys own that course's phases - the assessment lecturer journey, for
instance, asserts `1/4 final` on its graph tail, which seeded assessments would break.

**The seed is authoritative.** Each file deletes the rows it owns and inserts them
again, so re-running reconciles drift rather than failing on a duplicate key, and a
lazily created default `course_phase_config` gets replaced instead of preserved.
Deleting a seeded student is the one destructive step - `course_participation.student_id`
is `ON DELETE CASCADE` and would take an unrelated enrollment with it - so a guard
refuses to run when a seeded student is referenced outside the seeded courses.

**e2e cutover.** The stack loaded `e2e/seed/e2e_seed.sql` through
`/docker-entrypoint-initdb.d`: a frozen copy of the core schema pinned at
`schema_migrations = 24` while core is at 28, and none of the seven phase databases
was ever seeded. A one-shot `seed` service now applies the shared files after the
servers start, and `e2e-runner` waits on `service_completed_successfully`. Every
fixture keeps its pinned UUID, so the specs are untouched apart from two consequences
of the seed no longer owning phase types: the configurator page object resolves the
palette's test ids from `GET /api/course_phase_types`, and `SEEDED_COURSE_COUNT`
accounts for the demo course.

**Keycloak.** `CreateCourseGroupsAndRoles` only runs on the course-create API path, so
a seeded course gets no roles and only PromptAdmin can see it. `keycloakConfig.json`
now carries what that path would have built for the demo course. Students are
unaffected: their access is derived in the database from `matriculation_number` and
`university_login`.

Two other things worth flagging:

- `make db` and `db-down` now cover all eight service databases. They handled only the
  core one, while `make servers` started eight services against eight databases.
- `scripts/seed.sh` is POSIX `sh`, not bash like the rest of `scripts/`: it also runs
  inside the `postgres:15.19-alpine` seed container, whose only shell is busybox ash.
  `make seed` runs it there too, so no host `psql` is needed and local development and
  CI take the same code path.

Deliberately out of scope: the `servers/*/database_dumps/` Go test fixtures (the SDK
`SetupTestDB` does not run migrations, so they hand-carry schema and drift - their own
problem), and `api-stress/lib/fixtures.py`.

## Reason / Link to issue

Closes #1564.

Also unblocks the seeded-data half of #2021, which asks for exactly this seed in a
data-only shape so load ordering stops mattering on Kubernetes.

## How to Test

```bash
make db
make servers      # the servers create the schemas
make seed
make seed         # re-runnable: reconciles, no duplicates
make seed-check   # cross-database ids resolve
```

Then `make clients` and log in as `admin`, `lecturer` or `student` (username ==
password) and open **iPraktikumDemo**: every phase renders seeded data.

Keycloak imports the realm only into an empty database, so an existing local install
needs `make db-down && rm -rf keycloak_postgres_data && make db` to pick up the demo
course roles.

The e2e suite is the end-to-end check of the cutover, one shard at a time:
`make test-e2e-shard SHARD=<name>`.

## PR Checklist

## Summary by CodeRabbit

- **New Features**
  - Added optional demo-data seeding across all service databases.
  - Added comprehensive demo courses covering application, interview, team allocation, assessment, presentation, certificate, and related workflows.
  - Added validation for cross-service seed references.
  - E2E environments now load shared test data automatically after migrations.
  - Added automatic readiness checks before E2E tests begin.
  - Client unit tests are included in the standard test command.

- **Documentation**
  - Added setup and contributor guidance for loading, validating, and maintaining seed data.


mathildeshagl (dismissed):

docker-compose.yml
```diff
@@ -738,3 +738,39 @@ services:
...
+    entrypoint: ["/seed.sh"]
+    volumes:
+      - ./seed:/seed:ro
+      - ./servers:/servers:ro
+      - ./scripts/seed.sh:/seed.sh:ro
+    environment:
+      - SEED_SQL_ROOT=/seed
+      - SEED_MIGRATIONS_ROOT=/servers
+      - SEED_DB_NAME=${DB_CORE_NAME}
+      - SEED_DB_USER=${DB_CORE_USER}
```
mathildeshagl, line 762: The seed uses the core credentials for all eight databases, while every `db-*` service uses its own `DB_<SERVICE>_*` (`docker-compose.yml:452` for assessment, same shape in `docker-compose.e2e.yml`). They all resolve to `prompt-postgres` today, so this works, but a developer who changes one service's password in `.env` gets an opaque psql auth failure from the seed rather than anything that points at the cause.

`scripts/seed.sh` already has the `SEED_<SVC>_HOST` / `SEED_<SVC>_PORT` pattern. Extending it to `SEED_<SVC>_USER` / `SEED_<SVC>_PASSWORD`, falling back to the current values, is a few lines and keeps the credential split the rest of the stack models.

docker-compose.e2e.yml
```diff
@@ -856,6 +856,58 @@ services:
   # The browser, the app's runtime config, the server, and Keycloak all address
   # each other by Docker service name on the compose network — so the realm's
   # redirect URIs (http://client-core/*) and the token issuer all line up.
+  # Applies the shared seed (seed/, the same files `make seed` uses) once every
+  # server has migrated its database. scripts/seed.sh waits per database until
+  # schema_migrations reaches the version the migration directory declares, so
+  # `service_started` is enough of a dependency here.
+  seed:
```
mathildeshagl, line 868: Now that this service seeds all eight databases, the `No seed: the phase server runs its own migrations on the empty DB and the tests create all phase data themselves` comments on the phase databases are no longer true: lines 210, 281, 349, 442, 515 and 658. Only the core `db` comment was updated.

scripts/seed-check.sh
```diff
@@ -0,0 +1,34 @@
...
+    [ "$file" = "$CORE" ] && continue
+    for id in $(grep -Eo "$CORE_OWNED" "$file" | sort -u); do
+        if ! grep -q "$id" "$CORE"; then
+            echo "seed-check: $(basename "$file") references $id, which seed/core.sql does not create" >&2
+            status=1
+        fi
+    done
+done
+
+[ "$status" -eq 0 ] && echo "seed-check: cross-database references OK"
```
mathildeshagl, line 33: Nothing runs this automatically, so it only fires when someone remembers `make seed-check`. Given that the point is catching a cross-database id typo that otherwise surfaces as an empty screen, please wire it into CI: `quality-servers.yml` looks like the natural home.

seed/assessment.sql
```diff
@@ -0,0 +1,143 @@
...
+
+-- Assessment window closed, results released, self-evaluation enabled and bound
+-- to the seed's own self-evaluation schema. peer_evaluation_schema and
+-- tutor_evaluation_schema are omitted so their column DEFAULTs apply.
+INSERT INTO course_phase_config (course_phase_id, assessment_schema_id, self_evaluation_schema, start, deadline,
+                                 self_evaluation_enabled, self_evaluation_start, self_evaluation_deadline,
+                                 evaluation_results_visible, grade_suggestion_visible, action_items_visible,
+                                 results_released, grading_sheet_visible, assessment_enabled)
+    VALUES ('f0000005-0000-0000-0000-000000000005', '5a000001-0000-0000-0000-000000000001', '5a000002-0000-0000-0000-000000000002',
+            '2026-07-01 00:00:00+02', '2026-09-20 23:59:00+02',
```
mathildeshagl, line 57: The demo course's time-sensitive rows are absolute dates and some have already passed: the self-evaluation deadline on the next line is 2026-06-30, and `seed/interview.sql` schedules its slots in March 2026. The assessment deadline here goes stale in about three weeks. A fresh contributor running `make seed` therefore gets a demo course with several phases already expired, which is the opposite of what the PR is for.

The e2e suite is unaffected, since `e2e/tests/assessment/helpers.ts:119` sets its own deadlines relative to `now`, so this is demo quality rather than test flake. Either anchor the time-sensitive rows to `NOW() + interval`, or state in `seed/README.md` that the demo models a completed semester so the dates read as deliberate.

mathildeshagl (changes requested):

docker-compose.yml
rappm: Done. `scripts/seed.sh` now resolves host, port, database name, user and password per service (`SEED_<SERVICE>_HOST`, `_PORT`, `_NAME`, `_USER`, `_PASSWORD`), each falling back to the `SEED_DB_*` core values, and both compose files pass every service its own `DB_<SERVICE>_*`. Changing one service's password in `.env` now fails against that database instead of quietly reusing core's. Extended it to the database name as well, since that is split the same way.

`docs/contributor/guide/seeding.md` describes the new resolution order.

docker-compose.e2e.yml
rappm: Fixed, all six. They now say the shared seed under `seed/` is data only and is applied by the `seed` service once the phase server has migrated the database. The assessment one keeps its note that the module's own migrations create the default assessment and evaluation schemas.

scripts/seed-check.sh
rappm: Added. `quality-servers.yml` has a `seed-check` job that runs `scripts/seed-check.sh` on every PR, so it no longer depends on someone remembering `make seed-check`.

seed/assessment.sql
rappm: Took your second option: documented rather than relative.

Every date in the seed belongs to one semester, `ios2526` (2026-04-01 to 2026-09-30), and the phases are dated in the order the course runs them: application closes in March, interviews follow, allocation in April, assessment over the summer, presentation and certificate at the end. Anchoring the deadlines to `NOW()` would put them at odds with the semester tag, the course's own start and end date, and the completed data hanging off them (grades entered, results released, certificates downloaded), so the timeline would stop being coherent.

`seed/README.md` now has a timeline table with the state each phase is in, says plainly that the demo ages once summer 2026 is past, and points at moving the one window you need either in the phase's configuration screen or in the seed file before loading. `seed/assessment.sql` points at that, and its comment no longer claims the assessment window is closed, which it is not yet.

e2e/seed/e2e_seed.sql
```diff
@@ -1,1846 +0,0 @@
--- ============================================================================
```
mathildeshagl, line 1: Two things hang off this deletion.

The first is the red `core` shard above. It is worth checking against a local `make test-e2e-shard SHARD=core` rather than only the privacy or seeding paths, because that is the shard whose fixture data this replaces wholesale.

The second is #2135, which adds its `privacy-student` and `Ida Inactive` fixtures to this exact file. I test-merged the two branches:

```
CONFLICT (modify/delete): e2e/seed/e2e_seed.sql deleted in pr-2108
and modified in pr-2135
```

Whichever of the two lands second has to port its fixtures into `seed/core.sql`. Not something to fix here, but worth agreeing on an order with #2135 so it is a deliberate port rather than a conflict resolution done under merge pressure.

seed/team_allocation.sql
```diff
@@ -0,0 +1,71 @@
...
+    VALUES ('3a000001-0000-0000-0000-000000000001', 'Team Kepler', 'f0000004-0000-0000-0000-000000000004'),
+           ('3a000002-0000-0000-0000-000000000002', 'Team Hubble', 'f0000004-0000-0000-0000-000000000004');
+
+INSERT INTO skill (id, course_phase_id, name)
+    VALUES ('3b000001-0000-0000-0000-000000000001', 'f0000004-0000-0000-0000-000000000004', 'Swift'),
+           ('3b000002-0000-0000-0000-000000000002', 'f0000004-0000-0000-0000-000000000004', 'UI/UX Design'),
+           ('3b000003-0000-0000-0000-000000000003', 'f0000004-0000-0000-0000-000000000004', 'Backend');
+
+INSERT INTO survey_timeframe (course_phase_id, survey_start, survey_deadline)
+    VALUES ('f0000004-0000-0000-0000-000000000004', '2026-04-01 08:00:00', '2026-04-06 23:59:00');
```
mathildeshagl: These literals carry no offset, and #2119 converts `survey_timeframe.survey_start` / `survey_deadline` from `TIMESTAMP` to `TIMESTAMPTZ`.

Today that does not matter: the column is zone-less, so the literal is stored as the wall clock written here. Once #2119 lands it does. The `db-*` containers run `TZ=Europe/Berlin` in both `docker-compose.yml` and `docker-compose.e2e.yml`, so the seed session inherits that zone and PostgreSQL resolves a bare literal against it. `2026-04-01 08:00:00` becomes `06:00:00+00`, and the demo survey window opens and closes two hours earlier than the README's timeline table says.

It is a two-hour skew on a demo course, so not severe in itself. It is the exact defect class #2119 exists to remove, though, and it will be reintroduced silently by merge order rather than by anyone's diff. Adding `+00` to these two values now costs nothing and is correct against both the current and the post-#2119 schema.

seed/self_team_allocation.sql
```diff
@@ -0,0 +1,27 @@
...
+
+DELETE FROM team WHERE course_phase_id = 'f0000008-0000-0000-0000-000000000008';
+DELETE FROM timeframe WHERE course_phase_id = 'f0000008-0000-0000-0000-000000000008';
+
+INSERT INTO team (id, name, course_phase_id)
+    VALUES ('2a000001-0000-0000-0000-000000000001', 'Team Aurora', 'f0000008-0000-0000-0000-000000000008'),
+           ('2a000002-0000-0000-0000-000000000002', 'Team Borealis', 'f0000008-0000-0000-0000-000000000008');
+
+INSERT INTO timeframe (course_phase_id, starttime, endtime)
+    VALUES ('f0000008-0000-0000-0000-000000000008', '2026-04-07 08:00:00', '2026-04-14 23:59:00');
```
mathildeshagl: Same as the `team_allocation` seed: `timeframe.starttime` / `endtime` become `TIMESTAMPTZ` in #2119, and these bare literals will then be read as Berlin wall clocks under the seed container's session zone. `+00` on both values.

mathildeshagl (changes requested):
All four of my previous comments are addressed, and addressed properly. `scripts/seed.sh` now resolves host, port, database name, user and password per service with a core fallback, and both compose files pass each `DB_<SERVICE>_*` through. All six e2e phase-database comments are corrected. `seed-check` runs from `quality-servers.yml`, and I ran the script myself: it passes and needs no database, so it belongs in that job. `seed/README.md` documents the pinned `ios2526` timeline with a plain statement that the demo ages once summer 2026 is past. Taking the documentation option on the dates was the right call, and the reasoning about the semester tag and the completed data hanging off those deadlines holds.

So this is not a request for changes about the seed content. It is about CI.

**`e2e-tests / core` and `e2e-tests / e2e` are red.** `tests/auth/session-persistence.spec.ts` fails on the initial run and both retries, identically:

```
Error: expect(page).toHaveURL(expected) failed
```

That is not a flake, and it is not pre-existing. The same shard is `SUCCESS` on #2119, #2123 and #2135, all of which contain that spec and sit on the same base. The failure is specific to this branch, and this branch is the one that deletes all 1846 lines of `e2e/seed/e2e_seed.sql` and repoints the e2e stack at `seed/`. It also adds the demo course roles to the Keycloak realm, which is the only other change anywhere near the auth path.

I could not pin the cause: the fixture phase `d0000001` still carries `applicationEndDate` `2099-12-31` in `seed/core.sql`, so the open-application gate is not it, and the assertion that fails is the redirect to `/authenticated`, which is session restore rather than seed data. Either way it needs an explanation before this merges — the whole point of the change is that the e2e stack and local development share one seed, and right now the e2e stack does not agree.

Three more things inline, none of them blocking on their own.

seed/team_allocation.sql
rappm: Done, both values now carry `+00`.

This stopped being hypothetical while I was on it: #2119 has landed, so `survey_timeframe` is already `TIMESTAMPTZ` on main, and its migration converts existing rows with `AT TIME ZONE 'UTC'` — which confirms UTC was the intended reading of the old naive values. Against the running e2e stack:

```
SHOW timezone;         -> Europe/Berlin
'2026-04-01 08:00:00'::timestamptz     -> 2026-04-01 08:00:00+02   (06:00Z)
'2026-04-01 08:00:00+00'::timestamptz  -> 2026-04-01 10:00:00+02   (08:00Z)
```

So the bare literal would have been exactly the two-hour skew you described. The seeded row now reads `2026-04-01 10:00:00+02`, i.e. 08:00Z.

seed/self_team_allocation.sql
rappm: Done, `+00` on both. Same reasoning as the `team_allocation` thread: #2119 has landed, so `timeframe` is already `TIMESTAMPTZ`, and its migration reads the old values as UTC.

e2e/seed/e2e_seed.sql
rappm: Both points landed, and the order resolved itself: #2135 merged first, so this branch does the port rather than #2135.

I merged main in and ported its fixtures into `seed/core.sql` deliberately. It was three PRs, not just #2135 — #2133 and #2134 also touched that file:

- `iPraktikumWelcome` and its Welcome Application phase, from #2133. It had taken `c0000002`, which the demo course already owns here, so the course is `c0000003` and `WELCOME_TEXT_COURSE_ID` follows it. Phase id and the seeded `welcomeText` are byte-identical to main.
- The Certificate Student Page Text phase and Stan's participation in it, from #2134.
- `privacy-student` and `Ida Inactive`, their iPraktikum participations, and their entries in the delete and reseed-guard lists, from #2135.

On the `core` shard: running it locally was the right call, because it caught something the diff does not show. Deleting `e2e_seed.sql` also deletes the pinned `course_phase_type` rows, and core's startup init only reproduces their `{CORE_HOST}` base urls when it is not in a development environment. The e2e stack never set `ENVIRONMENT` for `server-core`, so it defaulted to development and registered every phase service as `http://localhost:<port>` — unreachable from core's own container. `deletion-approval` failed on it:

```
failed to call CPM deletion endpoint: Post "http://localhost:8084/self-team-allocation/api/privacy/data-deletion":
dial tcp [::1]:8084: connect: connection refused
```

That spec only exists because #2135 landed, so this would have gone in undetected had the order been the other way. `server-core` now runs with `ENVIRONMENT=production` in `docker-compose.e2e.yml`, which reproduces exactly the base urls the old seed pinned. `core` is 151/151 green.

seed/core.sql
```diff
@@ -0,0 +1,740 @@
...
+
+-- ─── Inter-phase data dependencies ──────────────────────────────────────────
+--
+-- The DTO descriptors belong to the phase TYPES and are created by the core
+-- server on startup, so they are resolved by (type name, dto name) rather than
+-- pinned. This is the wiring that makes a downstream phase see an upstream
+-- phase's data: without it a phase's participants table shows no resolved
+-- columns at all. A rename upstream would drop an edge silently rather than
+-- fail, so each insert asserts how many edges it wrote.
+
```
mathildeshagl, line 670: Approving — all seven of my earlier threads are addressed properly, and the CI cause you found is convincing: deleting `e2e_seed.sql` also dropped the pinned `course_phase_type` rows, and core only reproduces their `{CORE_HOST}` base urls outside a development environment, so `ENVIRONMENT=production` on `server-core` restores exactly what the old seed pinned. All twelve e2e shards plus `seed-check` are green, and the three-PR fixture port reads as deliberate.

One thing I would still like, and I am agreeing with CodeRabbit here rather than adding something new: both of these inserts resolve their DTOs by joining on `(type name, dto name)` against rows the phase services create at startup. If a DTO is renamed, or a phase type has not been created yet when the seed runs, the JOIN yields nothing and the insert succeeds with zero rows and no error. Your own comment eight lines up states the consequence — "without it a phase's participants table shows no resolved columns at all" — which is a demo course that loads, looks fine, and is quietly wrong in the one place this seed exists to get right. `scripts/seed-check.sh` validates UUID literals, so it cannot see this.

The file already establishes the idiom for exactly this at line 75, so it fits without inventing anything:

```sql
DO $$
DECLARE
    n int;
BEGIN
    SELECT count(*) INTO n FROM participation_data_dependency_graph
     WHERE from_course_phase_id IN (
        'f0000001-0000-0000-0000-000000000001',
        'f0000002-0000-0000-0000-000000000002',
        'f0000004-0000-0000-0000-000000000004');
    IF n <> 8 THEN
        RAISE EXCEPTION 'seed: expected 8 participation dependency edges, got %. A DTO name no longer resolves.', n;
    END IF;
END $$;
```

and the same for the three `phase_data_dependency_graph` edges below.

Not blocking the merge — do it here or as a follow-up, whichever you prefer. The three remaining CodeRabbit threads are docs (MD040 fence language in `seed/README.md`, the `example_component` exception in `e2e/README.md`, the migration-guidance wording) and worth one sweep.

mathildeshagl (approved):

PR #2109, opened 2026-08-26
`Testing`: Verify changes in a real host browser

mathildeshagl:
## Summary

Adds a host-browser mode to the seeded e2e stack so a change can be observed in a real browser, and rewrites the `verify` skill to drive it with the Playwright MCP browser.

## Details

The e2e stack addresses every service by compose name (`http://client-core`, `http://keycloak:8080`), because the containerized Playwright runner's Chromium resolves those via Docker DNS. A browser on the host cannot: it fails with `ERR_NAME_NOT_RESOLVED` at the login redirect, and a token minted via `localhost:18081` against the unmodified stack is rejected by the servers as an issuer mismatch. The old `verify` skill documented this as a dead end and worked around it with a throwaway driver script inside the `e2e-runner` container.

- **`e2e/docker-compose.browser.yml`** — overlay on `docker-compose.e2e.yml` that makes host browser and servers agree on one issuer, `http://localhost:18081/realms/prompt`:
  - `client-core` gets `CORE_HOST=http://localhost:18090` and `KEYCLOAK_HOST=http://localhost:18081` in its `env.js` (generated by the container entrypoint, so no rebuild is needed).
  - `server-core` gets `CORE_HOST=http://localhost:4000`, the CORS allow-list origin.
  - Every token-validating server gets `localhost:host-gateway`, so `localhost:18081` resolves to the published Keycloak port from inside its container.
  - `seaweedfs-s3` is published on 18333 with `S3_PUBLIC_ENDPOINT` pointing there, so presigned URLs handed to the browser resolve on the host.

  The e2e realm already whitelists `http://localhost:4000/*` as a redirect URI and web origin, so no realm change was needed.
- **`Makefile`** — `make verify-up` (with `SKIP_BUILD=1`) and `make verify-down`. Both reuse the existing `E2E_ENV_KEYS` unset so `e2e/.env.e2e` stays authoritative.
- **`.mcp.json`** — registers the Playwright MCP server. Enabling it per-developer stays in the gitignored `.claude/settings.local.json`, so a first launch after pulling this prompts for approval as usual.
- **`.agents/skills/verify/SKILL.md`** — rewritten around the MCP browser. The container-driver approach stays documented as a fallback for reproducing the suite's DNS-name environment.
- **`AGENTS.md`, `.gitignore`** — document the new targets; ignore the `.playwright-mcp/` scratch output.

CI is untouched: the new overlay is opt-in via the `verify-*` targets, and `make test-e2e-shard` still runs the base compose file alone.

## Reason / Link to issue

Verifying a client or core-server change at its surface meant either writing a throwaway Playwright script into the checked-in `e2e/` directory or relying on tests. Driving the real UI is now a first-class step, matching how the browser is used for verification in our other repos.

## How to Test

1. `SKIP_BUILD=1 make verify-up` (drop `SKIP_BUILD` for a cold build, ~10 min).
2. Confirm readiness: `curl -s -o /dev/null -w '%{http_code}\n' http://localhost:18090/api/hello` and the same against `http://localhost:4000`.
3. Open `http://localhost:4000/management` in a browser and log in as `lecturer` / `lecturer`. The redirect must go to `http://localhost:18081/...` with `redirect_uri=http://localhost:4000/management`.
4. Click into a phase (e.g. Assessment) and confirm the micro-frontend renders — this exercises the remote fetch, the phase server's token validation, and the `prompt-sdk` round-trip to core.
5. `make verify-down`.

Verified on this branch: login succeeded with issuer `http://localhost:18081/realms/prompt`, the core API accepted the token, the CORS preflight returned `Access-Control-Allow-Origin: http://localhost:4000`, and the assessment, interview, certificate and self-team-allocation phase APIs all returned 200 with a token and 401 without.

## PR Checklist

## Summary by CodeRabbit

- **New Features**
  - Added browser-based verification for the seeded application stack using Playwright.
  - Added streamlined commands to start and stop the verification environment.
  - Added host-accessible service endpoints, including support for browser-based file interactions.

- **Documentation**
  - Expanded verification guidance with login, screenshots, console inspection, API testing, environment overrides, and troubleshooting notes.
  - Documented the browser verification workflow and fallback testing approach.

- **Chores**
  - Excluded local Playwright tooling output from version control.
  - Added configuration for the Playwright browser automation server.

### DIFF
--- a/.agents/skills/verify/SKILL.md
+++ b/.agents/skills/verify/SKILL.md
@@ -1,43 +1,97 @@
 ---
 name: verify
-description: Run PROMPT 2.0 and observe a change in the real UI or API. Boots the seeded e2e stack (core client + server + Keycloak + Postgres + SeaweedFS) without the Playwright suite and drives it from inside the compose network. Use when verifying a client or core-server change at its surface rather than through tests.
+description: Run PROMPT 2.0 and observe a change in a real browser. Boots the seeded e2e stack (core client + server + Keycloak + Postgres + SeaweedFS) in host-browser mode and drives it with the Playwright MCP browser. Use when verifying a client or core-server change at its surface rather than through tests.
 ---
 
-# Verify a change against a running PROMPT stack
+# Verify a change in a real browser
 
-The `docker-compose.e2e.yml` stack is the fastest handle: it boots the core client, the
-core server, every phase service, Keycloak with an imported realm, and a fully seeded
-database (`e2e/seed/e2e_seed.sql`) on ports that coexist with a dev stack — client 4000,
-core API 18090, Keycloak 18081.
+The `docker-compose.e2e.yml` stack is the fastest handle: core client, core server, every
+phase service, Keycloak with an imported realm, and a fully seeded database
+(`e2e/seed/e2e_seed.sql`) on ports that coexist with a dev stack — client 4000, core API
+18090, Keycloak 18081, Mailpit 18025.
 
-## Boot the app without running tests
+`e2e/docker-compose.browser.yml` overlays it for **host-browser mode**, and
+`make verify-up` boots that combination. Drive it with the **Playwright MCP** browser
+(`mcp__playwright__*`), the same way changes are verified in FASTLANE.
 
-The Makefile exports `.env` / `.env.dev`, whose host-mode overrides outrank
-`--env-file e2e/.env.e2e`, so unset those keys (this is what `make test-e2e-*` does):
+## 1. Boot the stack
 
 ```bash
-env -u DB_CORE_HOST -u DB_CORE_PORT -u DB_CORE_USER -u DB_CORE_PASSWORD -u DB_CORE_NAME \
-    -u KEYCLOAK_HOST -u CORE_HOST -u SERVER_ADDRESS \
-  docker compose -f docker-compose.e2e.yml --env-file e2e/.env.e2e up -d client-core server-core
+make verify-up               # add SKIP_BUILD=1 to reuse images already built
 ```
 
-`depends_on` pulls in the databases, Keycloak and SeaweedFS. A cold build is ~10 min;
-afterwards startup is under a minute. Readiness:
+A cold build is ~10 min; afterwards startup is under a minute. Readiness:
 
 ```bash
 curl -s -o /dev/null -w '%{http_code}\n' http://localhost:18090/api/hello   # core API
 curl -s -o /dev/null -w '%{http_code}\n' http://localhost:4000              # core client
 ```
 
-Tear down with `make test-e2e-down` (removes volumes, so the seed is fresh next boot).
+Tear down with `make verify-down` (removes volumes, so the seed is fresh next boot).
 
-## Drive it from inside the compose network, not from the host
+## 2. Drive the browser
 
-The client's runtime config points the browser at `http://keycloak:8080`, a name only
-Docker DNS resolves — a host browser dies on `ERR_NAME_NOT_RESOLVED` at the login
-redirect, and a Keycloak token minted via `localhost:18081` is rejected by the server as
-`{"error":"Invalid token"}` (issuer mismatch). Run driver scripts in the `e2e-runner`
-container instead: it sits on the network and ships Playwright plus Chromium.
+Navigate to `http://localhost:4000/management`, log in, and interact. Seeded accounts are
+in `e2e/src/data/roles.ts` — **username equals password** (`lecturer`, `admin`, `student`,
+…). Seeded courses, phases and participants are in `e2e/src/data/constants.ts`.
+
+The login form is Keycloak's own page; log in the way the suite does
+(`e2e/src/pages/LoginPage.ts`): fill `#username` / `#password`, click `#kc-login`, then
+wait for the app shell. Sequence with the MCP tools:
+
+1. `browser_navigate` → `http://localhost:4000/management`
+2. `browser_snapshot` to get refs for the Keycloak fields
+3. `browser_type` into `#username` / `#password`, `browser_click` `#kc-login`
+4. `browser_snapshot` again once redirected back, then click through to the change
+5. `browser_take_screenshot` for the evidence, `browser_console_messages` for errors
+
+Snapshots and console logs land in the gitignored `.playwright-mcp/` on their own, but
+`browser_take_screenshot` resolves its `filename` against the repo root — pass an explicit
+`.playwright-mcp/<name>.png` or the image lands in the working tree, where only a
+`git status` catches it. Always report what you observed, and show a screenshot of the
+changed surface.
+
+## 3. Exercising real write paths
+
+A lecturer token drives any core API directly from the host, which beats hand-editing the
+seed when you need a state the fixtures don't have:
+
+```bash
+TOKEN=$(curl -s -X POST http://localhost:18081/realms/prompt/protocol/openid-connect/token \
+  -d client_id=prompt-client -d username=lecturer -d password=lecturer \
+  -d grant_type=password | python3 -c 'import sys,json;print(json.load(sys.stdin)["access_token"])')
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -194,6 +195,28 @@ test-e2e-ui: ## Interactive Playwright UI in Docker - then open http://127.0.0.1
...
+# Same seeded stack as the e2e suite, overlaid so the client, Keycloak and the
+# core API are reachable as localhost:<published port>. That lets a browser on
+# the host complete the login redirect and mint a token the servers accept.
+# Driven by the `verify` skill.
+VERIFY_COMPOSE = docker compose -f docker-compose.e2e.yml -f e2e/docker-compose.browser.yml --env-file e2e/.env.e2e
+
+verify-up: ## Boot the seeded stack for host-browser verification (SKIP_BUILD=1 reuses existing images)
+	unset $(E2E_ENV_KEYS); \
+		$(if $(SKIP_BUILD),true,$(VERIFY_COMPOSE) build); \
+		$(VERIFY_COMPOSE) up -d client-core server-core
```
rappm: The recipe chains `build` and `up -d` with `;` and no `set -e`, so a failed build is ignored and `up -d` starts the previously built images. Make takes the exit status of `up`, so it reports success and the banner below still prints the URLs.

Concretely: introduce a TypeScript error in a client component, run `make verify-up`, and the `client-core` build fails while the last good image boots. The browser then shows the old UI and you "verify" code that was never built. That matters more here than in `test-e2e`, since observing the change under build is the entire point of this target.

`test-e2e-shard` already guards exactly this with `@set -e` before `$(E2E_BUILD)`. Same fix here:

```suggestion
	set -e; unset $(E2E_ENV_KEYS); \
		$(if $(SKIP_BUILD),true,$(VERIFY_COMPOSE) build); \
		$(VERIFY_COMPOSE) up -d client-core server-core
```

.agents/skills/verify/SKILL.md
```diff
@@ -1,43 +1,97 @@
...
+## 3. Exercising real write paths
+
+A lecturer token drives any core API directly from the host, which beats hand-editing the
+seed when you need a state the fixtures don't have:
+
+```bash
+TOKEN=$(curl -s -X POST http://localhost:18081/realms/prompt/protocol/openid-connect/token \
+  -d client_id=prompt-client -d username=lecturer -d password=lecturer \
+  -d grant_type=password | python3 -c 'import sys,json;print(json.load(sys.stdin)["access_token"])')
+curl -s -H "Authorization: Bearer $TOKEN" http://localhost:18090/api/courses | head -c 400
```
rappm: This hits a route that is not registered. `servers/core/course/router.go:23-24` builds the group as `/courses` and registers `course.GET("/", ...)`, so the full path is `/api/courses/`. Gin's `RedirectTrailingSlash` is not overridden anywhere in `servers/core`, so it stays at its default of true and a GET to `/api/courses` returns a 301.

Without `-L`, the snippet prints `<a href="/api/courses/">Moved Permanently</a>` instead of course JSON, which reads like a rejected token to whoever is following the "confirm the API accepts the token" step.

```suggestion
curl -s -H "Authorization: Bearer $TOKEN" http://localhost:18090/api/courses/ | head -c 400
```

`/api/courses/self` is another option: it sits behind `authMiddleware` only, so it needs nothing but a valid token.

.mcp.json
```diff
@@ -0,0 +1,8 @@
+{
+  "mcpServers": {
+    "playwright": {
+      "command": "npx",
+      "args": ["@playwright/mcp@latest"]
```
rappm: `@latest` resolves whatever is newest on npm at every launch and runs it with the developer's full permissions, with no lockfile or integrity pin behind it.

The skill names specific tools (`browser_navigate`, `browser_snapshot`, `browser_type`, `browser_take_screenshot`, `browser_console_messages`). An upstream rename or signature change silently breaks the documented workflow for everyone who pulls, with no pinned version to bisect against. The current release is 0.0.79 and the 0.0.x line moves fast, so this is a real risk rather than a theoretical one.

```suggestion
      "args": ["@playwright/mcp@0.0.79"]
```

rappm (changes requested):
The overlay approach is the right call here, and the merged config does what the header comment claims. Rendering it with `docker compose -f docker-compose.e2e.yml -f e2e/docker-compose.browser.yml --env-file e2e/.env.e2e config` shows `server-core` picking up `CORE_HOST=http://localhost:4000`, `KEYCLOAK_HOST=http://localhost:18081` and `S3_PUBLIC_ENDPOINT=http://localhost:18333` with `S3_ENDPOINT` untouched, `client-core` getting `CORE_HOST=http://localhost:18090`, and all seven phase servers getting `extra_hosts: localhost=host-gateway` without duplication. No token-validating server is missed.

A few things I checked that hold up, so nobody has to re-derive them:

- Overriding `CORE_HOST` on the phase servers is safe. It only feeds `CORSMiddleware` there; the SDK round-trip uses `SERVER_CORE_HOST`, which the overlay leaves alone.
- Presigning builds a separate S3 client on `S3_PUBLIC_ENDPOINT`, so the SigV4 host matches the published `localhost:18333`.
- Sharing a compose project with `test-e2e-shard` self-heals rather than cross-contaminating. `docker compose run` recreates a dependency whose config diverged, and `down` only removes services declared in the given files.

One blocking issue plus two smaller ones inline. Requesting changes for the Makefile one: `verify-up` silently boots stale images when the build fails, which defeats the purpose of the target.

Makefile
mathildeshagl: Fixed in 0f1a2976. Confirmed the failure mode: the recipe took the exit status of `up -d`, so a broken build booted the last good image and still printed the banner.

I applied your suggestion literally — `set -e` without the leading `@` — rather than the `@set -e` form from `test-e2e-shard`. The original recipe echoed its commands, and adding `@` would have silently changed that output alongside the fix. Say the word if you would rather have it silent for consistency with the e2e target.

.agents/skills/verify/SKILL.md
mathildeshagl: Fixed in 99eaa2a0 — the snippet now hits `/api/courses/`.

Verified your reading of `servers/core/course/router.go:23-24`: the group is `/courses` with `GET("/")` registered, and nothing in `servers/core` overrides `RedirectTrailingSlash`, so the slashless GET really did return a 301 body where the step promised course JSON.

I went with the trailing slash over `/api/courses/self` to keep the snippet exercising the role middleware, not just `authMiddleware` — the step is meant to show the API accepting a lecturer token, not merely a valid one.

.mcp.json
mathildeshagl: Fixed in 9d62c2fc — pinned to `@playwright/mcp@0.0.79`. Confirmed via `npm view @playwright/mcp version` that 0.0.79 is the current release, so the pin does not move anyone backwards today.

Worth noting for later: the pin now needs a deliberate bump, and the skill names the tools it depends on (`browser_navigate`, `browser_snapshot`, `browser_type`, `browser_take_screenshot`, `browser_console_messages`), so a bump should be checked against those.

rappm (approved):
All three points from my earlier review are addressed:

- `make verify-up` now fails on a broken build instead of booting the last good image (0f1a2976).
- The skill hits `/api/courses/`, so the step no longer prints a 301 body where course JSON was promised (99eaa2a0).
- The Playwright MCP server is pinned to 0.0.79 rather than tracking `@latest` (9d62c2fc).

I merged main into the branch to clear the conflict in `.agents/skills/verify/SKILL.md`. Both sides had rewritten the same two gotchas, and main's wording is the current one: `RequireAuth` replaced `onLoad: 'login-required'` in #2111, and #2115 turned the `Unexpected token '<'` symptom into plain 404s for missing assets. I kept main's text for both bullets and carried over the offline `gravatar.com` note, which is still accurate since the avatar resolves through `getGravatarUrl` in the shared package.

PR #2111, opened 2026-08-26
`Authentication`: Fix the lost session when switching between apply and management

### LINKED ISSUE #1724: Apply ↔ Management: session is lost and re-login is forced

rappm:
## Summary

Moving between the application site and the app forced a fresh authentication. The Keycloak
session is now restored from the stored refresh token, so a student crosses between
`/apply/:phaseId` and `/management/**` without re-authenticating, in the SPA as well as after
a full page load or in a new tab.

## Details

**Session handling (`clients/core/src/keycloak/`)**

- `KeycloakProvider` owns the single Keycloak instance and initializes it exactly once.
  `useKeycloak` is now a plain context accessor, so creating and initializing the instance is
  no longer a hook side effect guarded by a context value that was mutated in place
  (`setKeycloakValue` was dead code and no re-render was ever triggered).
- `init` no longer uses `onLoad: 'login-required'`. When both tokens are in `localStorage`
  they are handed to `init`, which exchanges the refresh token for a fresh access token in a
  single `POST /protocol/openid-connect/token` - no authorization redirect, no lost SPA
  state, no half-filled application form thrown away. With no stored tokens `init` resolves
  unauthenticated and issues no request at all, so the public application form stays
  independent of Keycloak.
- New `RequireAuth` wraps the management console and `/apply/:phaseId/authenticated`. It
  renders a loading page until a session exists and asks Keycloak for a login only then, so
  no query can fire with a missing or expired token. `ApplicationAuthenticated` therefore no
  longer gates its query on reading `jwt_token` out of `localStorage`.
- `checkLoginIframe` is disabled: it depends on third-party cookies, which browsers block, so
  it only cost a probe iframe plus a 5s poll. A `storage` listener replaces the part that
  mattered - when another tab removes the tokens, this tab drops its own in-memory session
  instead of rendering an authenticated UI whose requests 401.
- Failure paths clear the session and let `RequireAuth` ask for a login, replacing the
  `alert('Session expired. Please log in again.')` plus forced logout. `logout` clears the
  stored tokens before redirecting (it previously cleared them after `keycloak.logout()`
  navigated away, so they survived) while keeping the in-memory id token that Keycloak needs
  as `id_token_hint`.
- Restoring a session can fail for reasons a login cannot fix (an unreachable Keycloak, a realm
  that will not issue tokens). `init` keeps the rejection reason - normalized, because
  keycloak-js rejects with an `Error` for transport failures but with the raw OIDC payload for a
  refused grant - and `RequireAuth` counts its login redirects in `sessionStorage`, cleared as
  soon as a session exists. After three attempts it stops redirecting and renders that reason
  with a retry instead of a loading page the user can never leave.

**Application entry point**

- `/apply/:phaseId` asked every visitor to log in, including a student arriving from the app
  with a live session. With a session and a phase that offers the university login it now
  forwards to the authenticated form. Anonymous visitors, and phases without a university
  login, keep the card unchanged.
- The header already switches to the user menu and "Go to App" once a user is known; because
  the session is now restored on the public pages, it does so there too.

**Fixtures**

- The seeded full-course application phase gets `universityLoginAvailable: true`, so the
  fixtures cover both variants: that phase for the university login, and the open application
  phase for the manual and external path.

## Trade-offs on the record

These are intended, and are listed here so they are not rediscovered from a log later:

- **A session ended at the identity provider is no longer noticed by an open tab.** With
  `checkLoginIframe` disabled the tab keeps rendering an authenticated UI until its next token
  refresh fails. The `storage` listener only covers a logout in another tab of the same browser,
  not a logout or a session revocation on the Keycloak side. Accepted: the iframe check depends
  on third-party cookies that browsers block, so it bought very little for a probe iframe and a
  5s poll.
- **An exfiltrated refresh token is worth more than it was.** The tokens were already in
  `localStorage`, but they only signed individual requests while the session itself came from
  the SSO cookie; the refresh token now restores a full session. Accepted for this application.
- **A logged-in visitor can no longer apply as an external student.** On a phase with
  `universityLoginAvailable`, an authenticated visitor is forwarded past the card, so
  "Continue without a TUM-Account" is unreachable for them. Intended: that button exists for
  applicants who have no university account, and an authenticated visitor by definition has one.

## Reason / Link to issue

Closes #1724. #1973 removed the explicit `logout()` calls from this flow and left the
initialization fix as the remaining part of the issue.

## How to Test

1. `make test-e2e-shard SHARD=core` covers it end to end, including the new
   `e2e/tests/auth/session-persistence.spec.ts`.
2. Manually, with the dev stack running and the network tab open, logged in as a student:
   - open `/apply/<application phase id>`: the header shows the user menu and the page goes
     straight to the authenticated form, with no request to `/protocol/openid-connect/auth`;
   - reload the page: exactly one `POST /protocol/openid-connect/token` with
     `grant_type=refresh_token`, form still prefilled;
   - "Go to App", then browser Back: authenticated in both directions, and the same in a new
     tab;
   - clear `localStorage` and reload `/management`: one redirect to Keycloak, silent through
     the SSO cookie;
   - in a second tab, log out: the first tab stops showing an authenticated UI;
   - as an anonymous visitor, the open application phase still shows the login card and the
     external application flow completes without any Keycloak traffic.
3. Point `KEYCLOAK_URL` at an unreachable host and open `/management`: three quick redirects,
   then an error page naming the failure with a retry button, instead of a loading page.

## PR Checklist


clients/core/src/keycloak/RequireAuth.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+  useEffect(() => {
+    if (isAuthenticated) {
+      loginRequested.current = false
+      return
+    }
+    if (isInitialized && !loginRequested.current) {
+      loginRequested.current = true
+      login()
+    }
```
mathildeshagl: This can loop with no way out and nothing on screen.

`loginRequested` only guards within a single page load. If the token exchange keeps failing, the sequence is: `init` rejects, `clearSession()` leaves `isInitialized: true` / `isAuthenticated: false`, this effect calls `login()`, Keycloak redirects back, the ref is fresh again because the document reloaded, and `init` fails again. The user sees `LoadingPage` throughout, and there is no state in which anything else can render.

The code this replaces at least surfaced the failure (`alert('Session expired. Please log in again.')` / `alert('Authentication error: ...')`). I am not asking for the alerts back, but there needs to be some terminal state: an attempt counter persisted across the redirect (`sessionStorage`, cleared on `onAuthSuccess`), or an error surface that offers a retry, so a misconfigured realm or a rejected refresh token does not present as a hung page.

clients/core/src/keycloak/KeycloakProvider.tsx
```diff
@@ -1,31 +1,168 @@
...
+    const storedRefreshToken = localStorage.getItem(REFRESH_TOKEN_KEY)
+
+    instance
+      .init({
+        checkLoginIframe: false,
+        ...(storedToken && storedRefreshToken
+          ? { token: storedToken, refreshToken: storedRefreshToken }
+          : {}),
+      })
+      .catch(() => clearSession())
```
mathildeshagl: The other half of the same problem: every init failure collapses into `clearSession()`, which is indistinguishable from "no session stored" as far as `RequireAuth` can tell, so it responds by asking for a login again.

Worth separating the cases the caller can act on. A rejected or expired refresh token is expected and should lead to a login; a network failure reaching Keycloak, or a realm that will not issue tokens at all, is not, and should surface. Even keeping the caught error in state and rendering it once the retry budget from the `RequireAuth` comment is spent would be enough.

clients/core/src/keycloak/KeycloakProvider.tsx
```diff
@@ -1,31 +1,168 @@
...
+    instance.onTokenExpired = () => {
+      void instance.updateToken(5).catch(() => undefined)
+    }
+
+    const storedToken = localStorage.getItem(TOKEN_KEY)
+    const storedRefreshToken = localStorage.getItem(REFRESH_TOKEN_KEY)
+
+    instance
+      .init({
+        checkLoginIframe: false,
```
mathildeshagl, line 131: Question rather than an objection, but I want it to be a decision on the record.

Disabling the login-iframe check is defensible: it depends on third-party cookies, which browsers block, so it mostly cost a probe iframe and a 5s poll. The consequence is that a session terminated at the identity provider is no longer noticed by an open tab; it keeps rendering an authenticated UI until the next refresh fails. The `storage` listener covers a logout from another tab in the same browser, not a logout or a session revocation on the Keycloak side.

Second, related point: the tokens were already in `localStorage` before this PR, but they only signed individual requests, and the session itself came from the SSO cookie. Now the refresh token restores a full session, so its exfiltration is worth more than it was.

Both read as acceptable to me for this application. Please confirm they are intended and say so in the PR description, so this is not something we rediscover later from a log.

clients/core/src/publicPages/application/ApplicationLoginPage.tsx
```diff
@@ -89,6 +91,11 @@ export const ApplicationLoginPage = () => {
   const externalStudentsAllowed = applicationPhase.externalStudentsAllowed
   const universityLoginAvailable = applicationPhase.universityLoginAvailable
 
+  // An already logged-in student has no reason to see the login card again.
+  if (isAuthenticated && universityLoginAvailable) {
+    return <Navigate to={`/apply/${phaseId}/authenticated`} replace />
```
mathildeshagl, line 96: Product question: on a phase with `universityLoginAvailable`, an authenticated visitor is now redirected past the card unconditionally, so "continue as an external student" and "continue without login" become unreachable for anyone with a live session.

That is almost certainly the intent for a student arriving from the app, and it is what makes the bug fix visible. But it does remove a path that exists today, and the phase is configured with both options. If a logged-in user is never meant to apply externally, that is fine and needs no change; if they might be, the redirect wants a condition, or the authenticated form wants a way back to the card.

mathildeshagl (changes requested):
The mechanism is right, and the refactor is a clear improvement: one Keycloak instance owned by the provider, initialized once, with `useKeycloak` reduced to a context accessor. Handing the stored tokens to `init` instead of `onLoad: 'login-required'` is the correct way to make this a refresh-token exchange rather than an authorization redirect, and the e2e spec asserts exactly the right thing (zero `/openid-connect/auth` requests, at least one `grant_type=refresh_token`).

What I would like addressed before merging is the failure path rather than the happy path. Every error now resolves to the same silent state, and one of them can loop. Details inline. The third comment is a question about the security posture and the fourth is a product question; neither needs a code change if the answer is that both are intended.

clients/core/src/keycloak/RequireAuth.tsx
rappm: Fixed in 5ee4abb5. `RequireAuth` now counts its login redirects in `sessionStorage` under `login_attempts`, cleared as soon as a session exists. After three attempts it stops redirecting and renders `ErrorPage` with the init failure message and a retry button that resets the counter, so a rejected refresh token or a misconfigured realm ends on a screen the user can act on instead of a permanent `LoadingPage`.

clients/core/src/keycloak/KeycloakProvider.tsx
rappm: Fixed in 5ee4abb5. `init` now keeps the rejection in `initError` on the context instead of collapsing it into `clearSession()` alone. The first attempts still lead to a login, which is right for the expected case of an expired refresh token; once the retry budget in `RequireAuth` is spent the stored error is what gets rendered, so an unreachable Keycloak or a realm that will not issue tokens surfaces rather than looping.

clients/core/src/keycloak/KeycloakProvider.tsx
rappm: Both are intended, and both are now written down in the PR description under "Trade-offs on the record": a session ended at the identity provider is not noticed by an open tab until its next refresh fails (the `storage` listener only covers another tab of the same browser), and the refresh token in `localStorage` is worth more now that it restores a full session rather than only signing requests.

clients/core/src/publicPages/application/ApplicationLoginPage.tsx
rappm: A logged-in visitor is not meant to apply externally, so no change here. "Continue without a TUM-Account" exists for applicants who have no university account, and an authenticated visitor by definition has one; on a phase with `universityLoginAvailable` the card offers only that button and the university login, so the forwarded path is the one they would have taken. Recorded in the PR description with the other trade-offs.

mathildeshagl (approved):

PR #2123, opened 2026-08-28
`Testing`: Collect every client test and report failures on the pull request

### LINKED ISSUE #2122: `General`: Client tests outside `src/` are silently skipped, and failures report nothing
### Problem or opportunity

The Vitest runner added in #2110 already runs in CI: `quality-clients.yml` has a `test-client` job
running `yarn test`, and both `dev.yml` and `prod.yml` call that workflow, so a red client test blocks
`build-clients`, `e2e-tests`, and `deploy-dev`. Three things around it are still rough.

**1. A test outside a workspace `src/` directory is silently never collected.**

`clients/vitest.config.mts` collects from two places:

```ts
include: ['*.test.ts', '*/src/**/*.test.{ts,tsx}'],
```

`clients/shared/` holds the Module Federation scaffolding every remote builds on (`runtime/mountRemote.tsx`,
`rspack/federatedDependencies.mjs`). It is not a workspace and has no `src/`, so nothing there can be
covered. The same applies to any test a contributor puts beside a workspace rather than under its `src/`.
Vitest does not warn about a pattern that matches nothing: the file is skipped, the run stays green, and the
author sees a passing pipeline.

**2. A failing client test reports nothing outside the raw job log.**

`test-client` runs the default reporter and uploads no artifact. `test-servers.yml` writes
`TestResults-<service>.json` per service and uploads it on `always()`. On the client side, finding out which
assertion failed means opening the job and scrolling.

**3. The shared setup action installs nothing, and every client job installs twice.**

`.github/actions/setup-node` ends with `yarn install`. A composite action's `run` steps execute at the
repository root, where there is no `package.json`, so corepack falls back to yarn 1. It resolves nothing and
writes a stray root `yarn.lock`, which is what `/yarn.lock` in `.gitignore` is covering for:

```
Run yarn install
yarn install v1.22.22
info No lockfile found.
success Saved lockfile.
Done in 0.03s.
```

Both jobs in `quality-clients.yml` then repeat `yarn install` as an explicit step of their own, under
`./clients`, where it picks up the pinned Yarn 4.18.0 and does the real work in ~24s. The action's install is
dead weight in all ten jobs (the nine-module matrix plus `test-client`).

### Desired outcome

- A `*.test.ts(x)` file anywhere under `clients/` is collected, wherever the code it covers lives.
- A failing client test is visible on the pull request diff, and the run leaves a result artifact like the Go
  jobs do.
- Each client CI job installs once, with the yarn version the workspace pins, and leaves no stray lockfile.

### Proposed solution

- Widen the Vitest `include` to `['**/*.test.{ts,tsx}']`. Vitest's default `exclude` already skips
  `node_modules` and `dist`, and the config root is `clients/`, so the Playwright suite in `e2e/` stays out.
- Run `test-client` with the `github-actions` reporter so failures annotate the diff, plus the `junit`
  reporter, and upload the result file with `if: always()`.
- Give the composite's install step `working-directory: ./clients`, and drop the explicit `yarn install` from
  both jobs in `quality-clients.yml`.

### Additional context

No production code changes. Related to #1997.

rappm:
## Summary

The Vitest runner from #2110 collected tests from two directories only, reported failures nowhere but the raw
job log, and ran behind a setup action that installed nothing. This fixes all three, without touching
production code.

## Details

**Collect tests wherever they live.** `include` was `['*.test.ts', '*/src/**/*.test.{ts,tsx}']`, so a test
beside the Module Federation scaffolding in `clients/shared/`, or beside a workspace rather than under its
`src/`, was never picked up. Vitest reports no error for a pattern that matches nothing, so such a file is
skipped and the run stays green. The pattern is now `['**/*.test.{ts,tsx}']`; Vitest's default `exclude`
already skips `node_modules` and `dist`, and the config root is `clients/`, so the Playwright suite stays out.

The first module the old pattern could not reach now has a test: `federatedDependencies` builds the shared
singleton config every remote is compiled against, and reads its versions from `clients/package.json`. It
throws when one of those packages is missing from the root, which nothing observed until now.

**Report failures on the pull request.** `test-client` ran the default reporter and uploaded nothing, while
the Go job has written a result file per service since it was added. It now also runs the `github-actions`
reporter, which annotates the failing assertion on the diff, and the `junit` reporter, whose output is
uploaded as an artifact on `always()`.

**Install once.** `.github/actions/setup-node` ends with `yarn install`, but a composite action's `run` steps
execute at the repository root, where there is no `package.json`. Corepack fell back to yarn 1, which
resolved nothing and wrote a stray root `yarn.lock` (what `/yarn.lock` in `.gitignore` has been covering for):

```
Run yarn install
yarn install v1.22.22
info No lockfile found.
success Saved lockfile.
Done in 0.03s.
```

Every client job then installed again in a step of its own, under `./clients`, where the pinned Yarn 4.18.0
did the real work in ~24s. The action's step now runs in `./clients`, and the duplicate is gone from both
jobs, so all ten jobs install once.

## Reason / Link to issue

Closes #2122

## How to Test

1. `cd clients && yarn test` collects 11 files, including `shared/rspack/federatedDependencies.test.ts`. On
   `main` the same file is skipped and the count is 10.
2. `yarn test --reporter=github-actions` against a deliberately failing assertion prints a
   `::error file=...,line=...` annotation.
3. In this PR's CI, `quality-clients / test-client` runs one `yarn install` on Yarn 4.18.0 and uploads a
   `Vitest-results-clients` artifact.

## PR Checklist


mathildeshagl (approved):
Verified locally: `cd clients && yarn test` collects 11 files and 185 tests, and the file the old pattern could not reach is the new `shared/rspack/federatedDependencies.test.ts`, so the count claim is exact.

The `setup-node` fix is right for the reason you give, and it is safe: that composite action is used in exactly two places, both in `quality-clients.yml`, and both jobs already default to `working-directory: ./clients`, so the job-level default was never reaching the action's own `run` step. Dropping the duplicate `yarn install` from both jobs follows.

`include: ['**/*.test.{ts,tsx}']` with the config root at `clients/` keeps the Playwright suite out on its own, since `e2e/` sits above the root, and Vitest's default `exclude` covers `node_modules` and `dist`.

Four nits, none worth another round:

- The new test derives its expectations from the same `clients/package.json` that `federatedDependencies` reads, so what it actually pins is the contents of `SINGLETON_PACKAGES` rather than the versions. That is the useful half, but the `throw` branch — the one behavior the module has that nothing observed before — stays uncovered. One `expect(...).toThrow()` against a missing package would close it.
- Neither `clients/shared/**` nor the root `sharedPackageImport.test.ts` is included by any `tsconfig.json` in the `quality-client` matrix, since each module's `include` is relative to its own directory. So a type error in either test file passes CI. Pre-existing for the root file, but this PR adds the second one, and `shared` is already linted by Biome in every matrix leg, so a `tsconfig` for it would not be out of place.
- `/yarn.lock` in `.gitignore` was covering the stray root lockfile you have now stopped producing. It can go.
- `TestResults-*.xml` covers the Vitest artifact but the Go jobs write `TestResults-<service>.json`, which is still untracked-but-unignored. `TestResults-*` would cover both.

Issue #2092, opened 2026-08-23
`General`: Service construction is welded to route registration behind package-level singletons


rappm:
### Problem or opportunity

Every Go service except `presentation` reaches its own logic through a mutable package-level singleton. `InitXModule(routerGroup, queries, conn)` constructs the service, assigns it to a package variable, and registers routes, all in one call that returns nothing.

| Server | Singleton variables | References |
| --- | --- | --- |
| `core` | 19 | 360 |
| `assessment` | 14 | 176 |
| `team_allocation` | 8 | 71 |
| `self_team_allocation` | 7 | 36 |
| `interview` | 5 | 33 |
| `certificate` | 4 | 29 |
| `example_server` | 3 | 6 |
| `presentation` | **0** | 0 |

Two consequences:

**There is no way to obtain a service without also registering routes.** Tests that want a service build a throwaway router to get one. `servers/assessment` does this nine times, seven of them as the literal `gin.New().Group("/dummy")`.

**Setup order is part of the contract but is written nowhere.** `AssessmentServiceTestSuite.SetupSuite` has to know to assign four globals in the right sequence before a single assertion can run:

```go
suite.service = AssessmentService{queries: *testDB.Queries, conn: testDB.Conn}
AssessmentServiceSingleton = &suite.service
assessmentCompletion.InitAssessmentCompletionModule(gin.New().Group("/dummy"), *testDB.Queries, testDB.Conn)
scoreLevel.InitScoreLevelModule(gin.New().Group("/dummy"), *testDB.Queries, testDB.Conn)
categoryAssessment.InitCategoryAssessmentModule(gin.New().Group("/dummy"), *testDB.Queries, testDB.Conn)
coursePhaseConfig.CoursePhaseConfigSingleton = coursePhaseConfig.NewCoursePhaseConfigService(*testDB.Queries, testDB.Conn)
```

Miss one and the failure is a nil pointer dereference somewhere unrelated. Package-level mutable state also rules out running tests in the same package in parallel.

This is not a coverage problem. `assessment` has the highest test-to-source ratio of any large service at 0.82, against 0.39 for `core`. The tests are thorough; they just cannot reach the code without reaching around it.

### Desired outcome

Construction and route registration are separate calls. A test obtains a service with a constructor and nothing else. Cross-package calls take a dependency rather than reading a global. Wiring lives in `main.go`.

### Proposed solution

`servers/presentation` already does this and is in production, so this is an adoption task rather than a design one. Its shape:

[...]

PR #2128, opened 2026-08-29
`Team Allocation`: Replace service singletons with constructors

### LINKED ISSUE #2095: `Team Allocation`: Replace service singletons with constructors

### Scope

- **8** singleton variables, **71** references, **8** `InitXModule` functions.
- **8** singleton assignments across `_test.go` files.
- **4** copies of `handleError`, tracked separately.

### What to do

### Done when

- No `var XServiceSingleton` remains in `servers/team_allocation`.
- `make test-team-allocation` passes.

rappm:
## Summary

Removes every package-level service singleton from `servers/team_allocation`. Each module now
exposes a `NewXService(...)` constructor and a separate `RegisterRoutes(...)`, so a service can be
obtained without also registering routes, and `main.go` is the only place that knows the wiring
order.

## Details

`main.go` was deleted:

- `NewXService(queries, conn, deps...)` builds the service.

The package-level functions and gin handlers that read `XServiceSingleton.queries` / `.conn` became
methods on the service, using the receiver's fields. Every route path, HTTP method, middleware,
middleware order and role list is unchanged, including `tease`'s `GET /tease/course-phases`, which
deliberately uses `keycloakTokenVerifier.KeycloakMiddleware()` while its sibling routes use the
role-based middleware.

Cross-package and adapter details:

- `config` used to call `survey.GetSurveyTimeframe` through the survey singleton. `ConfigService`
  now receives a `surveyTimeframeProvider` interface declared on the consumer side, and `main.go`
  constructs the survey service first and passes it in.
- The SDK-facing adapters for the config and copy endpoints keep their role as thin adapters but
  are now package-private and hold their service, so there is no zero-value handler that compiles
  and then nil-derefs at request time.
- `privacy` gained a `router.go` with `RegisterRoutes`, and its two package-level handlers became
  `DataExportHandler` / `DataDeletionHandler` methods.
- `allocation` and `team` build the tutor-scoping middleware from `service.queries` instead of
  taking a second `db.Queries` argument, so there is one source of truth.
- Dropped `ConfigService.conn` and `AllocationService.conn`, both assigned and never read.

The eight test suites that used to assign a singleton now construct the service they need. Two
latent test bugs disappear along the way: the two `allocation` suites used to overwrite each
other's `AllocationServiceSingleton`, and `config` could nil-deref the survey singleton if a
request arrived between two module inits.

for `servers/certificate`.

## Reason / Link to issue

Closes #2095. Part of #2092.

## How to Test

1. `make test-team-allocation`
2. `cd servers/team_allocation && go build ./... && go vet ./...`
3. Start the service and confirm the team allocation phase still loads: skills, teams, the survey,
   allocations, the TEASE endpoints, phase config, phase copy, and the privacy export/deletion
   endpoints all behave as before.

## PR Checklist

## Summary by CodeRabbit

* **Refactor**
  * Improved the reliability and maintainability of team allocation features by streamlining service and route handling.
  * Team, survey, skills, allocation, configuration, copying, TEASE, and privacy capabilities now use consistent service-based processing.
  * Existing routes, permissions, error handling, and user-facing behavior remain unchanged.

* **Tests**
  * Updated automated coverage to reflect the streamlined service setup and route registration.


mathildeshagl (approved):
Checked the claim that the HTTP surface is unchanged rather than reading for it: I extracted every route registration and every `.Group()` call before and after and diffed them. They are identical modulo the `service.` receiver, including the middleware order, the role lists, and `tease`'s `GET /tease/course-phases` keeping `keycloakTokenVerifier.KeycloakMiddleware()` while its siblings use the role-based one. Builds and vets clean.

The `surveyTimeframeProvider` interface declared on the consumer side is the right way to break the `config` to `survey` dependency, and it removes a real failure mode rather than only a global: the old `ConfigHandler{}` was a zero-value struct that compiled and then read `ConfigServiceSingleton` at request time. Dropping the `conn` fields that were assigned and never read is a genuine simplification, not churn.

One nit: `ConfigService.GetPhaseConfig` is exported but its only caller is `HandlePhaseConfig`, three lines above it in the same file. Unexport it unless something outside the package is meant to reach it.

One merge-order note: this and #2119 both touch `survey/service.go` and `tease/service.go`. Git merges them without conflict, but the merged test package does not compile — `survey/service_test.go` still calls the package-level `SetSurveyTimeframe` that this PR turns into a method. Whichever lands second needs a one-line fix. Details on #2119.

Issue #2072, opened 2026-08-20
Self Team Allocation: Timeframe validation is off by the server UTC offset


rappm:
## What happened?

`ValidateTimeframe` in `servers/self_team_allocation` compares timestamps that were written with the
server's local time zone against `time.Now()`, so timeframe enforcement is wrong by the server's UTC
offset. On a host running `TZ=Europe/Berlin` the self-team-allocation window is shifted by 1-2 hours:
students are rejected with `request is outside the allowed timeframe` (HTTP 500) inside the
configured window, and accepted outside it.

The visible symptom today is three failing tests, which pass in CI only because CI runs in UTC:

```
--- FAIL: TestTeamRouterTestSuite/TestAssignTeamRoute          (expected 200, got 500)
--- FAIL: TestTeamRouterTestSuite/TestAssignTeamRouteTeamFull  (expected 400, got 500)
--- FAIL: TestTeamRouterTestSuite/TestLeaveTeamRoute           (expected 200, got 500)
```

### Root cause

1. `timeframe.starttime` / `endtime` are `TIMESTAMP` (without time zone) in
   `servers/self_team_allocation/db/migration/0001_schema.up.sql:30-31`.
2. `servers/self_team_allocation/timeframe/service.go:47-49` writes them as
   `pgtype.Timestamp{Time: startTime}` where `startTime` carries the local zone. pgx's timestamp
   codec calls `discardTimeZone` (`pgx/v5@v5.10.0/pgtype/timestamp.go:234-240`), which rebuilds the
   value from the local **wall clock** and labels it UTC. The stored instant therefore shifts by the
   UTC offset.
3. On read, `GetTimeframeDTOFromDBModel` returns that UTC-located value unchanged, and
   `ValidateTimeframe` (`team/service.go:166`) compares it against `time.Now()`, which is
   local-zoned. The two sides are offset by the host's UTC offset.

With `TZ=Europe/Berlin` (UTC+2), a window of `now-1h .. now+1h` is stored as `now+1h .. now+3h`, so
`time.Now()` lands before the start and the request is rejected.

## Steps to reproduce

```bash
cd servers/self_team_allocation
TZ=UTC              go test ./team/... -run TestTeamRouterTestSuite -count=1   # ok
TZ=Europe/Berlin    go test ./team/... -run TestTeamRouterTestSuite -count=1   # FAIL (3 tests)
[...]

PR #2129, opened 2026-08-29
`Self Team Allocation`: Replace service singletons with constructors

### LINKED ISSUE #2096: `Self Team Allocation`: Replace service singletons with constructors

### Scope

- **7** singleton variables, **36** references, **6** `InitXModule` functions.
- **13** singleton assignments across `_test.go` files.
- **3** copies of `handleError`, tracked separately.

### What to do

This service has the highest test-to-source ratio in the repository at 0.89, so the safety net for the change is already there.

### Done when

- No `var XServiceSingleton` remains in `servers/self_team_allocation`.
- `make test-self-team-allocation` passes.

rappm:
## Summary

Removes every package-level service singleton from `servers/self_team_allocation`. Each module now
exposes a `NewXService(...)` constructor and a separate `RegisterRoutes(...)`, so a service can be
obtained without also registering routes, and `main.go` is the only place that knows the wiring
order.

## Details

`RegisterRoutes(...)`, and the per-module `main.go` was deleted. The package-level functions and
gin handlers that read `XServiceSingleton.queries` / `.conn` became methods on the service. Every
route path, HTTP method, middleware, middleware order and role list is unchanged.

Module-specific notes:

- `team` holds two services behind a single `InitTeamModule`. Both `TeamsService` and
  `AssignmentService` are now constructed explicitly and bound through one package-private handler
  struct, because only `assignTeam` and `leaveTeam` reach the assignment service and both need the
  team service first for the timeframe and capacity checks.
- `team` used to call `timeframe.GetTimeframe` through that package's singleton. `TeamsService`
  now receives a `timeframeProvider` interface declared on the consumer side, so `main.go`
  constructs the timeframe service first. Construction order and registration order therefore
  differ: routes are still registered team-first, exactly as before.
- `copy` lost its `CopyService` entirely. Both of its fields were assigned at startup and never
  read, and `HandlePhaseCopy` returns `nil` unconditionally, so there was nothing for a
  constructor to build. `SelfTeamCopyHandler` stays as the SDK adapter.
- `privacy` gained a `router.go` with `RegisterRoutes`, its two package-level handlers became
  `DataExportHandler` / `DataDeletionHandler` methods, and its fields are now unexported like
  every sibling service.
- The config endpoint's SDK adapter is package-private and holds its service, so a zero-value
  handler can no longer compile and then nil-deref on the first request.
- Dropped four connection pool fields that were assigned and never read, on the allocation,
  assignment, timeframe and config services.

All nine test suites that used to assign a singleton now construct the service they need. No
assertion was changed, weakened or removed.

for `servers/certificate`.

## Reason / Link to issue

Closes #2096. Part of #2092.

## How to Test

1. `make test-self-team-allocation`
2. `cd servers/self_team_allocation && go build ./... && go vet ./...`
3. Start the service and confirm the self team allocation phase still loads: teams, the timeframe,
   self assignment and leaving, tutors, allocations, phase config, phase copy, and the privacy
   export/deletion endpoints all behave as before.

Note on local test runs: `team`'s `TestAssignTeamRoute`, `TestAssignTeamRouteTeamFull` and
`TestLeaveTeamRoute` fail on any host whose UTC offset exceeds an hour, and pass under `TZ=UTC`.
That is #2072, reproducible unchanged on `main`, and needs a `timestamptz` migration to fix. It is
not affected by this PR either way.

## PR Checklist

## Summary by CodeRabbit

- **Refactor**
  - Updated service and route management across team allocation features for more consistent operation.
  - Existing team, assignment, allocation, configuration, timeframe, copy, and privacy endpoints continue using their established routes, permissions, and response behavior.
  - Privacy export and deletion functionality remains available through the same registered endpoints.

- **Tests**
  - Updated automated coverage to validate the revised service and route setup.


mathildeshagl (approved):
Same verification as #2128: I diffed every route registration and `.Group()` call before and after, and they match exactly once the `handler.` receiver is accounted for. No `Singleton` identifier survives anywhere in the service. Builds and vets clean.

The module-specific decisions all read as the right ones:

- Binding `TeamsService` and `AssignmentService` through one package-private handler struct is better than the alternatives, given only `assignTeam` and `leaveTeam` need both and both need the team service first.
- `timeframeProvider` on the consumer side matches the pattern in #2128, and construction order differing from registration order is fine because the routes still register team-first.
- Deleting `CopyService` outright rather than giving a constructor to two never-read fields and an unconditional `return nil` is the correct reading of what was actually there.
- Making `privacy`'s fields unexported brings it in line with its siblings.

One merge-order note: this and #2119 both touch `timeframe/service.go`. Git merges them cleanly, but the merged test package does not compile — `timeframe/service_test.go` still calls the package-level `SetTimeframe` that this PR turns into a method. Whichever lands second needs a one-line fix. Details on #2119.

Also: your note that the three `team` timeframe tests fail off UTC is #2072, which #2119 now fixes, so that caveat can come out of the description once that lands.

Issue #2091, opened 2026-08-23
`Assessment`: Core communication seam is nine mutable package-level function variables


rappm:
### Problem or opportunity

Nothing in `servers/assessment` declares what the phase needs from core. The dependency is spread across raw HTTP call sites, defensive map parsing, and **nine mutable package-level function variables** that exist only so tests can overwrite them:

```
coursePhaseConfig/evaluationReminderSend.go:21  getEvaluationReminderRecipientsForSendFn
coursePhaseConfig/evaluationReminderSend.go:22  getCoreCoursePhaseFn
coursePhaseConfig/evaluationReminderSend.go:23  sendManualReminderMailFn
coursePhaseConfig/evaluationReminderSend.go:24  updateCoreCoursePhaseFn
coursePhaseConfig/router.go:17                  getEvaluationReminderRecipientsFn
coursePhaseConfig/router.go:18                  sendEvaluationReminderManualTriggerFn
coursePhaseConfig/reminderRecipients.go:17      getParticipationsForCoursePhaseFn
coursePhaseConfig/reminderRecipients.go:18      getTeamsForCoursePhaseFn
evaluations/evaluationCompletion/service.go:35  getTeamsForCoursePhaseFn
```

Every test that touches the evaluation reminder path saves these, overwrites them, and restores them on cleanup. Test isolation depends on the deferred restore actually running, and two tests in the same package can never run in parallel.

Assessment is the outlier here. `core` has five such variables, and no other server has any.

Separately, `coursePhaseConfig/interPhaseCommunication.go` hand-walks core's JSON as `map[string]interface{}`: `parsePersons`, `parseTeam`, and `parseTeams` are 130 lines of type assertions that log a warning and skip on anything unexpected. Core's response shape has leaked inward, and a field rename in core surfaces as a silently empty team list rather than an error.

### Desired outcome

One place that states what assessment needs from core, one place that breaks when core changes, and reminder logic testable without a live HTTP endpoint or a global to overwrite.

### Proposed solution

Introduce a single interface for core access, roughly:

```go
type Directory interface {
    CoursePhase(ctx, authHeader, coursePhaseID) (CoursePhase, error)
    Teams(ctx, authHeader, coursePhaseID) ([]promptTypes.Team, error)
    Participations(ctx, authHeader, coursePhaseID) ([]Participation, error)
    UpdateCoursePhase(ctx, authHeader, req) error
[...]

PR #2130, opened 2026-08-29
`Assessment`: Replace service singletons with constructors

### LINKED ISSUE #2094: `Assessment`: Replace service singletons with constructors

### Scope

- **14** singleton variables, **176** references, **14** `InitXModule` functions.
- **35** singleton assignments across `_test.go` files.
- **9** throwaway routers built purely to reach a constructor: seven `gin.New().Group("/dummy")` in `assessments/router_test.go` and `assessments/service_test.go`, plus `gin.New().Group("")` in `categories/router_test.go` and `competencies/router_test.go`.

Assessment has the worst test-setup cost of any service here. `AssessmentServiceTestSuite.SetupSuite` assigns four globals in a required order before any assertion can run, and a missing one surfaces as an unrelated nil pointer dereference.

### What to do

Convert the package-level functions that read `XServiceSingleton` into methods. Pass `coursePhaseConfig`, `assessmentCompletion`, `scoreLevel`, and `categoryAssessment` into `NewAssessmentService` as dependencies rather than reading their globals.

`servers/presentation` is the reference implementation.

### Done when

- No `var XServiceSingleton` remains in `servers/assessment`.
- No test builds a router to obtain a service.
- `make test-assessment` passes.

### Note

Land this before #2091. That work wants the core client to be a field on the service, which is only possible once the service is constructed rather than assigned.

rappm:
## Summary

Removes all 14 package-level service singletons from `servers/assessment`. Each module now exposes
a `NewXService(...)` constructor and a separate `RegisterRoutes(...)`, so a service can be obtained
without also registering routes, and `main.go` is the only place that knows the wiring order.

## Details

`RegisterRoutes(...)`, and all 14 per-module `main.go` files were deleted. The package-level
functions and gin handlers that read `XServiceSingleton.queries` / `.conn` became methods on the
service. Cross-package collaborators arrive as constructor arguments typed as consumer-side
interfaces, so no package reads another package's global.

Every route path, HTTP method, middleware, middleware order and role list is unchanged. Nothing
about the HTTP surface moves.

The six commits are ordered consumer-first, which is what makes each one compile on its own without
leaving a compatibility shim behind. Converting a package's functions into methods breaks every
caller, and fixing a caller means giving it a constructor too, so a package and all its callers are
one atomic unit. Going bottom-up from `assessmentSchemas` would have collapsed the whole service
into a single commit. Going consumer-first means each group keeps calling its dependencies'
still-package-level functions until its own turn comes:

1. `privacy` + `copy`
2. `assessments` + `assessmentCompletion` + `categoryAssessment` + `actionItem` + `scoreLevel`
3. `evaluations` + `evaluationCompletion` + `feedbackItem`
4. `categories` + `competencies`
5. `assessmentSchemas` + `schemaModification`
6. `coursePhaseConfig`, plus the integration wiring

Beyond the mechanical split:

- `assessments/main.go` and `evaluations/main.go` used to register their sub-modules from inside
  their own `InitXModule`. That nesting is gone; `main.go` constructs and registers all fourteen
  itself, in the same flattened order as before.
- `coursePhaseConfig.RequireAssessmentEnabled()` was a package-level gin middleware factory reading
  the singleton, used across four other packages' routers at 11 registrations. It is now a method,
  and those four routers take an `assessmentGuard` dependency. The middleware is still built at
  registration time, in the same argument position.
- `schemaModification` had no singleton, no router and no `InitXModule`, but called into both hub
  packages and would have stopped compiling once they converted. It is now a service. Its
  `GetOrCopySchemaForWrite` drops its per-call `queries` argument, since every call site runs before
  any `Begin` and none passed a transaction-scoped handle.
- `NewCoursePhaseConfigService` takes the schema service, so the `schemas` dependency can no longer
  be omitted.
- Dropped `conn` fields on the score level and action item services, both assigned and never read.

All 35 singleton assignments across 25 test files are gone, and so are the throwaway routers built
only to reach a constructor, including the seven `gin.New().Group("/dummy")` the issue calls out. No
test builds a router to obtain a service any more, and no assertion was changed, weakened or
removed.

for `servers/certificate`.

## Reason / Link to issue

Closes #2094. Part of #2092. Unblocks #2091, which needs the core client to be a field on the
service rather than a global.

## How to Test

1. `make test-assessment`
2. `cd servers/assessment && go build ./... && go vet ./...`
3. `grep -rn "Singleton\|Init[A-Za-z]*Module" --include='*.go' servers/assessment/ | grep -v docs/`
   returns nothing.
4. Exercise an assessment phase end to end: assessments, categories, competencies, schemas, action
   items, score levels, self and peer evaluations, feedback items, reminders, phase config, phase
   copy, and the privacy export/deletion endpoints.

## Notes for the reviewer

Three things a review flagged that this PR deliberately does not change:

- `coursePhaseConfig` now has two seam conventions side by side. Three of the eight monkey-patched
  function variables pointed at functions that became methods and had to become fields; the other
  five point at stateless package functions and were left alone rather than churned. Finishing or
  reverting the remaining five would be worth a follow-up.
- The two reminder seams are constructor-initialized fields on the service rather than constructor
  parameters, because they point at the service's own methods and a parameter form would force
  `main.go` into nil-means-default. They replace package-level mutable state with per-instance
  state, which is an improvement, but they are still fields a test can reassign.
- `actionItem`'s create, update and delete run `CheckAssessmentIsEditable` and then write in two
  separate non-transactional statements, so a concurrent mark-complete can slip between them. This
  race predates this PR and is unchanged by it. Its sibling writers in `assessments` and
  `categoryAssessment` do wrap both in a transaction, so this is worth a follow-up issue.

## PR Checklist

## Summary by CodeRabbit

* **Refactor**
  * Assessment, evaluation, category, competency, configuration, privacy, and copy services now use instance-based service management.
  * Route registration has been standardized across assessment features.
  * Existing assessment, evaluation, feedback, category, competency, score, reminder, copy, and privacy workflows retain their current behavior.
* **Tests**
  * Updated test coverage to use the new service setup and route registration patterns.


mathildeshagl (approved):
Fourteen modules is a lot of surface to take on faith, so I checked the invariant mechanically: every route registration and every `.Group()` call, before and after, extracted and diffed. Identical, including middleware order and role lists. The only apparent difference is `coursePhaseConfig.RequireAssessmentEnabled()` becoming `guard.RequireAssessmentEnabled()`, which is the same middleware built at the same point in the same argument position. Builds and vets clean.

The consumer-first commit ordering is the right justification for why this could not be smaller, and the four separate `assessmentGuard` declarations are correct Go rather than duplication worth removing — each consumer declaring the one method it needs is the point.

I appreciate the reviewer notes pre-declaring the two seam conventions and the `actionItem` non-transactional write. Both read as accurate scope calls. The `actionItem` race in particular is worth the follow-up issue you mention, given its siblings in `assessments` and `categoryAssessment` do wrap the check and the write together.

One nit, which is CodeRabbit's still-open comment and I think it is right: `categories/service_test.go:195` passes a course phase ID to `GetCategoriesWithCompetencies`, whose parameter is `assessmentSchemaID`, and then asserts `GreaterOrEqual(len(catsWithComp), 0)`. That assertion cannot fail, and the loop below it does not execute on an empty slice, so the test passes on a query that returns nothing. The wrong argument predates this PR, but this PR rewrites that line anyway, so it is a cheap moment to pass the schema ID from the fixture and assert a concrete count.

PR #2131, opened 2026-08-30
`Core`: Replace service singletons with constructors

### LINKED ISSUE #2093: `Core`: Replace service singletons with constructors
Part of the singleton removal tracked in the parent issue. `core` is the largest instance in the repository.

### Scope

- **19** singleton variables, **360** references, **14** `InitXModule` functions.
- **28** singleton assignments across `_test.go` files.

### What to do

Convert the package-level functions that read `XServiceSingleton` into methods on the service. Pass cross-package collaborators as constructor arguments instead of reading another package's global. `main.go` becomes the single place that knows construction order.

`servers/presentation` is the reference implementation: `NewService(...)` plus `RegisterRoutes(group, service)`, with the clock injected as a `now func() time.Time` field rather than patched as a package variable.

`core` also has five monkey-patched function variables that should become constructor arguments in the same pass:

```
courseMailing/send.go:29   sendMailFn
courseMailing/send.go:31   nowFn
mailing/manualMail.go:15   sendMailFn
mailing/manualMail.go:16   nowFn
mailing/router.go:14       sendManualMailFn
```

### Done when

- No `var XServiceSingleton` remains in `servers/core`.
- No test assigns a singleton or builds a router to obtain a service.
- `make test-core` passes.

Given the size, this is worth splitting across several PRs by package rather than landing as one.

rappm:
## Summary

Removes all 19 package-level service singletons from `servers/core`, the largest instance in the
repository. Each module now exposes a `NewXService(...)` constructor and a separate
`RegisterRoutes(...)`, so a service can be obtained without also registering routes, and `main.go`
is the only place that knows the wiring order.

## Details

`RegisterRoutes(...)`, and the per-module `main.go` was deleted. The package-level functions and
gin handlers that read `XServiceSingleton.queries` / `.conn` became methods. Cross-package
collaborators arrive as constructor arguments, so no package reads another package's global.

Every route path, HTTP method, middleware, middleware order and role list is unchanged.

### Why it is eight conversion commits, in this order

Converting a package's functions into methods breaks every caller, and fixing a caller means giving
it a constructor too, so a package and all its callers are one atomic unit. Converting bottom-up
from the hub packages would have collapsed the whole service into a single commit. The commits
therefore go consumer-first, so each one compiles on its own with no compatibility shims: a package
nothing calls converts alone, because it keeps calling its dependencies' still-package-level
functions until their own turn comes.

The corollary is that a commit converting a provider also rewires every already-converted consumer
of that provider, in the same commit. That is why, for example, the auth commit also touches
`privacy/service`.

1. `privacy` + `privacy/service`
2. `auditLog`
3. `applicationAdministration`
4. `auth` + `auth/service` + `coursePhaseType` + `courseMailing`
5. `course` + `course/copy`
6. `coursePhase` + `coursePhaseParticipation` + `coursePhase/resolution`
7. `student`, `instructorNote`, `mailing`, `keycloakRealmManager`, `course/courseParticipation`,
   `storage/files`, `storage/privacyexport`
8. `keycloakTokenVerifier` + `permissionValidation`

A ninth commit follows the conversion and only removes the nil guard the interface conversion made
dead (see the reviewer note below), so it is a cleanup rather than part of the ordering above.

### Beyond the mechanical split

- **The issue's reference count is misleading, in a way that shaped the plan.** Most references to
  the two hub packages need no dependency at all: 251 of the `permissionValidation` ones are role
  constants (`PromptAdmin`, `CourseLecturer`, …) and around 21 of the `keycloakTokenVerifier` ones
  are `Ctx*` context keys. The genuine singleton-backed calls number about 50, and every one lives
  in route-registration code as a middleware argument, never in a service body. That is what makes
  the final commit signature threading rather than a rewrite of business logic.
- **`storage/privacyexport` is included even though it has no identifier containing "Singleton".**
  Its lowercase package-level `adapter` plus `Init` / `InitWithAdapter` is the same hidden
  package-state pattern, and `privacy` reads it.
- **Monkey-patched function variables become service fields**, filled by the constructor with the
  previous defaults: `courseMailing`'s `sendMailFn`, `nowFn` and `runSendAsync`, and `mailing`'s
  `sendMailFn`, `nowFn` and `sendManualMailFn`. Package-level mutable state becomes per-instance
  state, so tests no longer patch a global every other test in the package shares.
- **Startup work that was hidden inside `Init*` is now explicit in `main.go`**, with identical
  behavior: `courseMailing`'s `ReconcileStuckCampaigns` background goroutine, `coursePhaseType`'s
  phase-type bootstrap including its `log.Fatal`s, `applicationAdministration`'s phase-type
  initialization, `storage/files`'s fatal-on-error construction, and `keycloakRealmManager`'s login
  probe. The realm manager deliberately does **not** become fail-fast: `main.go` registers its
  routes today even when the probe fails, so a hard failure would be a behavior change.
- **The audit middleware is still installed before Keycloak setup and before every module's route
  registration**, because gin snapshots middleware at registration time.
- Functions that never read a singleton stayed package-level rather than being converted for
  symmetry, which keeps the diff to what actually had to change. `CheckAccessControlByID` and
  `CheckAccessControlByRole` are the notable examples: one takes its check as an argument, the
  other only reads the gin context.
- Dropped a number of `conn` fields that were assigned and never read.

All 28 singleton assignments across the test files are gone. No test assigns a singleton or builds
a router to obtain a service any more, and no assertion was changed, weakened or removed.

This follows `servers/presentation` (the reference implementation) and #2116, #2128, #2129 and
#2130, which did the same for the phase services.

## Reason / Link to issue

Closes #2093. Part of #2092.

## How to Test

1. `make test-core`
2. `cd servers/core && go build ./... && go vet ./...`
3. `grep -rn "Singleton\|Init[A-Za-z]*Module" --include='*.go' servers/core/ | grep -v docs/`
   returns nothing.
4. Exercise core end to end: login, course and course phase CRUD, phase copy, applications and
   their file uploads, participations, students, instructor notes, mailing and course mailing
   campaigns, the audit log, and the privacy export and deletion flows.

## Notes for the reviewer

- The last commit removes a nil guard in `applicationAdministration.cleanupReplacedFiles`. It was
  meaningful against the old package-level singleton, but `s.files` is an interface now, so a
  typed-nil would satisfy the check and then panic. Every call site supplies a real service and
  `main.go` exits if one cannot be built, so the guard was dead as well as misleading.
- Two shape inconsistencies are left as follow-ups rather than churned here: `student` and
  `instructorNote` still reference `permissionValidation.CheckAccessControlByRole` directly instead
  of taking it as a parameter, and most packages hang gin handlers off the service struct while
  `auth`, `privacy` and `coursePhaseType` use a separate handler struct, which they need because
  their service lives in a sibling package.
- `keycloakTokenVerifier.TOP_LEVEL_GROUP_NAME` is dead: `keycloakRealmManager` declares and uses
  its own identical copy. Left alone as out of scope.

## PR Checklist


mathildeshagl (approved):
124 files, and the thing that makes it reviewable is that the invariant is checkable. I extracted every route registration and every `.Group()` call before and after and diffed them; they match once the receiver is accounted for. Two diffs looked like real changes and are not:

- `api.GET("/audit-log/status", keycloakTokenVerifier.KeycloakMiddleware(), ...)` becoming `authMiddleware()` is parameterization — `main.go` passes `tokenVerifier.KeycloakMiddleware`.
- `ingestAuditEvent(sink)` becoming `ingestAuditEvent(NewDBSink(queries))` is the same sink, built from the service's own queries instead of read off the singleton.

Builds and vets clean, and the whole test and e2e matrix is green.

The genuinely valuable part is not the singleton removal, it is that `permissionValidation` is now constructed before `initKeycloak` instead of after. The old ordering was load-bearing and undocumented, and the comment explaining why `InitAuditLogRoutes` had to come later existed precisely because of it. That class of startup-order footgun is gone. Keeping the realm manager non-fail-fast so route registration still happens when the login probe fails is the right restraint — that would have been a behavior change hiding inside a refactor.

I also checked the nil-guard removal in `cleanupReplacedFiles`. Your reasoning holds: against an interface field a typed nil satisfies `!= nil` and then panics on first use, so the guard was worse than absent, and `main.go` exits if the storage service cannot be built.

One thing I would change: `auditLog.RegisterRoutes` still calls `service.StartRetentionPruner(context.Background())` internally. That is the one lifecycle side effect not hoisted into `main.go` — `InitializePhaseTypes`, `ReconcileStuckCampaigns`, `InitializeApplicationCoursePhaseType` and `StartExportDeletionRoutine` all were, and the description lists that hoisting as a deliberate goal. A function named `RegisterRoutes` spawning a daily background goroutine is exactly the hidden startup work the rest of the PR removes. It is inert in the one test that calls it only because that test sets `AUDIT_ENABLED` empty. Moving the call next to the other four in `main.go`, behind the same `audit.Enabled()` check, would finish the job.

The two shape inconsistencies and the dead `TOP_LEVEL_GROUP_NAME` you flag are fine as follow-ups.

PR #2132, opened 2026-08-30
`Team Allocation`: Allow tutors to edit allocations for their assigned team

### LINKED ISSUE #1006: PLs should only see their own teams
I want to be able to filter out any other teams, which I am not interested in

### LINKED ISSUE #1872: Allow tutors to edit allocations for their assigned team
## Background

Part of #1698 (Improve Editor Rights Management). PR #1853 scopes tutor (`CourseEditor`) **read** access in Team Allocation to their assigned team: tutors now only *see* their own team's teams and allocations. The remaining, unaddressed part of #1698 is **write** access:

> Allow Team Allocation Edit (e.g. by giving Team Allocation an input team)

Today all allocation writes in `team_allocation` are `PromptAdmin` / `CourseLecturer` only. A scoped tutor can view their team but cannot edit its allocation.

## Goal

Let a scoped tutor edit the allocation of the team they are assigned to, without granting them access to other teams.

## Scope

- Add editor-writable allocation endpoint(s) (e.g. assign/move a participant within the tutor's team), protected for `CourseEditor` and gated by the tutor scoping middleware.
- Reuse the scoping resolved by `TutorScopingMiddleware` (prompt-sdk #89 / prompt-sdk PR #99) so a tutor can only write to `GetTutorTeamID(c)`; reject writes targeting any other team with `403`.
- Lecturers/admins keep unrestricted write access.
- Consider the "input team" UX referenced in #1698 (a team input on the allocation editor).

## Notes

- Depends on the tutor scoping already landed in #1853 and the SDK extraction in prompt-sdk PR #99.
- Needs tests for: tutor editing own team (allowed), tutor editing another team (403), lecturer editing any team (allowed).

rappm:
> [!IMPORTANT]
> **Blocked by `prompt-edu/prompt-sdk#116`** (issue `prompt-edu/prompt-sdk#115`).
>
> This PR defines the tutor **write** authorization policy locally, in
> `authorizeAllocationWrite` in `servers/team_allocation/allocation/router.go`. That policy is
> cross-phase access control, not a Team Allocation concern: when another module consumes a team
> with tutor input, the rules have to be the same whichever phase produced the team, and a second
> team implementation deriving the policy again is how the two diverge.
>
> The SDK PR moves it into a `tutorscope` package (`AuthorizeWrite`, `Access.Guard`,
> `NewPgxResolver`, `NormalizeLogin`). Once it is released this PR is rebased onto it and drops:
>
> - `authorizeAllocationWrite` -> `tutorscope.AuthorizeWrite`
> - the local `ErrTeamWriteDenied` -> `tutorscope.ErrWriteDenied`
> - the hand-built `pgtype.UUID` guard -> `access.Guard()`
> - the target-team check -> `access.AllowsTeam(request.TeamID)`
> - `servers/team_allocation/tutorscope` and `teamDTO.NormalizeUniversityLogin`, so
>   `team_allocation` keeps no tutor access control of its own
>
> Nothing below changes in behavior. The endpoints, the SQL guard and the test matrix stay as they
> are; only the owner of the policy moves.

## Summary

Lets a tutor edit the allocation of the team they are assigned to, without giving them any access
to the other teams of the phase. Adds the first per-participant allocation write endpoints to
`team_allocation` and gates them with the tutor scoping that PR #1853 already uses for reads.

## Details

Until now `servers/team_allocation/allocation` exposed only two `GET` routes: the sole allocation
write path was the TEASE workspace save, which deletes every allocation of the phase and reinserts
its payload, restricted to `PromptAdmin` / `CourseLecturer`. A scoped tutor could see their team but
could not change it.

**New endpoints**, both `PromptAdmin`, `CourseLecturer`, `CourseEditor` behind `TutorScopingMiddleware`:

- `PUT /api/course_phase/:coursePhaseID/allocation/:courseParticipationID` with `{ "teamID": "…" }`
- `DELETE /api/course_phase/:coursePhaseID/allocation/:courseParticipationID`

**Writes fail closed.** Reads deliberately fail open: the SDK's scoping middleware passes an editor
through untouched when it cannot resolve a tutor record (no `university_login`, or no `tutor` row),
and those editors keep full read access. Reusing that gate for writes would have handed exactly
those editors unrestricted write access to every team. `authorizeAllocationWrite` therefore grants
unrestricted writes only to `PROMPT_Admin`, `PROMPT_Lecturer` and course lecturers, and requires
every other caller to be an editor **with** a resolved tutor team.

**Authorization and mutation are a single statement.** Both new queries take an optional
`expected_team_id`; `COALESCE(expected_team_id, a.team_id)` means a `NULL` guard writes
unconditionally while a tutor's guard only matches a row still in their team. Reading the current
team first and then writing would have left a window in which a concurrent lecturer or TEASE save
moved the participant, so a tutor's write could land on a foreign team.

**A tutor may not poach from another team.** The target team must be their own, and the participant
must be unallocated or already in that team. Pulling a participant out of another team changes that
team's roster too, which is what the issue rules out.

**Participants are verified against core before any write.** `allocations.course_participation_id`
has no foreign key, so an invented UUID would otherwise become a row in the tutor's team. The new
`coreRequests.GetCoursePhaseParticipants` reads `GET /api/course_phases/{id}/participations` (which
core already permits for `CourseEditor`) and deliberately does **not** follow the resolution
endpoints it also returns, so an unrelated phase module being down cannot fail an allocation write.
The same call supplies the student names cached on the allocation row, which the team read endpoints
project; without it a tutor-created allocation would show a blank name. Core's single-participation
route is not usable here: its `GetCoursePhaseParticipation` DTO carries no `student` field, so it
cannot supply the names, and the SDK wrapper around it fans out to every registered resolution
endpoint. The cost is one list fetch per write, on a low-frequency staff action.

Also registers both routes in `api-stress/catalog/partial_team_allocation.json`, the declared source
of truth for the API surface.

### Known limitations, stated deliberately

- **TEASE saves still win.** `POST /tease/save` publishes a whole-phase draft: it clears every
  allocation of the phase and reinserts the payload the TEASE client loaded earlier, so it overwrites
  a tutor's direct edit made since that draft was loaded. Detecting this needs a version token the
  external TEASE client sends back, i.e. a contract change in another codebase; rejecting stale saves
  server-side without it would only surface as a broken save in a UI that cannot handle the error. A
  comment on `SaveTeaseWorkspaceAndAllocations` records the behaviour; optimistic concurrency for the
  TEASE save is worth a follow-up issue.
- **A global `PROMPT_Lecturer` who is also a tutor stays scoped.** That role is not in the routes'
  allowed set, so such a user is admitted as a course editor and their *reads* are tutor-scoped;
  treating the raw role as privileged would have unscoped only their writes.
- **Authorization is decided at the start of the request.** The tutor's team is resolved once by the
  middleware, so a tutor whose assignment is changed mid-request finishes with the old one. This
  matches the read-side semantics already shipped in #1853; making writes stricter than reads would
  be inconsistent without changing both.

### Not included

No editor-facing UI. `clients/team_allocation_component` has no allocation editor at all today (the
`/allocations` page links out to TEASE), and its `/allocations` and `/participants` routes are
lecturer-only and read from the lecturer-only TEASE endpoint. Opening them to editors means building
an allocation editor and migrating those pages onto the scoped endpoint, which is its own change with
its own UX decisions. This PR is the API and authorization foundation; the "input team" UX from #1698
follows separately.

## Reason / Link to issue

Closes #1872. Part of #1698 (Improve Editor Rights Management), continuing the read-side scoping from
#1853. Closes #1006. 

Blocked by `prompt-edu/prompt-sdk#116` (issue `prompt-edu/prompt-sdk#115`): the tutor write policy
belongs in the SDK as a cross-phase feature, not in this service. See the note at the top.

## How to Test

1. `cd servers/team_allocation && go test ./...` - the new cases in `allocation/router_test.go` and
   `allocation/service_test.go` cover tutor-own-team (allowed), tutor-foreign-team as both target and
   source (403), non-tutor editor (403), lecturer and admin (allowed), team from another phase (400),
   non-participant (400), core unavailable (502, no write), oversized body (413) and missing
   allocation (404).
2. Against a running stack, as a course editor recorded as a tutor of team A:
   - `PUT /team-allocation/api/course_phase/<phase>/allocation/<unallocated participant>` with
     `{"teamID":"<team A>"}` -> `200`
   - the same call with `{"teamID":"<team B>"}` -> `403`
   - `PUT` for a participant currently in team B, targeting team A -> `403`, and team B is unchanged
   - `DELETE` for a participant in team A -> `204`; for one in team B -> `403`
3. As a course lecturer, every one of those writes succeeds.

## PR Checklist

## Summary by CodeRabbit

* **New Features**
  * Added the ability to assign or move a participant to a team.
  * Added the ability to remove a participant’s team allocation.
  * Supported permissions for administrators, lecturers, and appropriately scoped editors/tutors.

* **Bug Fixes**
  * Added validation to prevent assignments involving invalid teams, phases, or participants.
  * Improved error handling for unauthorized, missing, oversized, or unavailable requests.

* **Documentation**
  * Updated API documentation with the new allocation operations, request format, permissions, and response codes.


rappm:
Converted to draft: this is blocked by prompt-edu/prompt-sdk#116 (issue prompt-edu/prompt-sdk#115).

The reason is scope ownership, not a defect here. `authorizeAllocationWrite` is a cross-phase
authorization policy that happens to be written in a Team Allocation router. Tutor scoping needs to
behave identically for every team implementation: `self_team_allocation` has a byte-identical `tutor`
table (without `university_login`, so it cannot even implement the resolver interface today), and
once a module like Assessment depends on a team with tutor input, it has to be safe to assume the
access control is the same whichever phase produced that team. A second implementation deriving this
policy again is exactly how the two drift, and a drift in either direction is a real authorization
bug: too strict locks tutors out, too lax gives every unresolvable editor write access to every team
in the phase.

The SDK PR moves the whole feature into a `tutorscope` package:

- `AuthorizeWrite` fails closed, the mirror image of the middleware's deliberate fail-open for reads
- `Access.Guard()` makes the atomic `COALESCE` guard idiom shared instead of folklore, and
  `AllowsTeam` covers the target-team direction the database cannot check
- `NewPgxResolver` over a canonical `tutor` table, so no phase needs an adapter or a generated query
- `NormalizeLogin` as the single definition of the stored login form
- the read middleware now records that it ran, so a route that forgets it gets a 500 wiring error
  instead of silently 403-ing every legitimate tutor

Once that is released this PR gets rebased onto it and deletes `authorizeAllocationWrite`,
`ErrTeamWriteDenied`, the hand-built `pgtype.UUID` guard, the target-team comparison, and
`servers/team_allocation/tutorscope`. The endpoints, the SQL guard and the test matrix are unchanged;
only the owner of the policy moves. Follow-ups tracked on the SDK issue: giving
`self_team_allocation` a `university_login` column and scoping its currently unscoped editor reads.

PR #2133, opened 2026-08-30
`Core`: Add an optional welcome text to the application form

### LINKED ISSUE #360: Provide a welcome/info text for applications
Currently, the application form does not provide instructors with the possibility to provide a welcome or information text.

Such an optional text, that can be freely configured by the instructor, can help students to confirm they are applying for the correct course.

It could be placed between the dates and the form elements:
Image

rappm:
## Summary

Lets instructors write an optional welcome/info text for their application, rendered to applicants
between the course dates and the application form. It is meant to help a student confirm they are
applying for the right course.

## Details

**The text has to be projected explicitly.** `GET /api/apply/:coursePhaseID` is unauthenticated and
does not return `restrictedData` (or `studentReadableData`) as a blob: `GetOpenApplicationPhase`
hand-projects three named keys. A new key is invisible to an applicant until it is added there, so
this is not a frontend-only change.

**The projection is coalesced on purpose.** `(restricted_data->>'key')::text` makes sqlc infer a
non-nullable `string`, which would fail the scan for any phase that never set the key and take the
whole apply endpoint down with it (the same sharp edge the e2e seed already documents for
`externalStudentsAllowed`). `COALESCE(..., '')` keeps the column non-null, and an empty string maps
to a nil pointer so `omitempty` drops it from the response.

**The value is validated on write.** It is served publicly, so `validateUpdateCoursePhase` now
type-checks `welcomeText` and caps it at 20 000 bytes. The rest of the metadata stays schema-less.

**Storage.** `restrictedData` on the course phase, alongside every other application setting. No
migration: the column is schema-less `jsonb` and `UpdateCoursePhase` merges shallowly, so a card that
writes only `welcomeText` cannot clobber the neighbouring keys. `studentReadableData` would not have
helped: it is for enrolled course students and is not projected into the apply DTO either.

**Rendering** reuses `FormDescriptionHTML`, the component that already renders instructor HTML for
question descriptions (DOMPurify, plus the hook forcing `target=_blank rel=noopener` on links). It
gains an optional `className` that defaults to today's styling, so the five existing callers are
untouched while the welcome text is not rendered smaller than the course description next to it. The
new `ApplicationWelcomeText` sits in `ApplicationHeader`, which is the dates block from the issue's
screenshot and the one component both apply pages (anonymous and Keycloak) already share.

**Editing** is a new settings card using the shared `DescriptionMinimalTiptapEditor`, mounted only
once the phase metadata has arrived. Emptying the editor still emits markup like `<p></p>`, so the
card normalizes that to "no text" and sends `null`, which removes the key rather than leaving a blank
string behind.

### Scope note on sanitization

The welcome HTML is sanitized in the browser at the render sink, not on the server. That is the
convention this path already follows: every application question's `description` is instructor HTML
served by the same unauthenticated endpoint into the same `FormDescriptionHTML` sink, with no
server-side sanitization. So this field joins an existing class rather than creating a new one, and
sanitizing only `welcomeText` server-side would leave question descriptions inconsistent. Moving all
instructor HTML behind a server-side allowlist means a new sanitizer dependency in core and a change
to the question path too, and is worth its own issue. The size and type cap above is included here
because it is cheap and specific to the public field.

### Not changed

- `GetAllOpenApplicationPhases` (the public open-course list) does not carry the text: it has no
  place to render it, and the list should not grow an HTML blob. The e2e API spec asserts this.
- The lecturer "Preview Application" dialog renders only `ApplicationFormView`, not
  `ApplicationHeader`, so it does not preview the welcome text. The settings card shows the rendered
  rich text directly, so this is accepted rather than fixed here.
- `PUT /api/course_phases/:uuid` authorizes the path `:uuid` but writes the body's `id`. That is
  pre-existing, affects every metadata write, and needs its own change; the new card sends matching
  IDs, so nothing here depends on it.

## Reason / Link to issue

Closes #360.

## How to Test

1. `cd servers/core && go test ./coursePhase/ ./applicationAdministration/` - covers the welcome text
   present, absent (the coalesce regression), and the write-side type and size validation.
2. `make test-e2e-shard PATHS="/tests/application/welcome-text\.spec\.ts$"`.
3. Manually: open an Application phase's **Settings**, write something in **Welcome Text**, save, then
   open `/apply/<phaseId>` in a private window - the text appears between the dates and the form.
   Clear the editor, save, and it disappears again.

## PR Checklist


mathildeshagl (approved):
The two non-obvious decisions here are both right, and both are the kind that would have caused a production incident if guessed at.

Projecting the key explicitly, because `GetOpenApplicationPhase` hand-projects named keys rather than returning the blob — a frontend-only change would simply have shown nothing. And coalescing the projection, because `(restricted_data->>'welcomeText')::text` makes sqlc infer a non-nullable `string`, which fails the scan for every phase that never set the key and takes the entire unauthenticated apply endpoint down with it. That second one is a genuine trap and the `COALESCE` plus nil-pointer-on-empty handling is the correct way out of it. Capping type and size on a publicly served field is cheap and right. Core builds and vets clean, and the whole matrix is green.

The scope note on sanitization is a fair argument: this field joins the class that every application question `description` already belongs to, and fixing only `welcomeText` server-side would leave that path inconsistent. I agree it is a separate issue.

**But that makes one thing worth fixing here.** This PR edits `FormDescriptionHTML`, and that component calls `DOMPurify.addHook('afterSanitizeAttributes', ...)` inside the render body. DOMPurify hooks are global and cumulative, so every render of every description stacks another duplicate hook onto the module-level instance — an unbounded list of identical hooks, and a global side effect performed during render.

I would not raise it as a drive-by, except that #2134 in this same batch fixes exactly this, deliberately, with a comment explaining why:

```
// Registered once, lazily: DOMPurify hooks are global, so registering per render
// would stack duplicates, but registering at import time crashes where there is
// no DOM
```

So the better pattern already exists in the same set of changes. While you are in this file adding `className`, hoisting the `addHook` out of the component body is two lines.

The same comparison shows the core sink is weaker than the new one in two other ways, both of which now matter more because this PR puts a large, prominent instructor-authored field on an unauthenticated public page:

- `DOMPurify.sanitize(htmlCode)` with no options, against #2134's `USE_PROFILES: { html: true }` and `ALLOW_UNKNOWN_PROTOCOLS: false`.
- The hook fires on `'target' in node`, which is true for many elements, and sets only `rel="noopener"`. #2134 scopes it to `A` with an `href` and sets `noopener noreferrer`.

None of that is a regression you are introducing, and I am approving on that basis. It is that the two PRs establish two different standards for the same job in the same batch, and the weaker one is the one on the public endpoint.

Two housekeeping items:

- The `github-advanced-security` comment on `ApplicationSettingsWelcomeText.tsx:26` is **stale**. It was raised against the regex tag-strip, and `5eb4f44b` replaced that with an inert `DOMParser` emptiness check — which is the right fix, and the code comment anticipates the flag. Worth confirming the alert auto-dismissed rather than leaving the thread looking open.
- This conflicts with #2108, which deletes `e2e/seed/e2e_seed.sql` while this modifies it (`modify/delete`). #2134 has the same conflict and #2135 already merged its additions to that file, so #2108's conflict surface is growing with every e2e fixture PR. Not yours to fix, but relevant to merge order.

PR #2134, opened 2026-08-30
`Certificate`: Add a customizable instructor text to the student download page

### LINKED ISSUE #1228: Customizable instructor text on student download page
## Context

The certificate phase includes a student-facing download page where students can preview and download their certificate once it has been released. Currently, this page shows a generic message.

Instructors often want to include course-specific information on this page, such as:
- Congratulatory messages
- Instructions for how to use the certificate (e.g., adding to LinkedIn)
- Contact information for questions
- Links to further resources or next steps
- Notes about certificate validity

## Current Behavior

The student download page (`StudentOverviewPage`) displays a fixed layout with a download button and certificate preview. There is no way for instructors to customize the text or add additional information shown to students.

## Proposed Solution

Add a configurable rich-text or markdown field in the certificate settings that instructors can use to write custom content. This text is then displayed on the student-facing download page above or below the certificate preview.

### Acceptance Criteria

- [ ] A text field (markdown or rich text) is available in the certificate settings page
- [ ] Instructors can write and save custom content for the student download page
- [ ] The student download page renders the instructor-provided text
- [ ] Empty/unset text gracefully falls back to a sensible default or no extra text
- [ ] Text is stored in the certificate configuration (e.g., as part of `restrictedData` or a dedicated field)

## Related

Part of the Certificate of Completion feature (#690).

rappm:
## Summary

Lets instructors write a custom message for the student-facing certificate download page — a
congratulation, how to add the certificate to a profile, who to contact, or why it is not available
yet — and renders it on that page.

## Details

**Storage: a column on the certificate service's own config, not `restrictedData`.** The issue
suggests `restrictedData`, but that would not work: `servers/core/coursePhase/router.go` blanks
`restrictedData` for anyone who is not an admin, course lecturer or course editor, so a student
reading the course phase gets `{}`. `course_phase_config` has no JSON column to piggyback on either,
so this adds migration `0003` with a `student_page_text` column (plus the matching column in the
hand-written test dump, which declares the table itself).

**A dedicated write endpoint, not a wider `PUT /config`.** `UpdateConfigRequest.TemplateContent` is
`binding:"required"`, so `PUT /config` is a whole-template replacement; relaxing that to allow
partial updates would let a text-only write erase a template. `PUT /config/student-page-text` mirrors
the `/config/release-date` endpoint that already exists for exactly this reason. Its query is an
**upsert**, so it also creates the config row — `UpdateReleaseDate` is a plain `UPDATE` and silently
does nothing when no row exists yet, which is a bug worth not copying.

**Omitted is not the same as null.** The request binds `studentPageText` as `json.RawMessage`, so an
absent key is a `400` rather than a silent clear, an explicit `null` clears the text, and a non-string
or oversized value is rejected. The value is capped at 20 000 bytes.

**Students read the text from the status endpoint they already call.** `StudentOverviewPage` makes
exactly one request, `GET /certificate/status`, so the text is attached there rather than to
`GET /config`. It is attached to *every* successful status — including "template not configured" and
"not yet released" — because that is precisely where an instructor wants to explain the wait. The
handler has five 200 exits, so one small helper decorates them and every existing field stays
byte-identical.

### Security fix included

`GET /config` admitted `CourseStudent` while returning the **full Typst template**, the updater's
name and the release state — even though `GET /config/template` is deliberately staff-only. Nothing
student-facing ever called it, so this was latent, but adding a student-relevant field to that DTO
would have made it live. `CourseStudent` is removed from the route; the only caller in this repo is
the lecturer-only settings page.

`CourseEditor` **is** left on `GET /config`, so editors can still read the template there even though
`/config/template` excludes them. That incoherence predates this issue and cuts both ways — editors
are already authorized to download any student's certificate — so it is a product call for a
maintainer rather than something to decide in this PR.

### Rendering

Instructor HTML is sanitized with DOMPurify at the render sink, the convention this repo already uses
for instructor-authored HTML. A Module Federation remote cannot import core's `FormDescriptionHTML`,
so the certificate remote gets a local equivalent. It is not a blind copy: the DOMPurify hook is
registered once at module scope rather than on every render, it is scoped to anchors, it sets
`rel="noopener noreferrer"`, and sanitizing is pinned to the HTML profile with unknown protocols
disallowed. The proper long-term home is `@tumaet/prompt-ui-components` in the `prompt-lib`
repository, which is outside this repo.

The editor is the shared `DescriptionMinimalTiptapEditor`. It reads its content only once when it is
created, so it is keyed on the stored value to re-hydrate after load and after save. An emptied
editor still emits markup like `<p></p>`; that is normalized to "no text" and sent as `null`, while
content that renders without any text of its own (a rule, an image) still counts as text.

## Reason / Link to issue

Closes #1228. Part of the Certificate of Completion feature (#690).

### Acceptance criteria

- [x] A text field (markdown or rich text) is available in the certificate settings page — rich text;
      the repo has no markdown renderer anywhere and the convention is tiptap → HTML → DOMPurify.
- [x] Instructors can write and save custom content for the student download page
- [x] The student download page renders the instructor-provided text
- [x] Empty/unset text gracefully falls back to no extra text (the existing generic copy stays)
- [x] Text is stored in the certificate configuration — a dedicated column, since `restrictedData`
      is blanked for students

## How to Test

1. `cd servers/certificate && go test ./...` — covers the round trip, the explicit-null clear, the
   omitted/non-string/oversize rejections, that a text-only write preserves the template and release
   date, and that the text appears on the status payload in both the configured and unconfigured
   states and disappears when cleared.
2. `make test-e2e-shard SHARD=certificate` — a lecturer writes the text through the UI and a student
   in a second browser context sees it; the sanitizer is exercised with a script tag, an inline
   `onerror` and a `javascript:` URL, none of which may survive or run.
3. Manually: open a certificate phase's **Settings**, write in **Student Download Page Text**, save,
   then open the phase as a student — the message appears above the certificate status.

## PR Checklist

## Summary by CodeRabbit

* **New Features**
  * Instructors can add optional rich-text messages to students’ certificate pages.
  * Messages appear across certificate states, including before release, and persist after saving.
  * Empty messages can be cleared, while unsafe HTML is removed and links are securely formatted.
  * Certificate settings now provide an editor, save controls, validation, and success or error feedback.

* **Documentation**
  * Added guidance for configuring and viewing student download-page messages.


mathildeshagl (approved):
The storage decision is correct and worth having written down: `servers/core/coursePhase/router.go` blanks `restrictedData` for anyone who is not admin, lecturer or editor, so the approach the issue suggests would have returned `{}` to exactly the audience the feature is for. Migration `0003` is numbered correctly, has a real down, and matches the repo's `BEGIN`/`COMMIT` convention. Certificate builds and vets clean, and the matrix is green.

Three things I think are better than they had to be:

- Mirroring `/config/release-date` with a dedicated `PUT /config/student-page-text` rather than relaxing `UpdateConfigRequest.TemplateContent`'s `binding:"required"`. Making `PUT /config` accept partial updates so a text write could go through it would have let a text-only save erase a Typst template, which is a nasty way to lose work.
- Making the new query an upsert instead of copying `UpdateReleaseDate`'s plain `UPDATE`, which silently does nothing when no config row exists. Noticing and not propagating an adjacent bug is the right instinct.
- `json.RawMessage` so that omitted, `null`, and a non-string value are three distinct outcomes rather than one silent clear. That distinction is usually skipped and then regretted.

Attaching the text to `GET /certificate/status` rather than `GET /config` is also right, and the reasoning for putting it on *every* successful status — including "not configured" and "not yet released" — is the actual product insight here. That is precisely when a student needs the explanation.

**On the authorization fix.** Good find, and I verified it is safe to make: `CourseStudent` was on `GET /config` while that endpoint returns the full Typst template, the updater's name and the release state, and the only caller of `getConfig` in this repo is the lecturer-only `SettingsPage`, so nothing student-facing breaks. Your reasoning for why it had to be fixed here rather than deferred is sound — adding a student-relevant field to that DTO would have turned a latent leak into a live one.

My one request: it currently lives inside `2afd79b4`, a feature commit titled "Store an instructor text for the student download page". An authorization tightening on an endpoint that was exposing a template to students should be its own commit, so it is findable by `git log` later and can be cherry-picked if it ever needs to go to a release branch independently. The PR does not need restructuring beyond splitting that one hunk out.

Leaving `CourseEditor` on `GET /config` and calling the editor/`/config/template` incoherence a product decision rather than settling it here is the correct call.

Two housekeeping items:

- The `github-advanced-security` comment on `StudentPageTextCard.tsx:28` is **stale** — `b9b5c21b` replaced the regex tag-strip with the inert `DOMParser` check. Worth confirming the alert auto-dismissed.
- This conflicts with #2108 (`modify/delete` on `e2e/seed/e2e_seed.sql`), as does #2133. Merge-order note only.

Also: `SanitizedHtml` is the better of the two instructor-HTML sinks in this batch, and I have pointed #2133 at it. When it eventually moves to `prompt-lib`, this is the version that should go.

Issue #1810, opened 2026-07-02
E2E: extend the seed to a realistic course setup


rappm:
## Context

The e2e seed (`e2e/seed/e2e_seed.sql`) currently contains 3 courses, 19 students, 2 course phase types, and exactly one course phase (an open application phase). There are no course participations and no course-scoped role assignments, so anything beyond read-only list views cannot be tested against it. Nearly all planned e2e coverage work depends on a richer seed.

## Scope

- [ ] Add a course with a full phase graph: application → interview → matching → team allocation → assessment
- [ ] Add course participations for seeded students
- [ ] Assign course-scoped Keycloak roles to the seeded `lecturer`, `course-lecturer`, `course-editor`, and `student` users (`e2e/keycloak/realm.json`)
- [ ] Update `e2e/src/data/constants.ts` with the new IDs
- [ ] Keep the seed a consistent dump of the current schema version so the server's startup `migrate up` stays a no-op (see `e2e/README.md`)
- [ ] Document the regeneration flow if it changes

## Definition of done

- `make test-e2e` is green with the new seed
- Constants cover every entity the planned suites need (courses, phases, participations, role users)

PR #2135, opened 2026-08-30
`Testing`: Add end-to-end coverage for the GDPR export and deletion flows

### LINKED ISSUE #1850: E2E: GDPR data export and deletion flow
## Context

The GDPR / data-privacy flow (subject data export and deletion) has no e2e coverage. It spans a student-facing request path and an admin review path in the management console, plus fan-out to course-phase modules, so it is worth exercising end to end in the seeded stack.

## Scope

Browser journeys against the seed:

- [ ] Student: initiate a data-export request; status banner reflects `pending` and updates to `completed`
- [ ] Student: download the produced export document via the signed download URL
- [ ] Student: initiate a data-deletion request; confirmation dialog and pending status banner
- [ ] Admin: privacy console lists export and deletion requests with correct status badges, pending sorted to the top
- [ ] Admin: approve a deletion request; verify the subject's core data is removed and the request reaches `completed` (poll the batch to completion)
- [ ] Admin: reject a deletion request; verify rejected state is distinct from the failure state
- [ ] Admin: admin-initiated deletion for inactive students (inactivity filter -> batch -> status polling)
- [ ] Admin: delete an export
- [ ] Role/permission check: a student cannot reach the `/admin/data-deletions` and `/admin/data-exports` endpoints
- [ ] Tests clean up any data they create (suite convention)

Non-Keycloak (external) subject requests can stay out of scope for a first pass.

Depends on #1810.

## Definition of done

- Journeys above are green in `make test-e2e` using the seeded `student`, `lecturer`, and admin roles
- Selectors live in Page Objects under `e2e/src/pages/`; `data-testid` added to the privacy console components where no stable selector exists

rappm:
## Summary

Adds end-to-end coverage for the GDPR data export and deletion flows: the student-facing request
pages, the admin privacy console, and the destructive paths that actually remove a subject's data.

### The journeys

- **`student-export.spec.ts`** — the overview page, requesting an export, the status banner while the
  work runs detached, polling to a terminal state, downloading a document through the real UI button,
  and an admin archiving the export with a rate-limit reset.
- **`deletion-approval.spec.ts`** — a student requests their own deletion, the admin console shows it
  as pending (and sorts pending first), an admin approves it, and the subject is genuinely gone from
  core afterwards.
- **`deletion-rejected.spec.ts`** — the same review dialog taking the other branch, asserting that the
  rejected state is distinct from the failure state, and that a rejected subject may request again.
- **`admin-initiated-deletion.spec.ts`** — the student table's "not modified in N years" filter, the
  `Delete Student Data` batch with its five-second countdown, polling to completion, and the subject's
  removal.
- **`api/privacy.api.spec.ts`** — `/api/privacy/admin/data-deletions` and `/api/privacy/admin/data-exports`
  answer 401 unauthenticated, 403 for a student and 200 for an admin, while a student can still reach
  their own endpoints.

### Fixtures the destructive paths need

Approving a self-service deletion request is only possible for a Keycloak-backed user, and it deletes
that student and cascades their participations. Doing it to `student` or `student2` would break the
application, assessment, matching, team-allocation and certificate suites. So the approval journey
owns **`privacy-student`**: a Keycloak user plus a student row and course participation that nothing
else references, existing solely to be deleted.

That fixture is consumed on first use, so the spec is **state-aware rather than skippable**. It reads
the subject's latest request and branches: no request → drive the whole flow; pending → approve;
in progress → poll; already succeeded → assert the terminal state and the subject's absence. A
`failed` run is never a passing branch — it is reported with the full payload, because a partial
failure can leave the subject either present or absent.

The inactivity journey gets its own **non-Keycloak** student with `last_modified` in 2015. A student
created during the run is stamped with the current time and can never match "not modified in 3 years",
so the filter would be untestable without a deliberately old fixture.

### Waiting

Both flows return 200 before a detached goroutine does the work, with a 30-minute server-side budget.
A shared helper polls the API on a bounded deadline, reports the last payload on timeout, and fails
immediately when it sees a failed run instead of waiting out the clock. Each affected spec raises its
own timeout, since Playwright's 60-second default would otherwise preempt the helper, and the budgets
are sized so that a test retried twice still cannot exhaust the shard's 20-minute global timeout.

Those polls also re-mint their token on a 401: the `prompt-client` Keycloak client caps access tokens
at **60 seconds**, so an `APIRequestContext` taken once in `beforeAll` starts failing partway through
any wait that outlives a minute — which reads like an auth bug rather than expiry.

Row locators are narrowed by status, because the console never removes terminal rows: a rejected
request and an archived export both stay listed, so on a CI retry a name-only locator matches the
previous attempt's row as well and every retry fails on a strict-mode violation.

### Selectors

Two `data-testid`/a11y additions to `clients/core`, where nothing stable existed: the privacy status
banner exposes its presentation state (`data-state`), and each export document's icon-only download
button gains an accessible name. Everything else uses existing roles and headings. Note the banner's
state is *presentation* state, not the backend enum: a `pending` export renders `in_progress` and a
`complete` one renders `success`.

## Known limitation, asserted rather than papered over

**Only the `Core` export document can succeed in the e2e stack.** `prompt-sdk` refuses a presigned
upload URL whose scheme is not HTTPS unless the host is `localhost`, and the e2e core signs against
`S3_ENDPOINT=http://seaweedfs-s3:8333`, so every phase module answers
`400 {"error":"upload URL must use HTTPS"}` and the export finishes as "completed with issues".
(`example_server` additionally has no `privacy/` package at all and answers 404.) Real deployments use
an HTTPS S3 endpoint, so this is an environment limitation rather than a product defect. The spec
therefore accepts the banner's `partial` state alongside `success` and asserts specifically that the
**Core** document is complete and downloadable, instead of quietly picking whichever document
happened to work. Making module fan-out work here would mean terminating TLS in front of SeaweedFS,
which is worth its own change.

## Cleanup, honestly

Exports are archived and their rate limit reset, so the suite is re-runnable. Deletion **request rows
are deliberately retained** — they are the audit trail of a GDPR request, there is no endpoint to
remove them, and an archived export keeps its metadata row by design (the console says so in its own
footnote). The suite therefore does not claim to leave no records; it leaves no *reusable state*.

## Reason / Link to issue

Closes #1850. Unblocked by #1810, which is merged.

Checklist from the issue:

- [x] Student: initiate an export; the banner reflects progress and updates on completion
- [x] Student: download the produced document via the signed URL
- [x] Student: initiate a deletion request; confirmation dialog and pending banner
- [x] Admin: the console lists exports and deletion requests with status badges, pending sorted first
- [x] Admin: approve a deletion; the subject's core data is removed and the request reaches success
- [x] Admin: reject a deletion; rejected asserted as distinct from failed
- [x] Admin: admin-initiated deletion for inactive students (filter → batch → polling)
- [x] Admin: delete an export
- [x] Role check: a student cannot reach the admin privacy endpoints
- [x] Tests clean up the state they create (see the note above on retained audit rows)

## How to Test

```bash
make test-e2e-shard PATHS="tests/privacy tests/api/privacy"
```

The suite joins the existing `core` shard; `e2e/shards.json` is updated so the partition check still
accounts for every spec.

## PR Checklist


mathildeshagl (approved):
This is a strong suite, and what makes it strong is that the hard parts are solved rather than avoided. Three in particular:

- Re-minting the token on a 401 inside `RoleApi` instead of taking one `APIRequestContext` in `beforeAll`. A 60-second access token lifespan against a wait that outlives a minute produces a failure that reads like an auth bug, and almost nobody finds that from the symptom.
- Narrowing row locators by status because the console never removes terminal rows. That is a strict-mode violation that only appears on retry, which is the worst kind to debug in CI.
- Asserting the `partial` banner state and specifically that the **Core** document is complete, rather than picking whichever document happened to work. The HTTPS/SeaweedFS limitation is a real environment constraint and you have written it down instead of letting the suite quietly assert less than it appears to.

The state-aware branching in place of skipping is the right choice for a consumed fixture, and refusing to treat `failed` as a passing branch is what keeps it honest.

Two things worth tightening:

- `admin-initiated-deletion.spec.ts`, the recovery branch: `all.some((request) => request.status === 'succeeded')` does not check which subject the succeeded request belongs to. `deletion-approval.spec.ts` creates a succeeded deletion for `privacy-student` in the same shard, so on a retry where the inactive subject's own deletion actually failed, this branch still passes. `student_id` is already in the type you declare on the line above — filtering on `request.student_id === SUBJECT.id` closes it.
- The seed change is global: two students and two course participations on iPraktikum `ios2425`, one of which is deleted mid-run. "How to Test" only covers `PATHS="tests/privacy tests/api/privacy"`. The full `core` shard is green in CI so this is largely answered, but it is the run worth quoting in the description, since the blast radius of the seed edit is wider than the specs it was added for.

One merge-order note, not a change request: #2108 deletes `e2e/seed/e2e_seed.sql` entirely. I test-merged the branches and it is a `modify/delete` conflict on that file. Whichever lands second has to port its fixtures into `seed/core.sql` — worth agreeing an order rather than discovering it at merge time.

The two client-side additions are minimal and correctly scoped, and the comment distinguishing the banner's presentation state from the backend enum is the kind of thing that saves the next person a wrong assumption.

PR #2136, opened 2026-08-31
`Interview`: Replace service singletons with constructors

### LINKED ISSUE #2102: `Interview`: Replace service singletons with constructors

### Scope

- **5** singleton variables, **33** references, **5** `InitXModule` functions.
- **8** singleton assignments across `_test.go` files.

```
interview/interviewSlot/service.go:28        InterviewSlotServiceSingleton
interview/interviewAssignment/service.go:24  InterviewAssignmentServiceSingleton
interview/config/service.go:14               ConfigServiceSingleton
interview/privacy/service.go:14              PrivacyServiceSingleton
interview/copy/service.go:15                 CopyServiceSingleton
```

### What to do

Three of the five (`config`, `privacy`, `copy`) are the generic sub-module scaffold that prompt-sdk#103 proposes to absorb entirely. Check that issue's status before rewriting them by hand, since the SDK helper may make the local versions disappear rather than change shape.

### Done when

- No `var XServiceSingleton` remains in `servers/interview`.
- `make test-interview` passes.

rappm:
## Summary

Removes every package-level service singleton from `servers/interview`. Each module now exposes a
`NewXService(...)` constructor and a separate `RegisterRoutes(...)`, so a service can be obtained
without also registering routes, and `main.go` is the only place that knows the wiring order.

## Details

`main.go` was deleted:

- `NewXService(queries, conn)` builds the service.

The package-level functions and gin handlers that read `XServiceSingleton.queries` / `.conn` became
methods on the service, using the receiver's fields. Every route path, HTTP method, middleware,
middleware order and role list is unchanged, including the student-facing ownership check on
`DELETE /interview-assignments/:assignmentId`.

Per module:

- `interviewSlot`, `interviewAssignment`: constructor takes queries and the pool.
  `fetchAllStudentsForCoursePhase` stays a package function, since it talks to core over HTTP and
  holds no state.
- `interviewReview`: never held a pool, so `NewInterviewReviewService(queries)` takes queries alone.
  `errScoreOutOfRange` moves from the deleted module file to `service.go`.
- `privacy`: gained a `router.go` with `RegisterRoutes`, and the two package-level handlers became
  `DataExportHandler` / `DataDeletionHandler` methods. The SDK takes these as function types rather
  than through an interface, so the method values are passed directly and no adapter type is needed.
- `config`, `copy`: `ConfigService` and `CopyService` each held a `db.Queries` and a `*pgxpool.Pool`
  that nothing ever read. The interview phase has nothing to configure and copies nothing between
  phases, so both handlers return a constant. Both services, their singletons and their module files
  are gone, and the two handlers are stateless package-private structs.

The copy handler was named `SelfTeamCopyHandler`, a copy-paste from the self team allocation
service. It is now `interviewCopyHandler`.

The test suites that used to assign a singleton now construct the service they need. The four
`config` and `copy` suites spun up a Postgres testcontainer solely to fill the unread singletons;
without them those suites need no database at all and run as plain tests.

This follows `servers/presentation` (the reference implementation) and #2116, which did the same for
`servers/certificate`.

## Reason / Link to issue

Closes #2102. Part of #2092.

## How to Test

1. `make test-interview`
2. `cd servers/interview && go build ./... && go vet ./...`
3. Start the service and confirm the interview phase still loads: slots, assignments, reviews, the
   score and score level data-graph outputs, phase config, phase copy, and the privacy
   export/deletion endpoints all behave as before.

## PR Checklist


mathildeshagl (approved):
Fifth in the series and the same verification holds: I extracted every route registration and every `.Group()` call before and after and diffed them, and they are identical once the receiver is accounted for — paths, methods, middleware order and role lists, including the student-facing `DELETE /interview-assignments/:assignmentId` whose privileged/self branch is untouched. No `Singleton` or `Init*Module` identifier survives in the service. Builds and vets clean here, and the whole matrix is green including `test-servers / test (interview)`.

I also scanned the diff for anything that is not a receiver rewrite. The only non-mechanical additions are the two handler type renames, the `errors` import that follows `errScoreOutOfRange` into `service.go`, and the rewritten test scaffolding. Nothing changed behavior.

Two things I think are genuinely good rather than just correct:

- Deleting `ConfigService` and `CopyService` instead of giving constructors to a `db.Queries` and a `*pgxpool.Pool` that nothing read. The follow-on is the better half: those four suites were booting a Postgres testcontainer purely to populate fields no code touched, and they now run as plain tests. That is real time back on every `make test-interview`.
- Catching `SelfTeamCopyHandler`. A copy-paste type name from another service is exactly the kind of thing that survives forever because it compiles.

One nit: `PrivacyService.Queries` stays exported while `conn` beside it is not. Nothing outside the package reads it — `main.go` only calls `NewPrivacyService` and `RegisterRoutes` — and #2129 unexported the equivalent field on its own privacy service in this same series. Lowercasing it here would finish that consistency.

No file overlap with any other open PR in the series, so this one carries none of the merge-order caveats I left on #2119, #2128 and #2129.

PR #2137, opened 2026-08-31
`New Phase`: Replace the example server service singletons with constructors

### LINKED ISSUE #2092: `General`: Service construction is welded to route registration behind package-level singletons

### LINKED ISSUE #2104: `Example`: Replace service singletons with constructors so new phases stop inheriting them

### Scope

- **3** singleton variables, **6** references, **3** `InitXModule` functions.
- **0** test files. `servers/example_server` has no tests at all.

```
example_server/config/service.go:22   ConfigServiceSingleton
example_server/example/service.go:16  ExampleServiceSingleton
example_server/copy/service.go:26     CopyServiceSingleton
```

### Why this is not the least important one

By raw count this is the smallest instance in the repository. By reach it is the largest, because `make new-phase` copies this service wholesale:

```
scripts/new-course-phase.sh:170   cp -R servers/example_server "$SERVER_DIR"
```

Every course phase scaffolded from now on inherits whatever shape `example_server` has, singletons included. Leaving it converted last means new phases keep arriving with the pattern the other sub-issues are removing. It is arguably worth doing **first** rather than last, as the reference the other conversions are checked against.

The same argument applies to `clients/example_component`, which the generator copies on line 25.

### What to do

`servers/presentation` is the reference implementation.

Two of the three (`config`, `copy`) are the generic sub-module scaffold that prompt-sdk#103 proposes to absorb entirely. Check that issue's status first.

### Done when

- No `var XServiceSingleton` remains in `servers/example_server`.
- A phase scaffolded with `make new-phase` compiles and starts with the new shape.
- `make test-example` passes.

### Note

Because this service has no tests, the conversion is unverified by anything except the generator smoke test in `.github/workflows/generator-smoke.yml`. Worth confirming that workflow actually exercises the generated server's startup path.

rappm:
## Summary

Removes every package-level service singleton from `servers/example_server`, and updates the two
scaffolding scripts, the service README and the Go rules to match. This service is what
`make new-phase` copies wholesale, so until it is converted every newly scaffolded phase arrives
with the pattern the rest of #2092 is removing.

## Details

`main.go` was deleted:

- `NewXService(queries, conn)` builds the service.

The handlers and business functions became methods on the service. `main.go` constructs all three
services in one block, then registers every route group. Every route path, HTTP method, middleware
and role list is unchanged.

- `example`: `GetExampleInfo` now takes a `context.Context` rather than the gin context, which is
  the split the other services use — gin stops at `router.go`.
- `config`, `copy`: `promptTypes.PhaseConfigHandler` and `promptTypes.PhaseCopyHandler` are
  structural interfaces, so `ConfigService` and `CopyService` implement them directly and the
  registration passes the service itself. `ExampleServerConfigHandler` and
  `ExampleServerCopyHandler` are gone: they held no state, and a zero-value handler is exactly the
  shape a scaffolded phase should not start from, because it compiles and then dereferences nil at
  request time once someone fills in the body.

Both services keep their `queries` and `conn` fields even though the placeholder handlers do not
read them yet. Unlike the interview service (#2136), which had the same two unused services and
lost them, their whole job here is to show a phase author where the queries and the pool belong.

Scaffolding and docs:

- `scripts/new-course-phase.sh` and the external-phase `template-repository/init.sh` renamed
  `InitExampleModule`, `setupExampleRouter`, `ExampleServerConfigHandler` and
  `ExampleServerCopyHandler` into the new phase's names. None of those identifiers exist any more,
  so those substitutions are dropped. `ExampleService` still needs renaming and now also covers
  `NewExampleService`; `RegisterRoutes` is package-scoped and needs none. The
  `exampleService` local variable that `main.go` now declares needs its own camelCase
  substitution: the `\bexample\b` fallback does not fire inside a compound identifier and the
  `ExampleService` rule is case-sensitive, so without it a scaffolded phase keeps the literal
  token. The generator smoke test catches exactly this.
- The service README's file map described `example/main.go` and "creates the service singleton".
- `.claude/rules/go/patterns.md` pointed at `example_server`'s `config/` and `copy/` packages as the
  pattern to copy while those packages still held singletons and zero-value handlers, so the
  guidance and the reference disagreed. It now records three rules, each with the case where the
  exception applies: construction is separate from route registration, handlers are methods on the
  service they use, and the SDK config and copy interfaces are implemented on the service rather
  than on a wrapper.

## Reason / Link to issue

Closes #2104. Last issue of and closes #2092.

## How to Test

1. `cd servers/example_server && go build ./... && go vet ./...`
2. `make new-phase NAME=scratch CLIENT_PORT=3099 SERVER_PORT=8099` on a throwaway branch, then
   confirm `servers/scratch` builds and contains no `Singleton` or `InitScratchModule`.
3. Start the service and confirm `GET /example-service/api/course_phase/{id}/info` still answers,
   and that the config and copy endpoints still return their placeholder 404.

## PR Checklist


.claude/rules/go/patterns.md
```diff
@@ -7,13 +7,29 @@ paths:
 
 - **Prefer `prompt-sdk` over custom implementations** for auth, CORS, inter-service fetching,
   validation, and shared domain types — see `prompt-sdk.md` for the API catalog.
+- **Construction is separate from route registration.** A module exposes
+  `NewXService(queries, conn, deps...)` returning the service, and
+  `RegisterRoutes(routerGroup, service, authMiddleware)` mounting it. No package-level
+  `XServiceSingleton`, no `InitXModule` that does both — a test must be able to obtain a service
+  without building a router. `main.go` constructs every service in one block, then registers every
+  route group, and is the only place that knows the wiring order. Reference: `servers/presentation`,
```
mathildeshagl, line 15: I checked the rule against what the services actually do, and the code in this PR matches the repo; it is this reference line that is slightly off.

`servers/presentation` exposes `presentation.NewService` and `RegisterRoutes(router, service)`. Everywhere else — assessment, interview, certificate, team_allocation, self_team_allocation — it is `NewAssessmentService` / `NewConfigService` / `NewCopyService` with `RegisterRoutes(group, service, authMiddleware)`, which is exactly the shape this PR gives `example_server` and exactly what `NewXService(...)` above describes. So `presentation` is the one service that does not follow the sentence it is cited for.

Either drop it and point at `servers/assessment` (or `servers/certificate`, which has the same `config`/`copy` pair as the example), or keep it and note that the bare `NewService` form is the exception. Documentation nit only — nothing in the Go changes here.

For the record, the rest of the rule checks out: `ServiceSingleton` and `InitXModule` appear nowhere in `servers/` after this PR, so `example_server` really was the last holdout and the rule now describes the whole tree.

servers/example_server/example/service.go
```diff
@@ -8,13 +8,23 @@ import (
...
+// ExampleService holds everything the module's business logic needs. Collaborators
+// arrive through the constructor, so a test can build the service on its own and
+// `main.go` stays the only place that knows the wiring order.
 type ExampleService struct {
 	queries db.Queries
 	conn    *pgxpool.Pool
 }
 
-var ExampleServiceSingleton *ExampleService
+func NewExampleService(queries db.Queries, conn *pgxpool.Pool) *ExampleService {
```
mathildeshagl, line 19: The stated reason for the seam is that "a test must be able to obtain a service without building a router" — and `example_server` still has no `*_test.go` anywhere, so the template advertises a testing seam it never demonstrates.

One small test against `NewExampleService` would fix that: it is the file a new phase author copies, and right now the scaffolding hands them a constructor with nothing showing what it is for. Something as thin as constructing the service with a zero `db.Queries` and a nil pool and asserting `GetExampleInfo` returns its placeholder string would document the pattern and give `make test` something to run for this service.

Entirely optional, and I would not hold the PR for it — the singleton removal stands on its own. Also verified the two rename scripts stay in sync: `CAMEL` is defined in both (`scripts/new-course-phase.sh:45`, `template-repository/init.sh:34`), the removed token rules have no remaining referents, and `template-repository/repo` carries only infra files since the Go source is extracted from `servers/example_server` at migration time.

mathildeshagl (approved):

PR #2140, opened 2026-08-31
`General`: Use the SDK student-ownership helper instead of per-service copies

### LINKED ISSUE #1984: Adopt the SDK student-ownership helper
## Summary

Delete the duplicated student-ownership helper and use the SDK version.

Blocked on prompt-edu/prompt-sdk#104 (SDK-side helper).

## What is duplicated (verified, 2 services)

`utils/ownership.go` is identical in `assessment` and `interview` except one error message string:

- `servers/assessment/utils/ownership.go:1-60`
- `servers/interview/utils/ownership.go:1-60`

Exports `GetUserCourseParticipationID`, `GetUserCourseParticipationIDErrorStatus`,
`ValidateStudentOwnership`, and four sentinel `Err*` vars.

## Work

- Delete both local files, import the SDK helper, and pass the per-service entity name for the one
  mismatched error message.

rappm:
## Summary

Deletes the duplicated student-ownership helper from the `assessment` and `interview` services and
uses the shared implementation that `prompt-sdk` v1.3.1 provides instead.

## Details

- Bumped `github.com/prompt-edu/prompt-sdk` from `v1.3.0` to `v1.3.1` in `servers/assessment/go.mod`
  and `servers/interview/go.mod`. The indirect upgrades in the `go.sum` diffs
  (`testcontainers-go` v0.44.0, the OpenTelemetry stack, and their transitive dependencies) come
  from the SDK's own requirements via MVS and are a faithful `go mod tidy` result, not hand edits.
- Deleted `servers/assessment/utils/ownership.go` and `servers/interview/utils/ownership.go`. The two
  files were identical apart from a single error string. `servers/interview/utils/` held nothing else,
  so the package and its test are gone; `servers/assessment/utils/` keeps its remaining files.
- Repointed every call site at `keycloakTokenVerifier.GetUserCourseParticipationID`,
  `GetUserCourseParticipationIDErrorStatus`, and `ValidateStudentOwnership`.
- The SDK's `ValidateStudentOwnership` takes the owned entity name as a third argument, which is what
  the duplication existed to vary. The evaluation-completion routes pass `"evaluation completions"`,
  preserving the current message verbatim. `createOrUpdateEvaluation` passes `"evaluations"`: it acts
  on evaluations, and the old shared sentinel could not tell the two apart.
- The `interview` copy's `ErrOwnershipValidation` had no call sites, so no interview-side message
  changes.

## Reason / Link to issue

Closes #1984.

The SDK-side helper it was blocked on shipped in `prompt-sdk` v1.3.1.

## How to Test

1. `cd servers/assessment && go build ./... && go vet ./... && go test ./...`
2. `cd servers/interview && go build ./... && go vet ./... && go test ./...`
3. As a student, hit an evaluation-completion route with another participant's
   `authorCourseParticipationID` and confirm the response is still `403`.

## PR Checklist


mathildeshagl (approved):
Checked this as a behavior-preservation question rather than reading the diff for shape, since it is auth code.

The SDK v1.3.1 implementation is the same logic as the deleted local copies, statement for statement: same three sentinel errors, same `GetUserCourseParticipationIDErrorStatus` mapping (401 / 400 / 500), same `uuid.UUID`-then-`string` type assertion fallback in `GetUserCourseParticipationID`, same `403` on mismatch. The only difference is the entity name moving from a package-level sentinel to a parameter, which is what the duplication existed to vary.

The one thing that could have broken quietly is that the local `ErrOwnershipValidation` was a comparable sentinel and the SDK returns `fmt.Errorf`, so `==` and `errors.Is` against it no longer work. I grepped the base for it: nothing outside the deleted files ever referenced it, so there is no call site to update. Worth stating explicitly since it is the kind of thing that compiles either way.

Every call site is repointed, the `"evaluation completions"` message is preserved verbatim, and `"evaluations"` on `createOrUpdateEvaluation` is a message change that makes it more accurate — the old shared sentinel could not distinguish them, exactly as you say. The dedup is also complete: no other service carries a copy, `servers/interview/utils/` is gone entirely, and `servers/assessment/utils/` keeps only its two unrelated files.

One note: this introduces SDK version skew. All eight services were uniformly on `v1.3.0`; assessment and interview are now `v1.3.1`, and #2141 makes `team_allocation` a third. The `go.sum` churn (testcontainers v0.44.0, the OTel stack) lands only in those services too, so their test-dependency versions diverge from the other five. Nothing breaks — the SDK is consumed per-module — but a repo that was uniform no longer is, and it is easier to reason about a single SDK version across the services. If the remaining five are getting bumped anyway, doing it in one sweep would be tidier than letting the skew persist between feature PRs.

Issue #984, opened 2025-12-08
Course Phase Deletion Problems


JGStyle:
Course phase deletion right now only deletes the course_phase record in the database and yields 500 status code

Deleting a course phase in reality should also delete the data from the course phase participants and so on - But we have to consider what happens when you delete a course phase that has a following course phase, and especially what happens when you delete the application coursephase

JGStyle:
Okay, so i've researched this about.

Right now when creating a course phase in the course phase graph, the only thing that happens is that the course phase records in core's main db is created and the phase graph is set up as connected in the brwowser, no communication with the affected course phase at all.

The same is for deletion: as the issue states, only the rows on cores main db are removed. the course phase doesn't know about this.

I guess an easy solution would be : if a course phase is created or deleted, simply also call the corresponding course phase at a defined endpoint. 

For this we should best first define the endpoint with a default handler in the sdk, then implement that in all participating CPMs. the core client, when submitting the deletion request could also submit the deletion request to the course phase servers
@prompt-edu/prompt-developer

JGStyle:
This is a bigger change and more of a feature

JGStyle:
After discussion: the plan is accurate. We will use these route schemes

* send (POST /) DELETE on course phase route (NOT `/create` and `/delete`)
* possibly add capability to /info

PR #2141, opened 2026-08-31
`Team Allocation`: Delete course phase data when a phase is removed

### LINKED ISSUE #1866: Team Allocation: Course phase deletion implementation
once https://github.com/prompt-edu/prompt-sdk/issues/98 is completed, 

we need to implement the newly introduced handler for course phase deletion.

As outlined in https://github.com/prompt-edu/prompt/issues/984, the idea is that the moment a course phase is permanently deleted from the configurator, this handler will be called by the core client to give this cpm a chance to delete related data.

In here we will implement to delete everything related to the course phase that is incoming

rappm:
## Summary

Implements the SDK's course phase deletion handler in the `team_allocation` service, so the service
drops everything it stores for a course phase the moment that phase is permanently deleted.

## Details

- Bumped `github.com/prompt-edu/prompt-sdk` from `v1.3.0` to `v1.3.1` in `servers/team_allocation`,
  which is where `CoursePhaseDeletionHandler`, `RegisterPhaseDeletionModule`, and
  `CapabilityPhaseDeletion` live.
- New `servers/team_allocation/coursePhaseDeletion/` package, laid out like the existing `privacy/`
  package: `service.go` holds the handler, `router.go` registers it through
  `promptTypes.RegisterPhaseDeletionModule`. The SDK protects the endpoint itself (PromptAdmin and
  CourseLecturer), so no roles are passed. It is mounted on `coursePhaseApi`, the group carrying
  `:coursePhaseID` that the SDK registrar requires.
- New `db/query/coursePhaseDeletion.sql` deletes the four phase-scoped root tables (`team`, `skill`,
  `survey_timeframe`, `tease_workspace`) inside one transaction, with regenerated `db/sqlc/` output.
  No migration is involved.
- `promptTypes.CapabilityPhaseDeletion: true` added to the service's `ServiceInfo`.

### Why the child tables are not deleted explicitly

`student_team_preference_response`, `student_skill_response`, `allocations`, and `tutor` are removed
by the `ON DELETE CASCADE` constraints declared in `db/migration/`. This is safe rather than merely
convenient: `allocations` and `tutor` both hold a composite foreign key
`(team_id, course_phase_id) REFERENCES team (id, course_phase_id)` with `team_id NOT NULL`, so a row
of a given phase cannot exist without a team of that same phase. There is no orphan path. The two
root queries carry a comment naming the tables that cascade off them, with a note to add an explicit
delete if a future migration drops one.

### Note on scope

The endpoint is not yet reachable in practice: `servers/core/coursePhase/service.go`
`DeleteCoursePhase` still only deletes the phase row and does not dispatch to phase modules, and
nothing reads the `phase.deletion` capability. That core-side orchestration belongs to the parent
issue #984 and is deliberately out of scope here. This PR builds the receiving end and establishes
the pattern for the sibling issues #1867 to #1870.

Separately, the blocker this issue was filed against, `prompt-edu/prompt-sdk#98`, shipped in
prompt-sdk v1.3.1 and can be closed.

## Reason / Link to issue

Closes #1866.

## How to Test

1. `cd servers/team_allocation && go build ./... && go vet ./... && go test ./...`
2. The new suite in `coursePhaseDeletion/service_test.go` covers the behavior end to end against a
   seeded testcontainer: a phase's own data is removed together with the cascaded child rows, a
   second course phase is left completely untouched, running the deletion twice succeeds, and
   deleting a phase that never stored anything succeeds.
3. Manually: `POST /team-allocation/api/course_phase/<coursePhaseID>/delete` as a PromptAdmin, then
   confirm the phase's teams, skills, survey timeframe, and TEASE workspace are gone while other
   phases still have theirs.

## PR Checklist

## Summary by CodeRabbit

* **New Features**
  * Added support for deleting all team-allocation data associated with a course phase.
  * Deletion includes teams, skills, surveys, workspaces, and related records.
  * Added the course-phase deletion capability to the team allocation service.
* **Bug Fixes**
  * Deletion runs atomically and safely handles repeated requests or phases without stored data.
* **Tests**
  * Added coverage confirming targeted data is removed while other course phases remain unaffected.


mathildeshagl (approved):
This is the one in this batch where the description could have been taken on trust and should not be, so I checked the two claims it rests on.

**The endpoint is protected.** `RegisterCoursePhaseDeletionEndpoint` in the SDK applies `keycloakTokenVerifier.AuthenticationMiddleware(PromptAdmin, CourseLecturer)` before the handler, and it mounts on `coursePhaseApi`, so `POST /team-allocation/api/course_phase/:coursePhaseID/delete` sits under the required prefix with the required roles. Passing no roles from our side is correct rather than an omission.

**The cascade reasoning is right, and it is the load-bearing part.** Six tables carry `course_phase_id` — `team`, `skill`, `survey_timeframe`, `tease_workspace`, `allocations`, `tutor` — and the four explicit deletes plus the cascades cover all of them, along with `student_team_preference_response` (0001:16) and `student_skill_response` (0001:36). I checked the orphan argument specifically: `allocations.team_id` is `UUID NOT NULL` with both `team_id` and `(team_id, course_phase_id)` cascading, and `tutor.team_id` is `uuid NOT NULL` with the composite cascade. No migration later relaxes either. So a row of a phase genuinely cannot outlive that phase's teams, and there is no path that leaves student data behind. For a deletion handler that is the whole question, and the answer is sound.

The query comments naming what cascades off each root delete, with the instruction to add an explicit delete if a future migration drops one, are the right thing to leave behind. So is `countRows` querying the raw tables in the test — the comment explaining that the generated queries join their parent and therefore cannot distinguish a cascade from an orphan is exactly the failure this suite needs to catch.

Two notes, neither blocking:

- This is the only suite in `team_allocation` using `SetupTest` rather than `SetupSuite`, so it boots a Postgres testcontainer per test method — four containers for four tests, where the nine sibling suites boot one each. Per-test isolation is defensible for destructive tests that mutate the fixture, but it is a deviation worth a sentence in the file, and it is real CI time.
- `db.GetTimeoutContext` is 10 seconds. A phase with enough teams and allocations to exceed that fails the whole handler. That is safe, because the work is in one transaction and rolls back cleanly, so there is no partial deletion — but it does mean the core-side orchestration in #984 has to treat a timeout as retryable rather than as "module reported done". Worth carrying into that issue.

Incidentally, bounding the context here is stricter than the sibling `privacy` deletion handler, which calls `s.conn.Begin(c)` on the request context directly. That one is arguably the one that should change.

PR #2142, opened 2026-08-31
`Self Team Allocation`: Fix the oversized create team button and empty state padding

### LINKED ISSUE #1734: Oversized button and inconsistent empty-state padding
Image

On the Team Allocation page the **"Create new team" button is overly large**. In the "No Teams Available" empty state, the **bottom padding is inconsistent with the top padding** (likely caused by the icon spacing). Resize the button and balance the empty-state padding.

rappm:
## Summary

Sizes the "Create new team" button to its content and balances the padding of the "No Teams
Available" empty state on the self team allocation page.

## Details

- `TeamCreationDialog.tsx` — the trigger carried `w-full max-w-lg mx-auto` with no `sm:` reset, so
  the width applied in the row layout too and stretched the button to 512px beside the auto-width
  "Refresh Teams" button. Removing the three classes lets it size to its content like its siblings.
  The surrounding container is `flex flex-col sm:flex-row items-start`, so the mobile column layout
  is unaffected: `items-start` already made every child content-width, and the create button was the
  only outlier.
- `TeamSelection.tsx` — the empty state wrapper is `p-12` and the icon carried `mb-4 mt-4`. Margins
  do not collapse inside a flex container, so the icon's `mt-4` stacked onto the padding and left
  64px above the icon against 48px below the text. Dropping `mt-4` makes it symmetric. `mb-4` stays,
  as it is the icon-to-heading spacing.
- `TeamCreationDialog.tsx` and `TutorImportDialog.tsx` — the leading icons carried `mr-2` on top of
  the `Button` primitive's own `gap-2`, putting the icon 16px from its label where the design system
  intends 8px. The neighbouring "Refresh Teams" button already used a bare `h-4 w-4`, so this makes
  the row consistent.

### Note on the component

The issue is labelled `Component: Team Allocation`, but the strings in its screenshot ("No Teams
Available", "Refresh Teams", "Create new team") belong to `self_team_allocation_component`, which is
where the fix lands. `team_allocation_component` has a different empty state ("No Teams Created
Yet") and is not affected.

`clients/team_allocation_component/.../TutorImportDialog.tsx:90` does still have the same
icon `mr-2` double-gap. It is a separate micro-frontend and not what the issue reports, so it is left
for a follow-up rather than swept in here.

## Reason / Link to issue

Closes #1734.

## How to Test

1. Open the self team allocation phase as a course participant with no teams created yet.
2. The "Create new team" button should be as wide as its label, sitting beside "Refresh Teams"
   rather than spanning the row.
3. The "No Teams Available" box should have equal space above the icon and below the description.
4. Narrow the window below the `sm` breakpoint and confirm the buttons stack left-aligned.

## Screenshots

| Before | After |
| ------ | ----- |
| before | Button sizes to its label, empty state insets match. |

## PR Checklist


mathildeshagl (approved):
All three fixes are correct and the diagnosis of each is right. `w-full max-w-lg mx-auto` with no `sm:` reset genuinely did stretch the trigger to 512px beside an auto-width sibling; margins genuinely do not collapse into flex padding, so `mt-4` on top of `p-12` gave 64px against 48px; and `mr-2` on top of the `Button` primitive's own `gap-2` genuinely doubles the icon gap. I also confirmed the leftover you call out is real — `team_allocation_component`'s `TutorImportDialog` still has the same `mr-2 h-4 w-4` — and leaving it for a separate micro-frontend is the right scope call.

**One correction to the description.** It says the mobile column layout is "unaffected" because `items-start` already made every child content-width. `items-start` is the default the container establishes, but `w-full` explicitly overrode it on this child. So before this change the create button was full-width (to `max-w-lg`, centred) in the mobile column, and after it is content-width. That is a visible change on mobile, not a no-op.

It may well be the change you want — it makes the button consistent with `Import Tutors` and `Refresh Teams` beside it, and the issue is about sizing. But a full-width primary action on narrow viewports is also a common deliberate pattern, and "unaffected" would stop a reviewer from thinking about which one is intended.

Which is the case for the screenshots. This is a three-file, visual-only PR whose entire content is pixel judgements, the checklist item is unticked, and there is now one claim in the description that a screenshot would have caught. Two before/after captures of that button row at mobile and desktop width would make this self-evidently right instead of argued.

PR #2143, opened 2026-08-31
`Core`: Tighten the sidebar navigation spacing

### LINKED ISSUE #1730: Sidebar navigation: wasted space and inconsistent spacing
The left navigation sidebar wastes vertical space (e.g. the "Courses" section is a single small item with a large empty area) and has inconsistent spacing between "Overview" and the "Course Phases" list. Tighten the layout and spacing.

Image

Image

rappm:
## Summary

Tightens the management console sidebar: evens out the gap between "Overview" and the "Course
Phases" list, and reclaims the vertical space the general sidebar's group headings were spending.

### Inconsistent spacing above "Course Phases"

`SidebarGroup` from `@tumaet/prompt-ui-components` carries a blanket `p-2`, so between two adjacent
groups the whitespace stacked from four sources: the first group's 8px bottom padding, the
`SidebarMenu` `gap-1`, the second group's 8px top padding, and the label's own `h-8` half-leading.
That put 28px above the label text against 8px below it, which is the asymmetry the issue reports.

Collapsing the facing paddings (`pb-0` on the first group, `pt-1` on the second) brings it to 16px
above and 8px below. The label keeps more room above than below so it still reads as heading the
phase list rather than floating equidistant between the two.

### Wasted space in the general sidebar

`InsideSidebarVisualGroup` set symmetric `mt-1 mb-1` on its heading and then offset the text
downwards with `transform translate-y-1`. Painting the text 4px below its own box meant the visible
result was roughly 28px above the label and 1px below it, so each heading both wasted space and sat
against its own items. Dropping the transform and the top margin, and tightening the group gap from
`gap-5` to `gap-3`, gives 12px above and 5px below and saves 28px across a full lecturer sidebar.

The change is class-only. No colors, borders, or backgrounds were touched, so both themes are
unaffected.

### Verified

`InsideSidebar` renders `Sidebar` with `collapsible='none'` and `AppSidebarContent` only mounts it
when expanded or on mobile, so neither component ever renders under a `data-collapsible="icon"`
ancestor and `SidebarGroupLabel`'s icon-mode `-mt-8` never engages. The mobile `Sheet` branch renders
the same markup, and the diff is vertical-only. The Playwright selectors for these elements are
role- and text-based, and there are no screenshot snapshots in `e2e/`.

## Reason / Link to issue

Closes #1730.

## How to Test

1. Open a course in the management console. "Overview" and the "Course Phases" block should read as
   evenly spaced, with the label clearly attached to the phases below it.
2. Go back to the course list, where the sidebar shows only the "COURSES" group. The heading should
   sit tighter to its item, and as a lecturer or admin the "STUDENTS" and "ADMIN" groups should be
   closer together.
3. Check both light and dark theme, and narrow the window so the sidebar renders in its mobile sheet.

## Screenshots

| Before | After |
| ------ | ----- |
| before | Even rhythm, label attached to the phase list. |

## Notes for the reviewer

Two things were deliberately left out of scope:

- The empty region at the bottom of the second screenshot is inherent to a navigation with one
  visible link. No spacing change reclaims it without adding content or restructuring navigation.
- The two sidebars still use different heading mechanisms: `InsideCourseSidebar` uses the shared
  `SidebarGroupLabel`, while `InsideGeneralSidebar` uses a bespoke `h3`. Replacing the `h3` with the
  primitive would make them consistent by construction rather than through two hand-tuned margin
  sets, but it also changes the label's weight and color, which is beyond the spacing fix this issue
  asks for. Worth a follow-up.

## PR Checklist


mathildeshagl (approved):
Class-only, both themes untouched, and the analysis of where the whitespace was coming from is careful — four stacked sources between adjacent `SidebarGroup`s is exactly the kind of thing that gets "fixed" by adding a negative margin instead of by removing the facing paddings. Collapsing them with `pb-0` and `pt-1` is the right lever, and keeping more room above the label than below it so it still reads as heading the phase list rather than floating between two groups is the correct instinct.

Dropping `transform translate-y-1` is the real fix in the second half. Painting text 4px below its own box is why the heading both wasted space above and sat against its own items, and no amount of margin tuning would have resolved that while the transform was there.

I also appreciate that you checked `SidebarGroupLabel`'s icon-mode `-mt-8` cannot engage, since `InsideSidebar` renders with `collapsible='none'` and `AppSidebarContent` only mounts it expanded or on mobile. That is the non-obvious way this kind of change breaks, and it is the thing worth verifying.

One note, the same one I left on #2142: every claim in this PR is a pixel measurement — 28px above against 8px below, 12px against 5px, 28px saved across a lecturer sidebar — and none of it is checkable by a reviewer without building and running the console. The `gap-5` to `gap-3` change also affects every group in the general sidebar, not just the ones with the transform. Two before/after screenshots of each sidebar would turn this from a description I am trusting into one I can confirm, and it is the same unticked checklist item.

Approving on the reasoning, which I have no specific doubt about.

PR #2149, opened 2026-09-01
`Core`: Consolidate the client cache keys and request modules

### LINKED ISSUE #2097: `General`: Client cache keys and request modules are shallow
### Problem or opportunity

The client data-access code in `assessment_component` and `core` is spread across many modules that each hold one fact and state no policy. Two symptoms, same seam.

**Cache keys are string literals with no owner.**

| Workspace | `queryKey` sites | `invalidateQueries` calls | Files invalidating | Distinct key namespaces |
| --- | --- | --- | --- | --- |
| `assessment_component` | 104 | 71 | 30 | 28 |
| `core` | 100 | 46 | 29 | 28 |

No workspace in the repository has a module that owns these keys. Which caches a write invalidates is knowledge held in 30 files and nowhere else. In the assessment component the same five-key fan-out is copy-pasted into six hooks:

```ts
queryClient.invalidateQueries({ queryKey: ['categories', phaseId] })
queryClient.invalidateQueries({ queryKey: ['selfEvaluationCategories', phaseId] })
queryClient.invalidateQueries({ queryKey: ['peerEvaluationCategories', phaseId] })
queryClient.invalidateQueries({ queryKey: ['tutorEvaluationCategories', phaseId] })
queryClient.invalidateQueries({ queryKey: ['assessments'] })
```

`useCreateCategory`, `useUpdateCategory`, `useDeleteCategory`, `useCreateCompetency`, `useUpdateCompetency`, `useDeleteCompetency`, and `useImportAssessmentSchema` all carry a copy. A seventh cache that depends on Category means editing every one of them, and a typo in a literal is a silent stale read rather than a build error.

**Request modules are one file per endpoint.**

| Workspace | Query modules | Mutation modules | LOC |
| --- | --- | --- | --- |
| `assessment_component` | 29 | 31 | 1,129 |
| `core` | 30 | 29 | 1,702 |

Each file exports one function that builds a URL, calls axios, and in most cases wraps the call in an identical `try / console.error / throw`. 52 files in the assessment component and 57 in core carry that block. Callers must learn 60 separate names with independently chosen parameter orders, and no file states the host, the header policy, or the error policy, because there is no file whose job that is.

The cost of having no such file is already visible. `network/mutations/sendEvaluationReminder.ts` is the only one of the 60 assessment modules that uses core's shared `axiosInstance` instead of `assessmentAxiosInstance`, so it resolves against a different host than its 59 siblings. It works today, and nothing would have flagged it.

### Desired outcome

- One module per workspace owning cache keys and the invalidation fan-out, with named events rather than repeated literals.
- One module per workspace owning the route templates, host selection, headers, and error handling.
- Adding an endpoint or a cache dependency is a change in one place.

### Proposed solution

Per workspace, two changes, in this order:

1. A cache module exposing typed key builders and named invalidation events (`schemaChanged`, `assessmentWritten`, `completionToggled`). Hooks call the event; the fan-out table lives in the module.
2. A request module namespaced by domain noun (`assessmentApi.evaluations.listMine(phaseId)`), over a shared request helper that owns host, headers, and error handling.

Tracked per workspace in the sub-issues, because both changes touch the same 30 files and splitting them across people would guarantee conflicts. The other seven workspaces have between 1 and 29 `queryKey` sites and are deliberately out of scope: a key module there would not earn itself.

### Additional context

Counts measured against `072b89ed`. Worth doing after #2090, since none of this is currently observable by anything but Playwright.

### LINKED ISSUE #2099: `Core`: Consolidate cache keys and request modules
Part of the client data-access work tracked in the parent issue.

### Scope

- **100** `queryKey` sites, **46** `invalidateQueries` calls across **29** files, **28** distinct key namespaces.
- **59** request modules (30 queries, 29 mutations), **1,702** LOC, **57** files carrying the same `try / console.error / throw`.

`core` is the largest instance of this pattern in the repository, slightly ahead of the assessment component.

### What to do

Two PRs, in this order.

**1. Cache module.** Typed key builders plus named invalidation events, replacing the bare literals at the call sites.

**2. Request module.** Namespaced by domain noun over a shared request helper owning host, headers, and error handling.

Because `core` is the shell that hosts every remote, land #2098 first and reuse the shape that comes out of it rather than designing the two independently.

### Done when

- No bare cache-key string literal outside the cache module.
- No route template or `console.error` block outside the request module.
- Both modules have unit tests.

rappm:
## Summary

Gives `core` the two modules the assessment component got in #2112: one that owns every react-query cache key and the fan-out each write performs, and one that owns every endpoint, the host, the headers and the error handling. This is the second and last workspace in scope, so it closes the parent issue too.

## Details

**Cache module — `src/network/cache/`**

`keys.ts` holds all 102 key sites as factories, grouped by domain noun, with the prefix hierarchy made explicit. `events.ts` names the twenty writes core performs and holds each fan-out once, so the 47 invalidation calls across 30 files become 30 event calls.

Three collisions the key module now states rather than hides:

- `course_phase` and `coursePhase` cache the same phase under two literals, so the archive dialog shares neither the shell's entry nor its invalidation.
- `application_form` and `applicationForm` read different endpoints despite the near-identical name.
- The two-element `application` key holds a phase id on the student page and a participation id in the staff dialog.

The events preserve today's behaviour exactly, including three inconsistencies the tests now pin. An application assessment invalidates the participations of every phase rather than the one written to, because the key it uses is unscoped. The same is true of a student gaining university data, which is at least what that write means. And the configurator's graph save sends one composite key that matches nothing, so it invalidates none of the four graphs it names.

**Request module — `src/network/client.ts` + `src/network/api/`**

`client.ts` owns the `/api` prefix, which of the two shared axios instances a route belongs on, the content type on writes, and one place that logs a failure once, sanitized, and rethrows the original error so the `onError` handlers keep reading `error.response.data.error`. It wraps each call rather than installing an interceptor: `axiosInstance` is a Module Federation singleton shared with every remote, and core has no business logging their requests.

`api/` groups the endpoints into 13 namespaces by the domain noun their route already uses, so a call reads `coreApi.applications.form(phaseId)`. The two application namespaces are split the way the routes are: `applications` is the staff-facing `/api/applications`, `apply` is the student-facing `/api/apply`, whose three unauthenticated endpoints are the only reason the second axios instance exists.

The 63 one-endpoint modules are gone, along with the 57 copies of `try / console.error / throw`.

**One deliberate change to what goes over the wire.** Writes now all announce `application/json`. The modules this replaces were split between `application/json-path+json`, `application/json` and nothing at all, and three reads sent a content type on a request with no body. Every core handler binds with `BindJSON`, which never reads the header, so the split was noise rather than a contract.

**Two declared return types were wrong and are now what they always were.** `deleteCoursePhase` said `Promise<string | undefined>` while returning the whole axios response, which `handleSave` awaited and discarded. And `getStudentsWithCourses` loses the `console.log` of every student it fetched.

`getStudents` is deleted rather than ported: nothing imported it.

`archiveCourses` is the one write that mirrors its result into the course store rather than invalidating a cache, so it moves to `shared/hooks/courseArchive` and the request module keeps only the endpoint it calls.

## Reason / Link to issue

Closes #2099
Closes #2097

## How to Test

`make test-clients` and `make lint`. The three test files are the contract:

- `network/cache/keys.test.ts` asserts every key literal, the `undefined` id slot the current keys deliberately keep, and the prefix semantics that make an unscoped invalidation reach its descendants.
- `network/cache/events.test.ts` seeds a `QueryClient` and asserts, per event, which entries end up invalidated and which are deliberately left alone.
- `network/api/routes.test.ts` is the route manifest: for all 88 endpoints it asserts the method, the resolved URL, the body, the query parameters and the content type, plus the behaviour that is not a route — the 401 that means "member of no course" and answers with an empty list, the two 204s that map to a ready status, and the two endpoints where the server sends null instead of an empty list.

The manifest diffs against the deleted modules per method and URL with only `GET /api/students/` missing, the one endpoint nothing imported.

Worth clicking through by hand: course creation and copy (the invalidate-then-refetch pair the three creation paths share), the application participants table, and the mail campaign editor.

## PR Checklist


clients/core/src/network/cache/keys.ts
```diff
@@ -0,0 +1,116 @@
...
+    // `inPhase` and `ofParticipation` are the same two-element shape holding different ids: the
+    // student-facing page caches its own application by phase, the staff dialog caches one
+    // application by participation
+    inPhase: (phaseId: Id) => ['application', phaseId] as const,
+    ofParticipation: (courseParticipationId: Id) => ['application', courseParticipationId] as const,
+    ofParticipant: (phaseId: Id, courseParticipationId: Id) =>
+      ['application', phaseId, courseParticipationId] as const,
+    participations: {
+      all: () => ['application_participations'] as const,
+      inPhase: (phaseId: Id) => ['application_participations', phaseId] as const,
```
mathildeshagl: This is the one thing I would change, and it is worth changing precisely because of what this module is for.

`participations.inPhase(phaseId)` has exactly two consumers, and neither of them reads participations:

- `ApplicationDataWrapper.tsx:27` — `coreApi.applications.additionalScoreNames(phaseId)`
- `useCourseConfiguratorDataSetup.ts:101` — same call

The participations list is cached under `participations.students(phaseId)` (`useGetApplicationParticipations.ts`, `ApplicationStudentDetailComponent.tsx`), and nothing at all caches students. So of the two names this PR introduces here, one says participations and holds score names, the other says students and holds participations.

They are siblings rather than prefixes, so nothing breaks today. But this is the module whose header promises that "where two entries look like siblings but are not, the comment says so", and this pair is the one place that carries no comment. The next person who needs a real participations-by-phase cache will reach for `participations.inPhase` and land on the score-names entry.

Suggested shape:

```ts
additionalScores: (phaseId: Id) => ['application_participations', phaseId] as const,
participations: {
  all: () => ['application_participations'] as const,
  inPhase: (phaseId: Id) => ['application_participations', 'students', phaseId] as const,
},
```

with a comment that the `students` segment is historical and does not mean students. `participations.all()` stays the prefix of both, which is what `applicationAssessmentSaved` and `studentUniversityDataChanged` rely on.

clients/core/src/network/cache/events.ts
```diff
@@ -0,0 +1,119 @@
...
+      coreKeys.keycloak.userSearch.all(),
+    ]),
+
+  coursePhaseChanged: (queryClient: QueryClient, phaseId: Id): void =>
+    invalidate(queryClient, [coreKeys.coursePhases.byId(phaseId)]),
+
+  // One composite key rather than one key per graph, which is what the configurator has always
+  // sent. `invalidateQueries` matches by prefix, so nothing caches under it and nothing is
+  // invalidated; the configurator reloads its graphs by other means.
+  coursePhaseGraphSaved: (queryClient: QueryClient): void =>
```
mathildeshagl, line 47: Naming these three and holding them still is the right call for a refactor — I would not want the behaviour change hidden in here. But between them, `applicationAssessmentSaved` and `studentUniversityDataChanged` invalidating unscoped, and this key matching nothing, they now have tests asserting they behave as they do, and a passing test is a much better argument for leaving something alone than a `// TODO` ever was.

Could you open a follow-up issue listing the three, so the next reader finds a decision rather than inferring one from `events.test.ts`? The `courseGraphs` keys the same module already exposes look like they make this one a two-line fix.

Not blocking the merge.

clients/core/src/network/client.ts
```diff
@@ -0,0 +1,104 @@
...
+  }
+}
+
+interface ReadOptions {
+  params?: Record<string, string | number | boolean>
+}
+
+interface WriteOptions extends ReadOptions {
+  /** Sent as the request body of a DELETE, which is how core deletes a batch. */
+  data?: unknown
```
mathildeshagl, line 58: `data` on `WriteOptions` only exists for `del`, but `post` and `put` take `WriteOptions` too and spread `...options` after the positional `data`:

```ts
post: async <T = void>(url: string, data?: unknown, options?: WriteOptions) =>
  (await send<T>(instance, description, { method: 'post', url, data, headers: JSON_HEADERS, ...options })).data,
```

So `coreRequest.post(url, body, { data: somethingElse })` type-checks and silently drops `body`. Nothing does that today, and the compiler will not stop anyone who tries.

Splitting the types keeps the footgun out of reach:

```ts
interface WriteOptions extends ReadOptions {}
interface DeleteOptions extends ReadOptions {
  /** Sent as the request body of a DELETE, which is how core deletes a batch. */
  data?: unknown
}
```

clients/core/src/network/api/auditLog.ts
```diff
@@ -33,21 +36,19 @@ export const buildAuditLogQuery = (
...
-}
+export const auditLog = {
+  inCourse: (
+    courseID: string,
+    filters: AuditLogFilters,
+    limit: number,
+    cursor?: AuditCursor,
+  ): Promise<AuditLogPage> =>
+    coreRequest.get(
+      `${API_PREFIX}/courses/${courseID}/audit-log?${buildAuditLogQuery(filters, limit, cursor)}`,
```
mathildeshagl, line 51: Small consistency point: `client.ts` grew a `params` option, and every other read in `api/` uses it (`courses.nameExists`, for one), but the two audit-log reads hand-build a query string and concatenate `?...` onto the URL instead. Two mechanisms for the same job in one module.

I suspect the reason is real — `params?: Record<string, string | number | boolean>` cannot express "omit this key when the filter is empty", which is exactly what `buildAuditLogQuery`'s `set()` helper does, and `URLSearchParams` also fixes the encoding. If so, a one-line comment saying that would stop the next person from "tidying" it into `params` and quietly sending `actorRole=undefined`.

clients/core/src/network/api/courses.ts
```diff
@@ -0,0 +1,80 @@
...
+import { isAxiosError } from 'axios'
+import { API_PREFIX, coreRequest } from '../client'
+
+const path = `${API_PREFIX}/courses`
+const ofCourse = (courseID: string) => `${path}/${courseID}`
+
+const UNAUTHORIZED = 401
+
+/** A user with access to no course gets a 401 here rather than an empty list. */
+const emptyOnUnauthorized = async <T>(read: Promise<T[]>): Promise<T[]> => {
```
mathildeshagl, line 24: Not a change request, just noting it while the code is open.

`send()` in `client.ts` logs before it rethrows, so the 401 this helper exists to treat as normal still writes a `Core request failed { status: 401 }` to the console — on every load of the course list, for every user who is a member of no course. That is faithful to what it replaces (the old `getAllCourses` also did `console.error(err)` before its 401 check), so it is not a regression from this PR.

It is a one-liner now that logging lives in one place, though, which it was not before — a `quiet` flag or an expected-status list on the read would make the empty-list case silent. Your call whether that belongs here or in a follow-up.

mathildeshagl (dismissed):

clients/core/src/network/cache/keys.ts
rappm: Confirmed both consumers and renamed to your shape in 729d98ee. `additionalScores(phaseId)` now holds the score names, `participations.inPhase(phaseId)` holds the participation list, and both literals are unchanged, so every invalidation prefix is the same as before. `participations.all()` stays the parent of both.

`keys.ts` carries the two comments (the score-names entry sharing the participation prefix, and `students` not meaning students), and `keys.test.ts` splits the assertion in two so the score-names key is named in the test as well.

clients/core/src/network/cache/events.ts
rappm: Opened #2157. It lists all three (`coursePhaseGraphSaved` matching nothing, `applicationAssessmentSaved` and `studentUniversityDataChanged` unscoped), quotes the composite key, and points at `courseGraphs` plus `coursePhases.types()` as the two-line fix for the first one. No code change here, so the tests keep asserting the current behavior.

clients/core/src/network/client.ts
rappm: Fixed in c75148b0, with one deviation from your snippet: `interface WriteOptions extends ReadOptions {}` trips Biome's `noEmptyInterface`, so instead of an empty extension I renamed the shared type and gave only `del` its own:

```ts
interface RequestOptions {
  params?: Record<string, string | number | boolean>
}

/** `data` stays off `RequestOptions` so it cannot silently override a `post`/`put` body. */
interface DeleteOptions extends RequestOptions {
  /** Sent as the request body of a DELETE, which is how core deletes a batch. */
  data?: unknown
}
```

`get`, `getResponse`, `post` and `put` take `RequestOptions`, `del` takes `DeleteOptions`. `post(url, body, { data: somethingElse })` is now a type error, and `params` stays available on every method. `applications.ts:72` is the only `del` with a body and still type-checks.

clients/core/src/network/api/auditLog.ts
rappm: Your diagnosis is right and the comment is in c75148b0, on `buildAuditLogQuery`:

```ts
/**
 * Built by hand rather than through the `params` option: every filter is optional, and `set` drops
 * the empty ones from the query string entirely, which a `params` record cannot express.
 */
```

One correction to the failure mode you predicted: switching to `params` would not send `actorRole=undefined`, because axios skips `undefined` values in its serializer. It would send `actorRole=` for a filter the user cleared to an empty string, since `set()`'s `if (value)` drops those and axios does not. Same class of bug, different value. The comment says "empty ones" rather than naming either case.

clients/core/src/network/api/courses.ts
rappm: Follow-up, in #2158. It records that this is not a regression, and that the fix is now cheap because logging sits in one place: a `quietStatuses` list on the request options, stripped from the axios config and checked in `send`, with `courses.list` and `listOwnIDs` passing `[401]`.

Keeping it out of this PR for the same reason you gave on `events.ts`: threading a new option through `send` and every method is a behavior change plus its own tests, and this PR is worth more if it changes none.

Issue #742, opened 2025-07-25
Matching: Enable config view


mathildeshagl:

Mtze:
No settings page in this phase -> not important

Issue #562, opened 2025-05-27
Create architecture for Templating


mathildeshagl:
**Goal**
Design how course and phase templating should work across services and the database.

**Key Questions**

* What metadata must core copy when templating?
* Who owns the templating UI – core or a separate service?
* How should deadlines within phases be handled?
* How does core identify if a phase is templatable (DB flag or logic)?
* What happens if a phase is not templatable?

**Additional Considerations**

* Who can create/apply templates?
* Should partial (single-phase) templates be supported?

Issue #556, opened 2025-05-27
Create architecture/spec for server-side copy


mathildeshagl:
- The core needs a unified way to call the server's copy functionality
- The server needs to define which data should be copied

PR #2151, opened 2026-09-01
`Core`: Sort about page contributors by commit count

rappm:
## Summary

The about page ordered its contributor cards by a hand-maintained `position` field that had drifted well out of date. This replaces it with the commit count the GitHub contributors API already returns, refreshes the role labels to match what people actually worked on, and adds two contributors who were missing entirely.

## Details

**Ordering is now derived instead of hand-maintained.** `position` is gone from `contributorMapping`. `ContributorList` sorts on the `contributions` count from the contributors API, so the list re-orders itself as people commit and cannot drift again. An optional `pinnedPosition` override covers the two entries whose place is not a function of commit count: the project manager, and the creator of the original PROMPT, who is not returned by the contributors endpoint at all and therefore has no commit count to sort on. The comparator lives in `contributorSorting.ts` so it can be unit tested.

The resulting order: Matthias Linhuber (pinned), Valeryia Andraichuk (pinned), Niclas Heun, Maximilian Rapp, Mathilde Hagl, Josef Schmid, Magnus Kühne, Robert Jandow, Maximilian Sölch, Benedikt Geisberger, Felix T.J. Dietrich, Philipp Nagy.

**New entries.** Magnus Kühne (`magkue`) as Developer Tooling & Testing Infrastructure, and Felix T.J. Dietrich (`FelixTJDietrich`) as Intro Course Phase Maintainer. Both are returned by the contributors API, so neither needs the special-case fetch that `airelawaleria` requires.

**Refreshed roles**, based on the areas each person's merged PRs and closed issues actually cover:

| Login | Before | After |
| --- | --- | --- |
| `mathildeshagl` | Templating | Maintainer, Core Platform & Course Templating Concept |
| `rappm` | Assessment | Maintainer, Assessment Phase |
| `JGStyle` | Student History and Data Management | Student Data, Privacy & GDPR Compliance |

Mathilde owns the templating architecture (#556, #562) and the cross-service rollout that came out of it (#717 through #742, including dynamic config endpoint registration in the SDK), plus most of the core course model work: descriptions, colour and icon, course settings, archiving, and the `Participants` subpage convention now written into the UI guidelines. Josef's work moved from student history into privacy and compliance: versioned re-consent, the core deletion routine, and export plus deletion across every phase service.

**Minor cleanups** in the same expression: dropped a `.filter()` that could never remove anything, keyed the cards on `login` rather than the array index, which matters now that the order is dynamic, and deduped by `login` so the special-case fetch cannot produce two cards with the same key if `airelawaleria` ever appears in the contributors response.

## Reason / Link to issue

No tracking issue. The list had gone stale: two active contributors were absent, the most active contributor of the last six months was labelled with one phase name, and the ordering had not been revisited since the numbers were first assigned.

## How to Test

1. Open the about page (`/about`) and confirm the contributor cards render in the order listed above.
2. Confirm each card still links to the right GitHub profile and shows the updated role text.

Verified locally: `vitest run` (196 tests across 14 files, all passing, including the 5 new cases in `contributorSorting.test.ts`), `biome check` (no new findings; the 167 pre-existing warnings are unchanged from `main`), and `tsc --noEmit` on `clients/core` (clean). The page itself was not opened in a browser, so the rendered order is worth an eyeball during review.

## Screenshots

Not attached — the visible change is card order and label text, both listed above.

## PR Checklist

## Summary by CodeRabbit

* **Improvements**
  * Contributor cards are now ordered with pinned contributors first, followed by contributors with the most commits.
  * Added support for prioritizing selected contributors while preserving commit-based ordering for others.
  * Duplicate contributor entries are now removed for a clearer list.
  * Updated contributor information and descriptions.
  * Added two contributors to the list.
  * Improved card identity handling for more reliable rendering.


clients/core/src/publicPages/legalPages/components/ContributorMapping.tsx
```diff
@@ -2,57 +2,58 @@ export const contributorMapping: {
...
-    position: 10,
+  FelixTJDietrich: {
+    name: 'Felix T.J. Dietrich',
+    contribution: 'Intro Course Phase Maintainer',
+  },
+  phnagy: {
+    name: 'Philipp Nagy',
+    contribution: 'TEASE - Team Allocation Decision Support System',
+  },
+  airelawaleria: {
```
mathildeshagl: This is the one thing I would like changed before this lands.

`airelawaleria` is not fetched from the contributors endpoint — she comes from the separate `/users/airelawaleria` call in `ContributorList.tsx`, which hard-codes `contributions: 0` because that endpoint does not report them. She previously carried `position: 3`; now she has no `pinnedPosition`, so she falls to the unpinned branch of the comparator and is ordered by a commit count that is 0 by construction. Commit-count sorting can never place her anywhere but dead last, on every render, permanently.

For "Creator of PROMPT - the concept on which PROMPT 2.0 is based" that reads as a side effect of the new mechanism rather than a decision — and `contributorSorting.test.ts` now pins it as intended behaviour ("sorts a contributor without commits in this repository last").

Giving her a `pinnedPosition` is the fix, and it is exactly what the new field is for: her place in the list should not depend on how many commits she has, which is what your own comment on line 5 says.

clients/core/src/publicPages/legalPages/components/ContributorList.tsx
```diff
@@ -50,33 +51,31 @@ export const ContributorList = () => {
...
-                >
-                  {contributor.name}
-                </a>
-                <p className='text-sm text-muted-foreground mt-1'>{contributor.contribution}</p>
-              </div>
-            </CardContent>
-          </Card>
-        ))}
+      {mappedContributors.map((contributor) => (
+        <Card key={contributor.login}>
```
mathildeshagl, line 65: The rename from `key={index}` to `key={contributor.login}` is the right call, but it exposes something the index key was hiding: `vali` is appended to `contributors` unconditionally, so the day `airelawaleria` has a commit in this repository she is in both arrays, passes the filter twice, and renders as two cards with the same React key.

Not reachable today, and not a reason to hold the PR. A `dedupe by login` before `.map()` would close it for good — or the special-case fetch could go away entirely once she appears in the contributors response.

clients/core/src/publicPages/legalPages/components/ContributorMapping.tsx
```diff
@@ -2,57 +2,58 @@ export const contributorMapping: {
...
+    name: 'Mathilde Hagl',
+    contribution: 'Maintainer, Core Platform & Course Templating Concept',
   },
   JGStyle: {
     name: 'Josef Schmid',
-    contribution: 'Student History and Data Management',
-    position: 7,
+    contribution: 'Student Data, Privacy & GDPR Compliance',
+  },
+  magkue: {
```
mathildeshagl, line 36: Separate from the sorting: this PR also adds two contributors (`magkue`, `FelixTJDietrich`) and rewrites six contribution descriptions, including yours and mine.

No objection to any of the content — the new descriptions are more accurate than what they replace. But none of it follows from "sort by commit count", and the PR body does not mention it, so a reader of the log will not find out that the about page's copy changed here. Worth a line in the description.

mathildeshagl (changes requested):

clients/core/src/publicPages/legalPages/components/ContributorMapping.tsx
rappm: Agreed, and fixed in 6c8c2def: `airelawaleria` now has `pinnedPosition: 2`, so her place no longer depends on a commit count that is 0 by construction. That puts her back near the top where `position: 3` had her, ahead of Niclas rather than behind him. Pinning only her was the smaller change, and pinning Niclas too would take the lead-developer slot out of the commit-count sorting that is the point of the PR. Say the word if you would rather have the exact pre-PR top three.

I also renamed the test you flagged to "sorts an *unpinned* contributor without commits in this repository last" (that behaviour is still what the comparator should do for someone who genuinely has no commits) and added a case asserting a pinned contributor with 0 commits keeps its pinned position.

clients/core/src/publicPages/legalPages/components/ContributorList.tsx
rappm: Good catch. Closed it in 6c8c2def with a `uniqueByLogin` dedupe before the `.map()`. `vali` is placed first in the array so that, on a duplicate login, the `Map` keeps the contributors-endpoint entry (real commit count) rather than the hard-coded `contributions: 0`. Left the special-case fetch in place for now; it can go away entirely once she shows up in the contributors response.

clients/core/src/publicPages/legalPages/components/ContributorMapping.tsx
rappm: The description does cover it, and has since the PR was opened: the **New entries** paragraph names both `magkue` and `FelixTJDietrich` with their labels, and **Refreshed roles** carries a before/after table for the three rewritten descriptions plus the reasoning behind them. So a reader of the log should find it. If you meant it belongs in the **Summary** rather than only in **Details**, the second sentence there already says the PR "refreshes the role labels to match what people actually worked on, and adds two contributors who were missing entirely" - happy to make that more prominent if it read as buried.

Separately, I updated the body for the pinning fix: the stated resulting order and the `pinnedPosition` paragraph now match the code.

mathildeshagl (dismissed):

clients/core/src/publicPages/legalPages/contributorSorting.test.ts
```diff
@@ -0,0 +1,67 @@
...
+  })
+
+  it('keeps a pinned contributor without commits in this repository at its pinned position', () => {
+    expect(
+      order([
+        contributor('prolific', 747),
+        contributor('nocommits', 0, 2),
+        contributor('lead', 500),
+      ]),
+    ).toEqual(['nocommits', 'prolific', 'lead'])
```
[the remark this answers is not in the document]
rappm: Right, the assertion was not carrying its own name: with `nocommits` the only pinned entry, it sorted first whatever its position was, so the test would have passed with `pinnedPosition: 1` too.

Fixed in 3e59bf0e by adding `alsoPinned` at position 1, so the expected order is now `['alsoPinned', 'nocommits', 'prolific', 'lead']` and position 2 is what fails if the comparator stops honouring it. Gave it 0 commits as well, which also covers two pinned entries tying on commit count.

PR #2153, opened 2026-09-01
`Infrastructure`: Exempt the packages we maintain from the Renovate minimum release age

### LINKED ISSUE #2152: `Infrastructure`: Exempt the packages we maintain from the Renovate minimum release age
`renovate.json` applies a blanket `minimumReleaseAge` of 7 days to every dependency. That cooldown
exists to protect us from compromised or broken third-party releases, but it also applies to the
three packages we publish ourselves:

- `@tumaet/prompt-shared-state` (`prompt-edu/prompt-shared-state`)
- `@tumaet/prompt-ui-components` (`prompt-edu/prompt-lib`)
- `github.com/prompt-edu/prompt-sdk` (`prompt-edu/prompt-sdk`)

For these we are the publisher, we review the release before it is cut, and the cooldown buys us
nothing. It only delays picking up our own fixes by a week, which is why the SDK version has drifted
across the services (`v1.3.0` in `core`, `certificate`, `self_team_allocation`, `example_server` and
`presentation`, `v1.3.1` in `team_allocation`, `assessment` and `interview`).

### Proposal

Add a `packageRules` entry to `renovate.json` that sets `minimumReleaseAge: "0 days"` for those three
packages, so a release we publish is proposed on the next Renovate run. Everything else keeps the
7-day cooldown.

rappm:
## Summary

Renovate applies a blanket `minimumReleaseAge` of 7 days to every dependency. This PR exempts the
three packages we publish ourselves, so a release we cut is proposed on the next Renovate run instead
of a week later.

## Details

Adds a `packageRules` entry to `renovate.json` that sets `minimumReleaseAge: "0 days"` for:

- `@tumaet/prompt-shared-state` (`prompt-edu/prompt-shared-state`)
- `@tumaet/prompt-ui-components` (`prompt-edu/prompt-lib`)
- `github.com/prompt-edu/prompt-sdk` (`prompt-edu/prompt-sdk`)

The cooldown exists to protect us from compromised or broken third-party releases. For these three we
are the publisher and review the release before it is cut, so the delay only holds back our own
fixes. It is why the SDK version has drifted across the services (`v1.3.0` in `core`, `certificate`,
`self_team_allocation`, `example_server` and `presentation`, `v1.3.1` in `team_allocation`,
`assessment` and `interview`).

The rule is scoped by exact package name, so every other dependency keeps the 7-day cooldown. It is
placed last in `packageRules`, alongside the existing security-update exemption, and does not change
grouping: these packages still land in the `all non-major dependencies` PR.

## Reason / Link to issue

Closes #2152

## How to Test

1. `npx --package renovate -- renovate-config-validator renovate.json` reports
   `Config validated successfully`.
2. On the next Renovate run, a fresh release of one of the three packages is picked up without
   waiting 7 days, while third-party updates keep the cooldown.

## PR Checklist


renovate.json
```diff
@@ -16,6 +16,15 @@
       "description": "No minimum age for security updates",
       "matchCategories": ["security"],
       "minimumReleaseAge": "0 days"
+    },
+    {
+      "description": "No minimum age for the packages we maintain ourselves",
```
mathildeshagl, line 21: The three names are right and complete — `@tumaet/prompt-shared-state` and `@tumaet/prompt-ui-components` are the only `@tumaet` entries across `clients/*/package.json`, and `github.com/prompt-edu/prompt-sdk` the only `prompt-edu` require across `servers/*/go.mod`.

One thing this rule does not do, and I think you want it to: it sets `minimumReleaseAge` but leaves `groupName` alone, so a minor/patch bump of these three still matches the first rule and lands inside the `all non-major dependencies` group PR. The 0-day exemption gets the version in immediately, but only as one entry in a shared branch alongside whatever else is pending, under `prConcurrentLimit: 3`. If the point is "our own release should be mergeable the moment it is out", these bumps need to arrive on their own:

```json
{
  "description": "No minimum age for the packages we maintain ourselves",
  "matchPackageNames": [
    "@tumaet/prompt-shared-state",
    "@tumaet/prompt-ui-components",
    "github.com/prompt-edu/prompt-sdk"
  ],
  "minimumReleaseAge": "0 days",
  "groupName": null
}
```

Majors are unaffected either way, since the grouping rule only matches `minor` and `patch`.

Not blocking — the change is an improvement as it stands.

Unrelated to your diff: `build-servers / self-team-allocation / merge` is red on `Get "https://registry-1.docker.io/v2/": context deadline exceeded` while booting buildkit. Docker Hub flake, needs a re-run.

mathildeshagl (approved):

PR #2159, opened 2026-09-02
`Core`: Fix the cache invalidations that reached less than their name suggested

### LINKED ISSUE #2157: `Core`: Decide the intent of three cache invalidations that reach less than their name suggests
Three of the write events in `clients/core/src/network/cache/events.ts` invalidate something other than what their name implies. #2149 named them and deliberately held their behavior still, and `events.test.ts` now asserts what each one actually does, so the current state is recorded rather than accidental. What is missing is a decision on whether it is the intended state.

**1. `coursePhaseGraphSaved` matches nothing**

It sends one composite key rather than one key per graph:

```ts
invalidate(queryClient, [
  [
    "courses",
    "participation_data_phase_graph",
    "phase_data_phase_graph",
    "course_phase_types",
    "course_phase_graph",
  ],
])
```

`invalidateQueries` matches by prefix, and nothing caches under that array, so no query is invalidated. The configurator reloads its graphs by other means today. `coreKeys.courseGraphs` already exposes the three graph keys plus `coursePhases.types()`, so replacing the composite with those four looks like a two-line fix.

**2. `applicationAssessmentSaved` is unscoped**

It invalidates `applications.participations.all()`, so an assessment saved in one phase refetches the participation list of every cached phase. It takes no phase id at all, so scoping it would change the signature and its call sites.

**3. `studentUniversityDataChanged` is unscoped**

It invalidates `applications.all()` and `applications.participations.all()`. This one is plausibly correct as written (a student gaining a university account changes the rows that render them in every phase), but it deserves the same explicit confirmation as the other two.

For each: confirm the current behavior is intended and say so in the comment, or scope/fix it. The tests in `events.test.ts` will fail loudly on any change, which is the point.

Follow-up to https://github.com/prompt-edu/prompt/pull/2149#discussion_r3913685244

rappm:
## Summary

Resolves the three cache invalidations in `clients/core/src/network/cache/events.ts` that reached
less than their name suggested: one is deleted as dead, one is scoped to the phase it belongs to,
and one is confirmed as deliberately unscoped but widened to the caches it actually makes stale.

## Details

**`coursePhaseGraphSaved` is deleted rather than fixed.** It sent one composite key that nothing
caches under, so it invalidated nothing. Replacing it with the four real keys would only have
swapped dead code for unreachable code: its single call site in `handleSave` runs right after
`await mutateParticipationDataGraph(...)`, and that mutation's `onSuccess` calls
`window.location.reload()`, which throws the whole React Query cache away. Deleting the event also
makes `handleSave`'s `queryClient` prop unused, which removes the file's only `any`, and lets
`Canvas.tsx` drop `useQueryClient`. The reload keeps a one-line comment saying why it is there.

**`applicationAssessmentSaved` is scoped to the assessed phase.** It invalidated
`applications.participations.all()`, so an assessment saved in one phase refetched the participation
list of every cached phase. It now takes the `phaseId` its call site already reads from the route
and invalidates `applications.participations.inPhase(phaseId)`. The write only touches the
participation row keyed by that phase and participation, and the dialog reads score, pass status and
comments from that same per-phase query, so nothing else can go stale.

**`studentUniversityDataChanged` keeps its unscoped application reach, and gains two caches it was
missing.** The unscoped reach is correct and now says so: the call site writes a global student
record and the student's rows render in the applications and participations of every phase, none of
which are known there. But `coreApi.students.update` PUTs the whole `Student` record, and two caches
it makes stale were not invalidated:

- `students.byId(studentId)`, read by `StudentDetailPage` through `useStudent`
- the university user search, which caches `coreApi.students.search` and whose `SearchStudents`
  query matches on email, matriculation number and university login, all fields this write sets

To reach every cached search, `coreKeys.applications.universityUsers` becomes an `all()` /
`forSearch()` prefix family shaped like the existing `keycloak.userSearch` family, with its key
literal unchanged.

Deliberately left alone, having checked each one: `students.enrollments` (course data, no profile
fields), instructor notes (no embedded profile fields), and `students.withCourses` (fetched straight
into component state, so there is no key to invalidate).

Known limitation, unchanged by this PR: `handleSave` returns early on any mutation failure, so a save
that fails partway leaves the earlier writes committed with no reload and no invalidation. That is a
partial-write problem which cache invalidation cannot fix, and the handler already carries a
`TODO: move this to the server side to enable transaction control!`. Worth its own issue.

## Reason / Link to issue

Closes #2157. Follow-up to #2149, which named these events and deliberately held their behavior
still, leaving the decision on their intent open.

## How to Test

1. Open a course in the course configurator, change the phase graph and save. The page still reloads
   and the graph comes back as before.
2. Open an application phase's participants list, open one application and change its score or
   acceptance status. The list updates. Switch to another application phase and confirm its
   participation list was not refetched.
3. In an application's details dialog, add university data for a student who has none. The
   application lists, the student's detail page and any cached university user search all pick the
   change up.
4. `cd clients && yarn vitest run core` and `make lint`.

## PR Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated assessment changes to refresh data only for the relevant phase.
  - Updated student information changes to refresh the affected student and related university-user search results.
  - Improved cache updates when scoping information is unavailable, helping ensure refreshed data is displayed correctly.
  - Corrected university-user search caching to keep search results and updates synchronized.
  - Improved course configuration save behavior and clarified page refresh handling for newly saved phases.


clients/core/src/managementConsole/applicationAdministration/pages/ApplicationParticipantsPage/components/ApplicationDetailsDialog/hooks/mutateAssessment.ts
```diff
@@ -19,7 +19,7 @@ export const useModifyAssessment = (courseParticipationID: string) => {
       )
     },
     onSuccess: () => {
-      coreCache.applicationAssessmentSaved(queryClient)
+      coreCache.applicationAssessmentSaved(queryClient, phaseId)
```
mathildeshagl, line 22: `phaseId` comes straight from `useParams`, so it is `string | undefined` here, and `type Id = string | undefined` in `keys.ts` lets that through silently.

If it is ever undefined, `inPhase(undefined)` builds `['application_participations', undefined]`, which prefix-matches nothing — so the invalidation reaches no cache at all. That is the exact bug class this PR is fixing, and against `main` it is a narrowing regression: `participations.all()` always matched, whatever `phaseId` was.

The inconsistency is visible two lines up — the `mutationFn` already defends with `phaseId ?? 'undefined'`, but the cache call does not. Worth resolving one way or the other before this lands: either guard the call site, or narrow `Id` so a missing id cannot silently produce a key that matches nothing.

Not reachable today, since the dialog only renders under the `:phaseId` route. The rest of the PR looks good to me: `coursePhaseGraphSaved` really was a no-op (and `handleSave`'s last mutation reloads the page anyway, so it was doubly pointless — nice to see the `queryClient: any` go with it), `students.byId` is a real key in `useStudent.ts` rather than a speculative one, and `universityUsers.all()` prefix-matches `forSearch` correctly.

mathildeshagl (changes requested):

clients/core/src/managementConsole/applicationAdministration/pages/ApplicationParticipantsPage/components/ApplicationDetailsDialog/hooks/mutateAssessment.ts
rappm: Fixed in 7be37126, and I went for the second option rather than a guard at this call site.

A guard here would only cover this one hook. Every id-scoped event in `events.ts` takes its id from `useParams`, so all fourteen of them can be handed an `undefined` today; narrowing the module's local `Id` to `string` compiles into fourteen call sites that would each need their own guard, and every one of those guards would have to choose between skipping the invalidation and passing a placeholder that matches nothing either way.

So the fix sits in the `invalidate` helper: a key is truncated at its first `undefined` segment before it is handed to `invalidateQueries`. `participations.inPhase(undefined)` becomes `['application_participations', 'students']`, which prefix-matches every phase's participation list. Over-invalidating costs a refetch; under-invalidating leaves stale rows on screen, so that is the direction to fail in. Every key in `coreKeys` opens with a literal, never an id, so a truncated key can never collapse to the whole cache.

Added three tests: the assessment event falling back to every phase when the id is missing (and still leaving the applications themselves alone), plus a two-id key showing the truncation keeps the segments before the missing one, so a missing campaign id does not reach another course.

Left `phaseId ?? 'undefined'` in the `mutationFn` alone. It is wrong in the same spirit, but it is the idiom at thirteen call sites across core and untangling it is its own change.

Issue #2162, opened 2026-09-03
`Core`: Provision per-service audit ingest keys and forward the audit environment


rappm:
Sub-issue of #2125.

## Context

`audit.NewCoreSink` reads the reporting service's shared secret from a single environment variable,
`AUDIT_INGEST_KEY`, and core validates it against `AUDIT_INGEST_KEYS`
(`service1:key1,service2:key2`, parsed in `servers/core/auditLog/router.go:250`). Today
`docker-compose.yml` forwards the `AUDIT_*` settings only to `server-core`, so no phase container
has a key at all and every phase sink would come back `nil`.

Because `AUDIT_INGEST_KEY` is a single global name, forwarding the bare variable to every container
would hand all services the same secret, which defeats the point of a per-service key. Each service
needs its own value, so the compose entry has to map a service-specific variable onto the name the
SDK reads.

## What is needed

- add one `AUDIT_INGEST_KEY_<SERVICE>` per phase to `.env.template` (and `.env.dev.template` where
  it applies), next to the existing `AUDIT_INGEST_KEYS` comment block
- forward `AUDIT_ENABLED` and `AUDIT_INGEST_KEY=${AUDIT_INGEST_KEY_<SERVICE>}` to every `server-*`
  service in `docker-compose.yml` and `docker-compose.e2e.yml`
- extend the `AUDIT_INGEST_KEYS` documentation so it lists every phase, using the same service names
  the services report in `promptTypes.RegisterInfoEndpoint` (`assessment`, `certificate`,
  `example-service`, `interview`, `presentation`, `self-team-allocation`, `team-allocation`,
  `intro-course`)
- mirror the same variables in the deployment configuration
- document the key provisioning in the admin guide

## Why this comes first

The per-service sub-issues of #2125 all append to the same `AUDIT_INGEST_KEYS` value and edit the
same compose files. Landing the environment plumbing once keeps those PRs limited to their own
`main.go` instead of conflicting with each other.

Issue #2125, opened 2026-08-28
`Core`: Enable audit capture in the phase services


rappm:
## Context

Core captures its own mutating requests into the central audit log, but no phase service does. None
of the services under `servers/` register `audit.Middleware` or build an `audit.NewCoreSink`, and
`docker-compose.yml` forwards the `AUDIT_*` settings only to `server-core`. So the platform-wide
audit log currently only ever shows `sourceService = "core"`.

The developer guide (`docs/contributor/guide/audit-log.md`) describes the shared API for exactly this
case: core writes to its own database, phases ship entries to core over HTTP through the
`POST /api/audit` ingest endpoint.

## What is needed

Per phase service (`assessment`, `interview`, `team_allocation`, `self_team_allocation`,
`certificate`, `presentation`, `example_server`, `infrastructure_setup`):

- register `audit.Middleware(audit.NewCoreSink(coreHost, "<service>"))` on the service's API group,
  before any module registers its routes (Gin snapshots the middleware chain, see the comment on
  `auditLog.InitAuditLogCapture`)
- give the service its own `AUDIT_INGEST_KEY` and list it in core's `AUDIT_INGEST_KEYS`
- forward `AUDIT_ENABLED` and `AUDIT_INGEST_KEY` in `docker-compose.yml`, the deployment
  configuration, and `docker-compose.e2e.yml`
- add `audit.Describe` labels for the routes whose derived label is not readable enough

The ingest contract itself was corrected in #2124 (it used to reject the `error` outcome outright and
to silently coerce malformed identifiers to an unscoped `NULL`), so the endpoint is ready for this.

rappm:
Broken down into sub-issues, one per service:

- #2162 shared environment and compose plumbing (per-service ingest keys). The other sub-issues
  depend on this one, since they all append to the same `AUDIT_INGEST_KEYS` value and edit the same
  compose files.
- #2163 assessment, #2164 certificate, #2165 example-service, #2166 interview, #2167 presentation,
  #2168 self-team-allocation, #2169 team-allocation
- prompt-edu/prompt-intro-course#217 intro course
- prompt-edu/github-challenge-server#15 GitHub challenge

Two notes on the service list above:

- `infrastructure_setup` is not covered by a sub-issue. The phase itself is still in review in #1593,
  so its audit wiring belongs in that pull request rather than in a parallel issue. (The current
  `infrastructureSetup` module lives in the intro course server and is covered by
[...]

PR #2182, opened 2026-09-09
`Assessment`: Report audit events to the central audit log

### LINKED ISSUE #2163: `Assessment`: Report audit events to the central audit log
Sub-issue of #2125.

## Context

`servers/assessment` never registers `audit.Middleware`, so none of its mutating requests appear in the
central audit log. The service already depends on `prompt-sdk v1.3.1`, which ships the `audit`
package, so this is wiring only.

## What is needed

In `servers/assessment/main.go`:

- register the capture middleware on the API group **before any module registers its routes**. Gin
  snapshots the middleware chain when a route or group is registered, so a late `Use` silently
  captures nothing (see the comment above `auditLog.RegisterCapture` in `servers/core/main.go:186`):

  ```go
  api := router.Group("/assessment/api")
  api.Use(audit.Middleware(audit.NewCoreSink(
      promptSDK.GetEnv("SERVER_CORE_HOST", "http://localhost:8080"), "assessment")))
  ```

  The service name has to match the one reported by `promptTypes.RegisterInfoEndpoint`
  (`assessment`), because core resolves the ingest key by that name.

  `NewCoreSink` returns `nil` when `AUDIT_ENABLED` is unset or no `AUDIT_INGEST_KEY` is configured,
  and `Middleware` is then a no-op, so this is safe to land before an administrator switches
  auditing on.

- advertise the capability in the existing `promptTypes.RegisterInfoEndpoint` call:

  ```go
  promptTypes.CapabilityAuditLog: audit.Enabled(),
  ```

- add `audit.Describe("...")` to the routes whose auto-derived label is too generic (assessment completion, evaluation completion, schema modification, phase copy).

## Depends on

The environment and compose plumbing (`AUDIT_ENABLED` plus a per-service `AUDIT_INGEST_KEY`) is
tracked in its own sub-issue of #2125. This issue covers the service code only.

## Reference

`docs/contributor/guide/audit-log.md`, section "Enabling a new phase service". The ingest contract
itself was corrected in #2124.

rappm:
## Summary

`servers/assessment` now reports its mutating requests to the central audit log through the shared
`audit` package in `prompt-sdk`. This is the last of the seven phase services in #2125.

## Details

- `main.go` registers `audit.Middleware(audit.NewCoreSink(sdkUtils.GetCoreUrl(), "assessment"))` on
  the `/assessment/api` group, before `coursePhaseApi` is derived from it. The ordering is load
  bearing: Gin snapshots the handler chain when a route or subgroup is registered, so a later `Use`
  would silently capture nothing. `NewCoreSink` returns `nil` when `AUDIT_ENABLED` is unset or no
  `AUDIT_INGEST_KEY` is configured, and `Middleware(nil)` is a pass-through, so this is safe to land
  before an administrator switches auditing on.
- The service name matches the `ServiceName` already reported by `promptTypes.RegisterInfoEndpoint`,
  because core resolves the ingest key by that name. The `audit.log` capability is advertised on the
  same endpoint.
- Labels are attached where the derived label (the verb plus the last non-parameter path segment,
  naively singularized) is wrong or ambiguous. The completion routes are the worst set in the whole
  platform:

  | route | derived | label |
  | --- | --- | --- |
  | `POST`/`PUT` `/student-assessment/completed` | Created/Updated completed | Saved an assessment completion |
  | `POST .../completed/mark-complete` | Created mark complete | Marked an assessment as complete |
  | `PUT .../course-participation/:id/unmark` | Updated unmark | Unmarked an assessment as complete |
  | `DELETE .../course-participation/:id` | **Deleted course participation** | Deleted an assessment completion |
  | `POST`/`PUT` `/evaluation/completed/my-completion` | Created/Updated my completion | Saved my evaluation completion |
  | `POST .../my-completion/mark-complete` | Created mark complete | Marked my evaluation as complete |
  | `PUT .../my-completion/unmark` | Updated unmark | Unmarked my evaluation as complete |
  | `PUT /config` | Updated config | Updated the assessment configuration |
  | `POST /config/release` | Created release | Released the assessment results |
  | `POST /config/unrelease` | Created unrelease | Unreleased the assessment results |
  | `POST /config/reminders/send` | Created send | Sent evaluation reminders |
  | `POST /category-assessment` | Created category assessment | Saved a category assessment |

  `DELETE .../course-participation/:id` is the one that mattered most: the derived label claims a
  course participation was deleted, when the route deletes an assessment completion. `POST`
  `/category-assessment` is an upsert, so "Created" is wrong whenever the row already existed.
  Every label is registered ahead of the auth middleware and the `RequireAssessmentEnabled` guard, so
  a denied request keeps the precise label. Routes whose derived label is already correct
  (`/competency`, `/category`, `/assessment-schema`, `/student-assessment/action-item`,
  `/evaluation/feedback-items`, the two deletes) are left alone, and four of them are pinned by a
  test so a change to the SDK's derivation cannot silently rewrite them.
- **Two routes are deliberately silenced with `audit.Skip()`**: `POST /student-assessment` and
  `POST /evaluation`. Neither is a save button. The grading form posts on every score-level
  selection (`AssessmentForm.tsx`) and the evaluation form posts on every field change through a
  `form.watch` subscription (`EvaluationForm.tsx`), with no debounce. A tutor working through a
  20-competency sheet for 60 students, or a cohort doing peer evaluations at the deadline, would
  produce tens of thousands of entries. The SDK caps deliveries in flight at 64 and drops the
  overflow with only a `Warn`, so the flood would not merely add noise: it would open gaps in the
  log that look exactly like nothing having happened. The auditable decision is the completion
  transition, and those stay recorded, as do the deletes. **This is a policy call worth a second
  opinion** - the alternative is to accept the volume and keep every intermediate score write.
- `POST /copy` gets both a route label and an explicit `audit.Record` scoped to the target phase,
  naming the source in its metadata. The copy route sits outside `:coursePhaseID`, so an
  automatically captured event would carry no course phase and no resolvable course, and would only
  ever appear in the global admin log rather than the course log. `CopyPhase` really does copy data
  here (the phase config and schema). Core's copyability probe, which posts a copy request whose
  source and target are the same phase and runs the real handler, is suppressed outright.
- `auditWiring_test.go` mirrors `main.go`'s wiring with every service in the same order, an
  in-memory sink and the real `RegisterRoutes` calls. It pins all twelve labels plus four derived
  ones, that reads record nothing, that the two autosave routes record nothing, the suppressed
  probe, and the explicit copy event on the real success path over a testcontainers database.

Deliberately out of scope, all worth a look:

- The deployment plumbing (`AUDIT_ENABLED` plus a per-service `AUDIT_INGEST_KEY` for
  `server-assessment`, and an `assessment:<key>` entry in core's `AUDIT_INGEST_KEYS`) is #2162, so
  the capture is inert in every environment until that lands.
- `promptTypes.CapabilityAuditLog: audit.Enabled()` follows the documented convention, but
  `audit.Enabled()` only reads `AUDIT_ENABLED` while the sink additionally needs `AUDIT_INGEST_KEY`,
  so `/info` can report `audit.log: true` while the middleware is a pass-through.
- Automatically captured events carry no entity id. The SDK backstop fills `EntityID` from a `:uuid`
  or `:id` path parameter only, and these routes use `:courseParticipationID`, `:assessmentID`,
  `:schemaID` and so on, so "Deleted an assessment completion" cannot say for which student. This is
  the sharpest case across the seven phase services, because these are grading actions. Widening the
  SDK's parameter fallback would fix it everywhere at once.
- The privacy export and deletion endpoints sit on `api` and are now captured with derived labels
  and no course scope. Whether the phase legs should be labelled or skipped because core already
  audits the request that fans out is a platform-wide call.

## Reason / Link to issue

Closes #2163.

## How to Test

1. Set `AUDIT_ENABLED=true`, `AUDIT_INGEST_KEY=<secret>` for the assessment service and
   `AUDIT_INGEST_KEYS=assessment:<secret>` for core, then start core and the service.
2. As a course lecturer, change the phase configuration, grade a student and mark the assessment
   complete, then unmark it, then release the results.
3. Open the course audit log in the management console. The configuration change, the completion,
   the unmark and the release each appear once under their own label, with `sourceService`
   `assessment`. The individual score clicks do not appear.
4. As a student, fill in a peer evaluation and mark it complete. Only the completion appears.
5. `cd servers/assessment && go test ./...` covers all of the above without the stack.

## PR Checklist

## Summary by CodeRabbit

- **New Features**
  - Added audit logging for assessment completions, category assessments, evaluation completions, course-phase configuration changes, reminders, releases, and phase-copy actions.
  - Assessment services now advertise audit-log support when enabled.
  - Audit records include human-readable action descriptions and relevant outcome details.

- **Improvements**
  - Autosave form updates and internal copy checks are excluded from audit logs to reduce unnecessary activity.
  - Read-only assessment requests continue without generating audit events.


rappm:
Closing as a duplicate of #2172, which covers the same issue and was opened first. The four things this branch had on top have been ported onto that PR in 323039ee:

- an explicit `audit.Record` scoping the phase copy to the target phase, since `/copy` sits outside `:coursePhaseID` and an auto-captured event would only ever reach the global admin log
- suppression of core's copyability probe, which posts a copy of a phase onto itself and runs the real handler
- `audit.Skip()` on `POST /student-assessment` and `POST /evaluation`, which are per-interaction autosaves rather than save buttons
- `auditWiring_test.go`, pinning the middleware chain order and every action label

The label wording on #2172 was kept throughout.

PR #2183, opened 2026-09-09
`Core`: Provision per-service audit ingest keys and forward the audit environment

### LINKED ISSUE #2162: `Core`: Provision per-service audit ingest keys and forward the audit environment

rappm:
## Summary

Gives every phase service its own audit ingest key and forwards the audit environment to it, so the
capture middleware in the phase services has something to authenticate with. Until this lands their
sinks are all `nil` and the audit log only ever shows `sourceService = core`.

## Details

`audit.NewCoreSink` reads the reporting service's shared secret from one fixed name,
`AUDIT_INGEST_KEY`, and core validates it against `AUDIT_INGEST_KEYS` (`service1:key1,service2:key2`).
Forwarding the bare variable to every container would hand all of them the same secret, which defeats
the point of a per-service key. So each service gets its own `AUDIT_INGEST_KEY_<SERVICE>` and the
compose files map it onto the name the SDK reads:

```yaml
- AUDIT_ENABLED
- AUDIT_INGEST_KEY=${AUDIT_INGEST_KEY_TEAM_ALLOCATION}
```

- `.env.template`: one variable per phase, plus the list of every service name that must appear in
  core's `AUDIT_INGEST_KEYS`. The bare `AUDIT_INGEST_KEY` stays, since a phase started on the host
  (`make server-<name>`, `make servers`) reads it straight from the environment rather than through
  compose; the comment now says those share one value.
- `docker-compose.yml` (7 phase services) and `docker-compose.prod.yml` (6, the example service is
  not deployed to production) forward `AUDIT_ENABLED` and the mapped key, placed with the other
  `AUDIT_*` entries the way `server-core` already has them.
- `.github/workflows/deploy-docker.yml` writes the six production keys into `.env.prod` as
  environment secrets of the same name.
- `docker-compose.e2e.yml`: core had `AUDIT_ENABLED=true` but **no** `AUDIT_INGEST_KEYS`, so
  `parseIngestKeys("")` yields an empty map and any phase event would have been rejected with a 401.
  Core now has the list and every phase container the matching key, both as one obvious fixture
  value, so phase to core reporting works as soon as a phase registers the capture middleware.
- `docs/admin/audit_log.md` documents the per-service variable names, how to generate a key, and
  what a mismatch looks like.

Everything fails safe. An unset or empty `AUDIT_INGEST_KEY` makes `NewCoreSink` return `nil` and the
middleware a pass-through, and `parseIngestKeys` drops blank keys, so an unprovisioned service cannot
authenticate with an empty string. Keys can be added one service at a time.

### Two things worth knowing

**Core's `AUDIT_INGEST_KEYS` is deliberately still its own secret** rather than composed in the
workflow from the six per-service ones. Composing it would make the two sides impossible to desync,
which is tempting, but it would also silently drop the phases hosted outside this repository:
`intro-course` and `github-challenge` need an entry in core's list and have no
`AUDIT_INGEST_KEY_*` variable here. The admin guide instead states the sync requirement explicitly
and says what a mismatch looks like, because a wrong key is answered with a 401 that the SDK does not
retry, so the events are dropped with only a phase-side log line.

**The e2e keys are groundwork, not coverage yet.** No phase service registers the capture middleware
on `main` (that is the rest of #2125, one PR per service), so nothing posts to `POST /api/audit`
today and no spec asserts a non-`core` `sourceService`. The comment in the compose file says as much.
Once the phase PRs land, the natural follow-up is an e2e assertion that a phase mutation shows up
with its own `sourceService` - that assertion cannot be added here, because it would fail on this
branch.

`docker-compose.minimal.yml` is left untouched: it has no `AUDIT_*` settings on core either, so
adding only the phase side would be incoherent, and nothing in the Makefile or the docs references
that file.

## Reason / Link to issue

Closes #2162.

## How to Test

1. `docker compose -f docker-compose.yml config`, and the same for `docker-compose.prod.yml` and
   `docker-compose.e2e.yml`: all three parse, and every `server-*` phase service resolves
   `AUDIT_ENABLED` plus an `AUDIT_INGEST_KEY`.
2. With nothing set in `.env`, the resolved `AUDIT_INGEST_KEY` is empty for every phase, so auditing
   stays off. This is the default and it is what CI exercises.
3. Set `AUDIT_ENABLED=true`, `AUDIT_INGEST_KEY_TEAM_ALLOCATION=<key>` in `.env` and
   `AUDIT_INGEST_KEYS=team-allocation:<key>`, then `docker compose config` and confirm only that one
   service resolves the key.
4. In the e2e stack, `docker compose -f docker-compose.e2e.yml config` shows the fixture key on all
   seven phase services and the matching list on core.

## PR Checklist


rappm:
Closing as a duplicate of #2171, which covers the same issue, was opened first, and is more complete. Nothing is ported: #2171 does everything this branch did (the per-service `AUDIT_INGEST_KEY_<SERVICE>` variables mapped onto `AUDIT_INGEST_KEY` in every compose file, the deploy workflow, and the admin guide) plus three things this branch missed:

- a distinct key per service in e2e rather than one shared fixture, so a mismatched key would actually be caught
- the e2e keys in `e2e/.env.e2e` instead of hardcoded in the compose file
- `scripts/new-course-phase.sh` updated, so a new phase gets the variables automatically

Issue #2024, opened 2026-08-10
Core: Privacy deletion retry issue for partially failed deletion


JGStyle:
(originated in #1906)

In the main deletion, core data and cpm deletion happen all in parallel.

In the case that core succeeds but a cpm fails, a later retry won't be possible because the subject identifiers can't be assembled on the core anymore (as they've been deleted).

We could either only delete the core data if all others succeeded.

Or maybe we gather the subject identifiers first and store them on the deletion request object. That way we still have them available for a later retry, and only remove them once all succeeded.

JGStyle:
I think storing the subject identifiers in the deletion object is the better approach here, we could still delete everything and not have to hold back onthe core deletion

PR #2195, opened 2026-09-21
`Core`: Delete the phase modules' data when a phase or course is deleted

### LINKED ISSUE #2146: Core: Call the course phase deletion endpoints when a phase or course is deleted
### Problem or opportunity

The course phase deletion handlers are being implemented in every course phase module (#1866 to #1870, with #1866 delivered in #2141), but **nothing ever calls them**. Core deletes only its own rows, so every module keeps its data for a phase that no longer exists.

Two separate paths are affected:

**1. Deleting a single course phase.** `servers/core/coursePhase/router.go:192` calls `CoursePhaseService.DeleteCoursePhase`, whose whole body is:

```go
func (s *CoursePhaseService) DeleteCoursePhase(ctx context.Context, id uuid.UUID) error {
	return s.queries.DeleteCoursePhase(ctx, id)
}
```

The client does not fan out either: `clients/core/src/network/mutations/deleteCoursePhase.ts` sends one `DELETE /api/course_phases/:id` to core and nothing else.

**2. Deleting a whole course.** `course_phase` is declared `FOREIGN KEY (course_id) REFERENCES course(id) ON DELETE CASCADE` (`servers/core/db/migration/0001_schema.up.sql:52`), so `CourseService.DeleteCourse` (`servers/core/course/service.go:458`) removes every phase of the course in a single database cascade. That path never touches the per-phase deletion logic at all, so wiring up only case 1 would still leave every module's data behind whenever a course is deleted.

Until this is built, #1866 to #1870 are inert: the endpoints exist, are registered, and advertise the `phase.deletion` capability, but no caller reaches them.

### Desired outcome

Permanently deleting a course phase, or a course containing phases, removes the phase-scoped data held by the course phase modules as well as the rows in core's own database.

### Proposed solution

Core already does exactly this fan-out for privacy deletion, and that code is the natural template:

- `servers/core/privacy/service/deletion_handler.go:55-66` iterates the course phase types, skips any whose `BaseUrl` does not parse, and builds the target as `cpt.BaseUrl + sdkTypes.PrivacyRouteDataDeletion`.
- `servers/core/privacy/service/deletion_requestDeletionFromCPM.go` performs the call with the caller's `Authorization` header and a dedicated timeout, and the handler records a per-subrequest status so a partial failure is visible.

Steps:

1. Dispatch to every course phase module that advertises the `phase.deletion` capability when a phase is deleted, reusing the discovery above.
2. Cover the course deletion path too, so cascading a course through its phases still notifies the modules.
3. Decide what happens on a partial failure. The SDK requires the module handlers to be idempotent, so a retry is safe, but core should not report success while a module still holds the data. #2024 tracks the equivalent problem for privacy deletion and is worth reading first.
4. Settle the route scheme before more modules copy it. The discussion on this issue's parent concluded "send (POST /) DELETE on course phase route (NOT `/create` and `/delete`)", but prompt-sdk v1.3.1 shipped `RegisterCoursePhaseDeletionEndpoint` as `POST /delete`, which is what #2141 implements. One of the two needs to give.
5. Add a shared route constant in the SDK. `promptTypes/privacy.go` exports `PrivacyRouteDataExport` and `PrivacyRouteDataDeletion`, but there is no equivalent for phase deletion; the registrar hardcodes the literal `"/delete"`. Without a constant, core would hardcode the path independently, which is how the two sides drift apart.

### Additional context

Related: parent #984, module implementations #1866 (PR #2141), #1867, #1868, #1869, #1870, and prompt-edu/prompt-sdk#98, which is delivered in prompt-sdk v1.3.1 but still open.

rappm:
## Summary

Core now asks the course phase modules to drop their phase-scoped data before it deletes a course
phase or a course, so the deletion handlers the modules already implement are finally reached and a
deleted phase no longer leaves its data behind in every module.

## Details

The receiving end has existed since #2141: the modules register the SDK's phase deletion endpoint
and advertise the `phase.deletion` capability, but no caller reached it. Core deleted only its own
rows, and a course deletion cascaded straight through `course_phase` without the modules ever
hearing about it.

**The fan-out** (`servers/core/coursePhase/moduleDeletion.go`)

- One new query, `GetCoursePhaseDeletionTargets`, resolves the phases to delete to their type's
  `base_url` in a single round trip. No migration; `db/sqlc/` is regenerated and committed.
- Targets whose type is core-implemented (`base_url = 'core'`) have no module to ask and are
  skipped. `{CORE_HOST}` is resolved through the existing resolution service, which grew a
  `ResolveBaseURL` helper that `ReplaceResolutionURLs` now uses as well.
- Targets are grouped by resolved base URL, and one goroutine per module runs the capability probe
  plus that module's deletions. Every request carries the caller's `Authorization` header and a
  10 s deadline, redirects are never followed, and all errors are collected with `errors.Join`, so
  eight unreachable modules cost about ten seconds rather than eighty.
- `DELETE <baseURL>/course_phase/<coursePhaseID>` is the route prompt-sdk v1.3.3 registers, which is
  the scheme the parent issue asked for. The `POST /delete` of v1.3.1 is gone; every service in this
  repo is already on v1.3.3, so nothing on the module side changes.

**Why discovery goes through `/info` and not through the deletion endpoint's status code**

Inferring "this module has no deletion endpoint" from a 404 or 405 is not safe here. Module paths
are of the form `/team-allocation/api/...`, which `clients/nginx/spa.conf` does not exclude from the
SPA fallback (it returns 404 only for `/api` and `/api/`), so a reverse proxy that has lost the route
for a module answers a `DELETE` with **405** from nginx's static handler and a `GET` with the client
document under status **200**. Reading either as "nothing to delete" would silently orphan the
module's data. Core therefore reads the capability from `GET <baseURL>/info`, once per module, and
requires a body that decodes into the SDK's `ServiceInfo`; the SPA document fails that decode. The
client already treats `{baseUrl}/info` as the contract for every phase type in
`clients/core/src/network/api/system.ts`.

**Failure semantics**

The fan-out runs first and core keeps its rows unless it completed, which is the
"only delete the core data if all others succeeded" option discussed in #2024. It applies cleanly
here because, unlike privacy deletion, a failure destroys nothing the retry needs, and the SDK
requires the module handlers to be idempotent. A module that reports `phase.deletion: false` is
skipped: the sibling issues #1867 to #1870 are still open, and core cannot do better than record a
warning for a module that has not implemented the handler yet.

The consequence worth knowing: a course or phase whose module is unreachable, misrouted, or does not
serve `/info` cannot be deleted until that is fixed. That is deliberate. If it turns out to bite
operationally, the follow-up is an explicit force flag, not a silent skip.

In `DeleteCourse` the fan-out goes ahead of the Keycloak group deletion, because a module's own
authorization resolves the course lecturer role through core and needs both the phase row and the
roles to still exist. A failing module therefore leaves the course completely untouched.

**Known limitations, none introduced by this change**

- A phase created between the id lookup and the cascading course delete is removed without its
  module being asked. Closing that window needs a lock held across the module calls.
- Core's audit middleware records the successful deletion, but the phase route's path parameter is
  `:uuid` rather than `:coursePhaseID`, so the SDK extractor cannot scope the event to the phase,
  and a failed deletion is a 500 and thus not recorded at all. Both predate this PR.
- The module audit events this feature activates are delivered asynchronously and may land after
  the phase row is gone.
- prompt-sdk has no exported constant for the `course_phase` path segment (the SDK hardcodes the
  literal in `resolution.go` itself), so core declares its own next to the URL builder. A shared
  constant needs an SDK release and is left for a follow-up there.

No client change is needed: the fan-out is entirely server-side and the existing
`DELETE /api/course_phases/:id` and `DELETE /api/courses/:id` calls are unchanged.

## Reason / Link to issue

Closes #2146.

## How to Test

1. `cd servers/core && go build ./... && go vet ./... && go test ./...`
2. `servers/core/coursePhase/moduleDeletion_test.go` covers the behaviour against a seeded
   testcontainer with an `httptest.Server` standing in for a module: the module data and the phase
   row are both gone on success with the caller's token forwarded; a module without the capability
   is not called; a core-implemented phase type triggers no request at all; two phases of the same
   module share one capability probe; `{CORE_HOST}` is resolved; and a deletion answering 404, 405,
   500 or a redirect, an `/info` that 404s, 500s or returns the SPA document, an unreachable module,
   and an unusable `base_url` each fail the deletion and leave the phase row in place.
   `servers/core/course/deleteCourse_test.go` proves a failing module stops the course deletion
   before the Keycloak groups are touched.
3. Manually, with the dev stack up: add a Team Allocation phase to a course, create a team in it,
   delete the phase, and confirm the team is gone from the team allocation database. Repeat by
   deleting the whole course. Then stop the team allocation service and confirm the deletion is
   refused instead of silently orphaning the data.

## PR Checklist

## Summary by CodeRabbit

* **New Features**
  * Course and course-phase deletion now removes associated data from supported external modules.
  * Deletion requests forward the caller’s authorization token to external modules.
  * External module URLs can use the core server host placeholder.

* **Bug Fixes**
  * Deletion stops safely when module cleanup fails, preventing later records and access controls from being removed.
  * Unsupported, unavailable, invalid, or unsuccessful module responses no longer result in incomplete phase deletion.
  * Module capability checks avoid unnecessary requests and support cleanup across multiple modules.


servers/core/coursePhase/router.go
```diff
@@ -189,7 +189,7 @@ func (s *CoursePhaseService) deleteCoursePhase(c *gin.Context) {
 		return
 	}
 
-	err = s.DeleteCoursePhase(c, id)
+	err = s.DeleteCoursePhase(c, c.GetHeader("Authorization"), id)
```
mathildeshagl: Non-blocking: a module failure is returned to the client as-is via `handleError`, which includes the module URL and up to 1 KB of its response body, and this path never logs it (the course path does). I'd log the error and return a generic message here, maybe with a 502 for module failures.

servers/core/course/service.go
```diff
@@ -455,20 +461,58 @@ func (s *CourseService) UpdateCourseData(ctx context.Context, courseID uuid.UUID
...
+	// Ask the phase modules before anything else: the course row cascades into its phases, and the
+	// module's own authorization resolves the course lecturer role through core, which needs both
+	// the phase row and the Keycloak roles. On failure nothing has been touched yet.
+	cleanedPhaseIDs, err := s.coursePhases.DeleteModuleDataForCourse(ctx, authHeader, courseID)
+	if err != nil {
+		log.Error("Failed to delete course phase module data for course: ", err)
+		return errors.New("failed to delete the course phase data held by the phase modules")
+	}
+
+	tx, err := s.conn.Begin(ctx)
```
mathildeshagl: Non-blocking: the old code used `db.GetTimeoutContext`, this transaction runs on the request context without a deadline. The `FOR UPDATE` lock and the pool connection are now held while the Keycloak deletion makes network calls. A bounded context for the transaction would be safer.

servers/core/course/service.go
```diff
@@ -455,20 +461,58 @@ func (s *CourseService) UpdateCourseData(ctx context.Context, courseID uuid.UUID
...
+		log.Error(err)
+		return errors.New("failed to delete course")
+	}
+	defer sdkUtils.DeferRollback(tx, ctx)
+	qtx := s.queries.WithTx(tx)
+
+	// Lock the course row until the delete commits. Inserting a phase takes a key share lock on
+	// its course, so no phase can be added from here on. A phase added while the modules were
+	// asked would be cascaded away without its module being asked, so the deletion is refused.
+	if _, err := qtx.LockCourseForDeletion(ctx, courseID); err != nil {
```
mathildeshagl: Non-blocking: `LockCourseForDeletion` is `:one`, so deleting a course that doesn't exist now hits `ErrNoRows` and returns a 500 (before it was a 200). Mapping `pgx.ErrNoRows` to a 404 would be cleaner.

mathildeshagl (approved):
LGTM, the fan-out fails closed and the race with phase creation is handled nicely. A few non-blocking follow-ups inline.

servers/core/coursePhase/router.go
rappm: Fixed in 9e712a78: module failures are now wrapped in `ErrModuleDeletionFailed`, logged, and answered with a generic 502 (no module URL or body). Other failures are logged and answered with a generic 500. The course deletion handler does the same.

servers/core/course/service.go
rappm: Fixed in 8975756b: the transaction now runs under a 30s timeout context. That is longer than the default 10s query timeout since it also covers the Keycloak calls.

servers/core/course/service.go
rappm: Fixed in 78606d01: `pgx.ErrNoRows` from the lock is passed on and the handler answers 404. Added a test for a nonexistent course.

